PR #46, opened 2019-09-18
provide `Hbar` wrapper which applies tinybar conversions

### LINKED ISSUE #45: implement wrapper for calculating tinybar amounts in various hbar units
The wrapper should deal internally in `Tinybar` and then have a constructor which takes a value and a unit of hbar.

abonander:
Also, overload all methods which take tinybar amounts to accept `Hbar`.

I'd ultimately like to deprecate the other options somehow and only except values in `Hbar` but not sure when I would wanna do that.

closes #45 

### DIFF
--- a/src/account/CryptoTransferTransaction.ts
+++ b/src/account/CryptoTransferTransaction.ts
@@ -8,11 +8,12 @@ import {
     CryptoTransferTransactionBody,
     TransferList
 } from "../generated/CryptoTransfer_pb";
-import {checkNumber, getProtoAccountId, reqDefined} from "../util";
+import {getProtoAccountId, reqDefined, toTinybarString} from "../util";
 import BigNumber from "bignumber.js";
 import {CryptoService} from "../generated/CryptoService_pb_service";
 
 import {AccountId} from "../typedefs";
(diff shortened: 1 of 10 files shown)


abonander:
cc https://github.com/launchbadge/eslint-config-launchbadge/pull/2

src/account/CryptoTransferTransaction.ts
```diff
@@ -42,31 +43,33 @@ export class CryptoTransferTransaction extends TransactionBuilder {
...
-        }
+    addSender(accountId: AccountId, amount: number | BigNumber | Hbar): this {
+        if (amount instanceof Hbar || amount instanceof BigNumber) {
+            if (amount.isNegative()) {
+                throw new Error('sending amount may not be negative');
+            }
 
-        return this.addTransfer(accountId, -amount);
-    }
+            return this.addTransfer(accountId, amount.negated());
```
abonander: This is really cool: Typescript can resolve common fields/methods for sum types and project a sum type from that; the return type of this method call is `Hbar | BigNumber` which suits the signature of `addTransfer()`.

PR #50, opened 2019-09-19
improve `Hbar` and tinybar range checking throughout SDK

abonander:
also document `number | BigNumber` as a `Tinybar` type and explain supported ranges

after merge I'll rebase #49 onto this one.

### DIFF
--- a/src/Hbar.ts
+++ b/src/Hbar.ts
@@ -1,5 +1,4 @@
 import BigNumber from "bignumber.js";
-import {checkNumber} from "./util";
 
 const hbarAsTinybar = new BigNumber(100_000_000);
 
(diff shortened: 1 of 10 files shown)


src/account/AccountCreateTransaction.ts
```diff
@@ -39,19 +40,19 @@ export class AccountCreateTransaction extends TransactionBuilder {
...
     }
 
     setReceiveRecordThreshold(threshold: number): this {
-        checkNumber(threshold);
+        tinybarRangeCheck(threshold);
         this.body.setReceiverecordthreshold(threshold);
         return this;
     }
 
     setSendRecordThreshold(threshold: number): this {
```
abonander: These two methods will have their signatures changed in my `AccountInfoQuery` PR.

src/Hbar.ts
```diff
@@ -21,12 +20,12 @@ function convertToTinybar(amount: BigNumber.Value, unit: HbarUnit): BigNumber {
 
 export type HbarUnit = keyof typeof tinybarConversions;
 
-/** The possible denominations of hbar in order by magnitude */
+/** The possible denominations of HBAR in order by magnitude */
```
abonander: HBAR is the spelling used on Hedera's site so I figured that's what should be used in the SDK.

src/Hbar.ts
```diff
@@ -37,29 +36,37 @@ export const hbarUnitSymbols = {
...
         this.tinybar = tinybar;
     }
 
     /**
-     * Calculate the Hbar amount given a raw value and a unit
-     * (assumes value is in hbar if not provided)
+     * Calculate the HBAR amount given a raw value and a unit.
      */
-    public static from(amount: number | BigNumber | string, unit: HbarUnit = 'hbar'): Hbar {
+    public static from(amount: number | BigNumber | string, unit: HbarUnit): Hbar {
```
abonander: We agreed earlier to drop the default here, it just didn't make it into the original PR.

abonander: Since this wasn't published yet this shouldn't be a breaking change.

PR #62, opened 2019-09-26
Feature/contract queries

danielakhterov:
This includes a pretty big refactor to our types (moving them to separate files) and also renames `account-info` to `AccountInfoQuery`


src/util.ts
```diff
@@ -164,7 +158,7 @@ export function normalizeTxnId(txnId: TransactionIdLike): TransactionId {
     }
 }
 
-export function getProtoAccountId(accountId: AccountIdLike): AccountID {
+export function accountIdToProto(accountId: AccountIdLike): AccountID {
```
abonander: I think this belongs in `types/AccountId.ts`

src/util.ts
```diff
@@ -173,7 +167,7 @@ export function getProtoAccountId(accountId: AccountIdLike): AccountID {
     return acctId;
 }
 
-export function getProtoFileId(fileIdLike: FileIdLike): FileID {
+export function fileIdToProto(fileIdLike: FileIdLike): FileID {
```
abonander: Same for this in `types/FileId.ts`

src/util.ts
```diff
@@ -182,7 +176,7 @@ export function getProtoFileId(fileIdLike: FileIdLike): FileID {
     return fileId;
 }
 
-export function getProtoContractId(contractIdLike: ContractIdLike): ContractID {
+export function contractIdToProto(contractIdLike: ContractIdLike): ContractID {
```
abonander: Same for this in `types/ContractId.ts`

src/types/TransactionReceipt.ts
```diff
@@ -0,0 +1,23 @@
+import {ExchangeRateSet, exchangeRateSetToSdk} from "./ExchangeRateSet";
+import {TransactionReceipt as ProtoTransactionReceipt} from "../generated/TransactionReceipt_pb";
+import {AccountId, accountIdToSdk} from "./AccountId";
+import {ContractId, contractIdToSdk} from "./ContractId";
+import {FileId, fileIdToSdk} from "./FileId";
+
+export type TransactionReceipt = {
```
abonander: We can probably stand to change `Transaction.executeForReceipt()` to return this type so people don't have to use the protobufs.

src/types/TransactionRecord.ts
```diff
@@ -0,0 +1,34 @@
...
+export type TransactionRecord = {
+    receipt: TransactionReceipt | null;
+    transactionHash: Uint8Array | string;
+    consensusTimestamp: Date;
+    transactionId: TransactionId;
+    memo: string;
+    transactionFee: number;
+    contractCallResult: ContractFunctionResult | null;
+    contractCreateResult: ContractFunctionResult | null;
+    transferList: TransferList | null;
```
abonander: `TransferList` is an unnecessary indirection here; just use `AccountAmount[]` directly.

abonander: I'd just call this field `transfers` at that point, actually.

abonander: And maybe inline the `AccountAmount` definition into this file if it's not used anywhere else.

src/util.ts
```diff
@@ -43,7 +37,7 @@ export function getProtoTxnId(transactionId: TransactionIdLike): TransactionID {
 }
 
 export function newTxnId(accountId: AccountIdLike): TransactionID {
```
abonander: This probably now belongs in `types/TransactionId.ts`

src/contract/ContractBytecodeQuery.ts
```diff
@@ -0,0 +1,47 @@
...
+import {Query} from "../generated/Query_pb";
+import {grpc} from "@improbable-eng/grpc-web";
+import {Response} from "../generated/Response_pb";
+import {SmartContractService} from "../generated/SmartContractService_pb_service";
+import {contractIdToProto} from "../util";
+import {ContractIdLike} from "../types/ContractId";
+import {ContractGetBytecodeQuery} from "../generated/ContractGetBytecode_pb";
+
+export type ContractBytecode = {
+    bytecode: Uint8Array | string;
```
abonander: Just return the `Uint8Array`, this is unnecessary.

src/contract/ContractBytecodeQuery.ts
```diff
@@ -0,0 +1,47 @@
...
+            errors.push(".setContractId() required");
+        }
+    }
+
+    protected getMethod(): grpc.UnaryMethodDefinition<Query, Response> {
+        return SmartContractService.contractCallLocalMethod
+    }
+
+    protected mapResponse(response: Response): ContractBytecode {
+        const contractResponse = response.getContractgetbytecoderesponse()!.getBytecode()!;
```
abonander: You can do `.getBytecode_asU8()` to just get a `Uint8Array`.

src/contract/ContractCallQuery.ts
```diff
@@ -0,0 +1,48 @@
...
+import {grpc} from "@improbable-eng/grpc-web";
+import {Response} from "../generated/Response_pb";
+import {SmartContractService} from "../generated/SmartContractService_pb_service";
+import {contractIdToProto} from "../util";
+import {ContractIdLike} from "../types/ContractId";
+import {ContractFunctionResult, contractFunctionResultToSdk} from "../types/ContractFunctionResult";
+import {ContractCallLocalQuery} from "../generated/ContractCallLocal_pb";
+
+export type ContractCall = {
+    result: ContractFunctionResult | null;
```
abonander: Same thing here, just return the `ContractFunctionResult`.

src/contract/ContractCallQuery.ts
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    protected getMethod(): grpc.UnaryMethodDefinition<Query, Response> {
+        return SmartContractService.contractCallLocalMethod
+    }
+
+    protected mapResponse(response: Response): ContractCall {
+        const contractResponse = response.getContractcalllocal()!.getFunctionresult()!;
+        return {
+            result: contractResponse == null ? null : contractFunctionResultToSdk(contractResponse)
```
abonander: If `mapResponse()` was called then `contractResponse` should not be null at this point; a `HederaError` would have been thrown before we got here.

src/proto/CryptoGetInfo.proto
```diff
@@ -37,5 +37,5 @@ message CryptoGetInfoResponse {
         Duration autoRenewPeriod = 13; // The duration for expiration time will extend every this many seconds. If there are insufficient funds, then it extends as long as possible. If it is empty when it expires, then it is deleted.
         repeated Claim claims = 14; // All of the claims attached to the account (each of which is a hash along with the keys that authorized it and can delete it )
     }
-    AccountInfo accountInfo = 2; // Info about the account (a state proof can be generated for this)
+    AccountInfoQuery accountInfo = 2; // Info about the account (a state proof can be generated for this)
```
abonander: Bad rename bruh.

src/types/ContractFunctionResult.ts
```diff
@@ -0,0 +1,23 @@
+import {ContractLogInfo, contractLogInfoListToSdk} from "./ContractLogInfo";
+import {ContractFunctionResult as ProtoContractFunctionResult} from "../generated/ContractCallLocal_pb";
+import {ContractId, contractIdToSdk} from "./ContractId";
+
+export type ContractFunctionResult = {
+    contractId: ContractId;
+    contractCallResult: Uint8Array | string;
+    errorMessage: string;
+    bloom: Uint8Array | string;
```
abonander: Same for `contractCallResult` and here, just use `Uint8Array` and then `.get*_asU8()`.

src/types/ContractLogInfo.ts
```diff
@@ -0,0 +1,20 @@
+import {ContractLoginfo as ProtoContractLoginfo} from "../generated/ContractCallLocal_pb";
+import {ContractId, contractIdToSdk} from "./ContractId";
+
+export type ContractLogInfo = {
+    contractId: ContractId;
+    bloom: Uint8Array | string;
+    topicList: (Uint8Array | string)[];
+    data: Uint8Array | string;
```
abonander: Same for these three, just `Uint8Array` will do.

src/types/ExchangeRate.ts
```diff
@@ -0,0 +1,15 @@
+import {ExchangeRate as ProtoExchangeRate} from "../generated/ExchangeRate_pb";
+
+export type ExchangeRate = {
+    hbarEquiv: number;
+    centEquiv: number;
+    expirationTime: number;
```
abonander: Might as well turn this into a `Date`.

src/types/ExchangeRateSet.ts
```diff
@@ -0,0 +1,14 @@
...
+    currentRate: ExchangeRate;
+    nextRate: ExchangeRate;
+}
+
+export function exchangeRateSetToSdk(exchangeRateSet: ProtoExchangeRateSet): ExchangeRateSet {
+    return {
+        currentRate: exchangeRateToSdk(exchangeRateSet.getCurrentrate()!),
+        nextRate: exchangeRateToSdk(exchangeRateSet.getCurrentrate()!)
+    }
+}
```
abonander: Just inline this file with `ExchangeRate`, IMO.

src/types/TransactionReceipt.ts
```diff
@@ -0,0 +1,23 @@
...
+    status: number;
+    accountId: AccountId | null;
+    fileId: FileId | null;
+    contractId: ContractId | null;
+    exchangeRateSet: ExchangeRateSet | null;
```
abonander: We had a discussion at the office and @mehcode and I think we prefer `fieldName?: FieldType` which requires unset values to be `undefined` instead of `null`. It's technically unidiomatic as per https://stackoverflow.com/questions/5076944/what-is-the-difference-between-null-and-undefined-in-javascript but semantically it's identical unless you use `typeof` and the syntax is much nicer.

abonander:
This is missing `CallParams`/`FunctionResult` from the Java SDK but I think it's fine if we do that in a follow-up PR as both those classes are pretty involved.

src/types/TransactionRecord.ts
danielakhterov: @abonander The reasoning behind using `transferList` as the name and `TransferList` type is to make it as similar to the protobufs as possible. I was thinking if one were to look at our types and then at the protobufs it would be apparent what is what simply by looking at the names/types. Do you you think we should ignore this and instead make it as simple as we can?

src/types/TransactionReceipt.ts
```diff
@@ -0,0 +1,23 @@
...
+}
+
+export function receiptToSdk(receipt: ProtoTransactionReceipt): TransactionReceipt {
+    return  {
+        status: receipt.getStatus(),
+        accountId: receipt.getAccountid() == null ? null : accountIdToSdk(receipt.getAccountid()!),
```
abonander: This could be simplified to `receipt.getAccountid() && accountIdToSdk(receipt.getAccountid()!)`

Is there a reason you're not using that form?

PR #61, opened 2019-09-25
mark unstable APIs with prefixed underscore so it translates to JS

abonander:
Signed-off-by: Austin Bonander <austin@launchbadge.com>

PR #65, opened 2019-09-27
Refactor types

danielakhterov:
This is a branch off of #62 so this should be merged only *AFTER* #62 is merged. 
Rebased onto #61


src/Transaction.ts
```diff
@@ -116,35 +118,35 @@ export class Transaction {
...
     }
 
     public async executeForReceipt(): Promise<TransactionReceipt> {
         await this.execute();
-        return this.waitForReceipt();
+        return receiptToSdk(await this._waitForReceipt());
     }
 
-    private getReceipt(): Promise<TransactionReceipt> {
+    private getReceipt(): Promise<ProtoTransactionReceipt> {
```
abonander: This needs an underscore prefix.

src/crypto/DeriveOpts.ts
```diff
@@ -0,0 +1,11 @@
...
+ * BIP-44 key derivation options as described here:
+ * https://github.com/bitcoin/bips/blob/33e6283/bip-0044.mediawiki#account
+ *
+ * `purpose = "44"` and `coin = "3030" (Hedera HBAR)` are hardcoded.
+ */
+export type DeriveOpts = {
+    account: number;
+    change: number;
+    index: number;
+}
```
abonander: Where is this file used?

PR #77, opened 2019-10-17
Fix file create and file append validation

QuestofIranon:
Properly checks if contents are null on `FileCreateTransaction` and validates that contents do no exceed 4kb in `FileCreateTransaction` and `FileAppendTransaction`


QuestofIranon:
Apparently 4kb isn't the actual file size limit, but the total transaction size limit according to @gregscullard

mehcode:
I'd rather the SDK *not* error and let Hedera error unless we can be 100% correct, all the time, even a year from now.

PR #83, opened 2019-11-06
Implement AccountUpdateTransaction

danielakhterov:
- Add Serialization test for AUT
    - Export AUT
    - Use proto `UInt64Value` instead of `uint`

### DIFF
--- a/src/account/AccountUpdateTransaction.ts
+++ b/src/account/AccountUpdateTransaction.ts
@@ -1,0 +1,83 @@
+import { TransactionBuilder } from "../TransactionBuilder";
+import { Transaction } from "../generated/Transaction_pb";
+import { TransactionResponse } from "../generated/TransactionResponse_pb";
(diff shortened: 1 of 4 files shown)


src/account/AccountUpdateTransaction.ts
```diff
@@ -0,0 +1,83 @@
...
+        const value = new UInt64Value();
+        const tinybar: Tinybar = threshold instanceof Hbar ?
+            (threshold as Hbar).asTinybar() :
+            threshold as Tinybar;
+        if (tinybar instanceof BigNumber) {
+            tinybarRangeCheck(tinybar);
+            value.setValue((tinybar as BigNumber).toNumber());
+        } else {
+            value.setValue(tinybar as number);
+        }
```
abonander: I would extract this to a function in `Tinybar.ts`. and dedup with `setSendRecordThreshold()`.

PR #84, opened 2019-11-07
Move automatic signing and update tests

danielakhterov:
Signing now happens in the `TransactionBuilder` when `.build()` is called
instead of `.execute()` on `Transction`. This allows easier user of SDK.


abonander:
@danielakhterov we need a way for the user to build an unsigned transaction; in the Java SDK we do this by making the `client` nullable in `TransactionBuilder`. This is probably a good time to duplicate that here.

danielakhterov:
@abonander I felt that belonged in a different PR; https://github.com/hashgraph/hedera-sdk-js/pull/86

PR #85, opened 2019-11-07
Implement `toString` for `Transaction`

danielakhterov:


__tests__/account/AccountCreateTransaction.test.ts
```diff
@@ -100,6 +102,11 @@ describe('AccountCreateTransaction', () => {
             "transactionvalidduration": {
                 "seconds": 120
             }
-        });
+        };
+        expect(txnBody).toStrictEqual(expectedTxBody);
+
+        expectedTx.body = expectedTxBody;
+
+        expect(expectedTx.toString()).toEqual(transaction.toString());
```
abonander: `expectedTx.toString()` should be spitting out `[object Object]` in this context. I would check the string value itself.

src/Transaction.ts
```diff
@@ -191,6 +191,16 @@ export class Transaction {
...
     }
+
+    public toString(): string {
+        const tx = this.toProto().toObject();
+        const bodybytes = tx.bodybytes instanceof Uint8Array ?
+            Buffer.from(tx.bodybytes as Uint8Array) :
+            Buffer.from(tx.bodybytes as string, "base64");
+        tx.body = TransactionBody.deserializeBinary(bodybytes).toObject();
+
+        return tx.toString();
```
abonander: Yeah, this isn't going to do what you expect: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/toString#Description

> If this method is not overridden in a custom object, toString() returns "[object type]", where type is the object type.

You probably want `JSON.stringify(tx)` instead.

__tests__/account/AccountCreateTransaction.test.ts
```diff
@@ -19,51 +19,75 @@ describe('AccountCreateTransaction', () => {
         const bodybytes = "Cg4KCAjcyQcQ258JEgIYAxICGAMYgMLXLyICCHhaPwoiEiDgyOwnWKWHn/rCJqE8DFFreZ5y41FBoN2Cj5TTeYiktzD//////////384//////////9/SgUI0MjhAw==";
 
         const tx = transaction.toProto().toObject();
-        expect(tx).toStrictEqual({
-            body: undefined,
+        let expectedTx = {
+            body: undefined as undefined | Object,
```
abonander: was this necessary?

__tests__/account/AccountCreateTransaction.test.ts
```diff
@@ -19,51 +19,75 @@ describe('AccountCreateTransaction', () => {
...
                 sigpairList: [
                     {
+                        pubkeyprefix: "4MjsJ1ilh5/6wiahPAxRa3mecuNRQaDdgo+U03mIpLc=",
                         contract: "",
-                        ecdsa384: "",
                         ed25519: "Vueml5zTmYI2f9O2kcw6/zKFaIPn5WCOKD/jvhwO+EFN55yepuYa566qbD8Z274nncCi/aqVrr/M6NIu91OnAQoOCggI3MkHENufCRICGAMSAhgDGIDC1y8iAgh4Wj8KIhIg4MjsJ1ilh5/6wiahPAxRa3mecuNRQaDdgo+U03mIpLcw//////////9/OP//////////f0oFCNDI4QM=",
-                        pubkeyprefix: "4MjsJ1ilh5/6wiahPAxRa3mecuNRQaDdgo+U03mIpLc=",
-                        rsa3072: ""
+                        rsa3072: "",
+                        ecdsa384: ""
```
abonander: it's not enforced but these keys should probably be sorted for sanity reasons

__tests__/account/AccountCreateTransaction.test.ts
danielakhterov: Unfortunately yes, if I remove `as undefined | Object` I get an error while assigning to that field later in the test.

__tests__/account/AccountCreateTransaction.test.ts
danielakhterov: The reason these fields are sorted the way they are is because the stringified json won't match otherwise.

abonander: Oh that's gross. At least it matches the ordering in the protobuf so it's hopefully stable.

PR #86, opened 2019-11-08
Update `Transaction` and `TransactionBuilder` to support nullable clients

danielakhterov:


src/TransactionBuilder.ts
```diff
@@ -95,24 +99,34 @@ export abstract class TransactionBuilder {
...
+        const txProto = new Transaction_();
+        txProto.setBodybytes(this._inner.serializeBinary());
 
-        return new Transaction(this._client, url, txn, this._inner, this._method);
+        return new Transaction(
+            txProto,
+            this._inner,
+            this._method,
+            this._client,
+            node == null ? node : node[ 0 ]
```
abonander: This can be shortened to `node && node[0]`

mehcode: `node?.[0]` works in Typescript 3.7 (released)

PR #90, opened 2019-11-14
Update `Transaction`, `TransactionBuilder`, and `QueryBuilder`

danielakhterov:
User changes: 
- The `client` will no longer be supplied in the constructor of various transactions or queries. Instead the client could be optionally passed into the `TransactionBuilder.build()` method to automatically sign the transaction, or the client is required be passed directly into `Transaction.execute()` and `QueryBuilder.execute()`. 
- Note: `TransactionBuilder.build()` will fail if a client is not provided *and* an explicit node account id is not set using `TransactionBuilder.setNodeAccountId()` method. 
- Removed `Transaction.executeForReceipt()`. Instead the user is now required to execute the transaction using `Transaction.execute()`, and then call `Transaction.waitForReceipt()` to explicitly wait for the receipt of the transaction which could take several minutes to complete. This two stage process is now required for users to determine where the error came from; executing the transaction or waiting for the receipt of the transaction.

### DIFF
--- a/__tests__/MockClient.ts
+++ b/__tests__/MockClient.ts
@@ -23,7 +23,7 @@ class MockClient extends BaseClient {
 
 export const mockClient = new MockClient();
 
-export const mockTransaction = new CryptoTransferTransaction(mockClient)
+export const mockTransaction = new CryptoTransferTransaction()
     .addSender({ shard: 0, realm: 0, account: 2}, 100)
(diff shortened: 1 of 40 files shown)


danielakhterov:
cc @donaldthibeau @atul-hedera

mehcode:
The corresponding PR in Java was approved so I'm marking this as approved.

PR #93, opened 2019-11-20
removed extra () in example code

SimiHunjan:


mehcode:
@SimiHunjan That `()` is needed to (immediately) invoke the function. It's a bit confusing at first glance but it's a fairly standard practice in JavaScript to immediately invoke a function expression (known as IIFE).

PR #94, opened 2019-11-20
Implement Contract Call encoding

danielakhterov:
Implements encoding function calls in the solidity binary format. This includes encoding parameter types `uint32`, `bytes`, `string`, and `bool` which means encoding now fully supports dynamic types. 
User changes:
- `ContractCallQuery` now supports `setFunctionParameters()` being passed a `CallParams` object which
computes the the solidity binary format from js types.

Note to Reviewers: This PR is branched off another PR; PR #90 so there are a changes that look duplicated.

### DIFF
--- a/__tests__/contract/CallParams.test.ts
+++ b/__tests__/contract/CallParams.test.ts
@@ -1,6 +1,5 @@
 import { CallParams, FunctionSelector } from "../../src/exports";
-
-const utf8 = require("utf8");
+import BigNumber from "bignumber.js";
 
 describe("CallParams", () => {
     const uint32 = new Uint8Array(4);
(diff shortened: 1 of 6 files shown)


abonander:
@danielakhterov this API doesn't appear to work like it does in the Java SDK, do you have a justification for changing it?

danielakhterov:
@abonander This was mostly based off the solidity abi webpage. I'll update the api to work more similarly to the Java one.

mehcode:
The equivalent feature is _in_ the Java SDK so I'm marking this as approved.

src/contract/CallParams.ts
```diff
@@ -0,0 +1,751 @@
...
+
+    public addByteArray(byteArray: Uint8Array[]): this {
+        this.conditionallyAddType({ ty: ArgumentType.bytes, array: true });
+
+        return this.addParam(byteArray);
+    }
+
+    public addAddressArray(addresses: string[] | Uint8Array[]): this {
+        if (!Array.isArray(addresses)) {
+            throw new TypeError("string[] and Uint8Array[] are both arrays, how is this possible?");
```
abonander: This is possible if this API is used improperly from a pure Javascript context (i.e. without type-checking). I think we need to have a discussion on whether we want to have these kinds of assertions everywhere or to simply rely on type annotations, because it's pretty inconsistent throughout the SDK.

I would say we should only have type assertions if it's necessary to satisfy `tsc`, otherwise it's on the user to provide the correct types (as is the case in most, but not all, JS APIs).

At the very least, this error should be more authoritative (`expected array, got ${typeof addresses}` or something).

src/contract/CallParams.ts
```diff
@@ -0,0 +1,751 @@
...
+
+        this.currentArgument += 1;
+
+        return this;
+    }
+
+    /**
+     * NOT A STABLE API
+     */
+    public toProto(): Uint8Array {
```
abonander: Unstable/internal APIs should be prefixed with an underscore.

src/contract/CallParams.ts
```diff
@@ -0,0 +1,751 @@
...
+
+    return {
+        ty: argument,
+        array
+    };
+}
+
+export class FunctionSelector {
+    private func?: string;
+    private needsComma = false;
```
abonander: These need an underscore prefix.

src/contract/CallParams.ts
```diff
@@ -0,0 +1,751 @@
...
+
+        this.conditionallyAddType({ ty: ArgumentType.address, array: true });
+
+        return this.addParam(par);
+    }
+
+    // Allows user to add any parameter they wish and they type will be guessed from the
+    // type and size of the parameter passed in
+    // For more control, use the `add<Type>` methods instead.
+    public addParam(param: string | boolean | number | Uint8Array | BigNumber |
```
abonander: I don't like this method, too easy to get wrong and too hard to debug when you do (because the function selector will change unpredictably).

If we want something like this I would force the user to specify the Solidity equivalent type instead of guessing it as they should know based on the signature of the function they're trying to invoke; however, a dynamic API like this would be difficult/verbose to duplicate in Java and our primary goal right now is parity so I'd prefer if we just dropped this one. If nothing else it deserves more thought (and not just in the context of this SDK alone) before we commit to it.

For now, I would simply add a comment somewhere to open an issue if someone needs a type that isn't provided here.

src/contract/CallParams.ts
```diff
@@ -0,0 +1,751 @@
...
+        }
+    }
+
+    private conditionallyAddType(ty: SoldityType, length?: number): void {
+        if (this.func.argumentTypes.length === this.currentArgument) {
+            this.func.addParamType(ty, length);
+        }
+    }
+
+    public addUint8(param: number): this {
```
abonander: All these integer methods are overkill; in the Java SDK it's just one method that has the user specify the byte-width of the integer.

__tests__/contract/CallParams.test.ts
```diff
@@ -0,0 +1,110 @@
...
+    const uint32 = 16909060;
+
+    const bytes = new Uint8Array(10);
+    bytes[ 1 ] = 1;
+    bytes[ 4 ] = 4;
+    bytes[ 9 ] = 8;
+
+    // const uint64 = new Uint8Array(8);
+    // const uint64View = new DataView(uint64.buffer);
+    // uint64View.setUint32(0, 4294967295);
```
abonander: (non-blocker) What's with the commented-out code?

PR #115, opened 2019-12-06
Implement `Transaction.getRecord`, and resolve several other unrelated issues

### LINKED ISSUE #106: Transactions and QueryBuilders should retry execute on BUSY
https://github.com/hashgraph/hedera-sdk-java/blob/master/src/main/java/com/hedera/hashgraph/sdk/HederaCall.java#L75-L76

### LINKED ISSUE #111: TransactionId.account should be TransactionId.accountId

### LINKED ISSUE #112: TransactionId.newId should not exist
I don't see the difference from the constructor?

### LINKED ISSUE #113: fromProto and toProto should be private/protected methods everywhere
We don't want to expose that association directly

### LINKED ISSUE #114: In BaseClient, too much of the operator internals is public
```export abstract class BaseClient {
    public operator?: Operator;
    public operatorSigner?: Signer;
    public operatorPublicKey?: Ed25519PublicKey;
```

Not sure why we store / have all of that and why its all public. Look at the Java SDK Client. I think a step is not storing `operator`.

danielakhterov:
closes #114 
closes #113 
closes #112 
closes #111 
closes #106

### DIFF
--- a/src/QueryBuilder.ts
+++ b/src/QueryBuilder.ts
@@ -198,7 +198,7 @@ export abstract class QueryBuilder<T> {
                     Math.random() * ((2 ** attempt) - 1));
 
                 if (Date.now() + delay > validUntilMs) {
-                    throw new Error(`timed out waiting for consensus on transaction ID: ${txId}`);
+                    throw new HederaError(status);
                 }
 
                 await setTimeoutAwaitable(delay);
--- a/src/QueryBuilder.ts
+++ b/src/QueryBuilder.ts
@@ -67,11 +67,11 @@ export abstract class QueryBuilder<T> {
         const payment = new CryptoTransferTransaction()
             .setNodeAccountId(nodeId)
             .addRecipient(nodeId, this.amount ? this.amount : amount)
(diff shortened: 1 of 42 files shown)


src/QueryBuilder.ts
```diff
@@ -151,12 +173,42 @@ export abstract class QueryBuilder<T> {
...
+                ResponseCodeEnum.BUSY,
+                ResponseCodeEnum.UNKNOWN,
+                ResponseCodeEnum.OK
+            ] as number[])
+                .includes(status)) {
+                const delay = Math.floor(receiptRetryDelayMs *
+                    Math.random() * ((2 ** attempt) - 1));
+
+                if (Date.now() + delay > validUntilMs) {
+                    throw new Error(`timed out waiting for consensus on transaction ID: ${txId}`);
```
abonander: Just throw `HederaError` with the last status. This will be talking about `transaction ID: null` any query but receipts anyway.

abonander: It'd be a little confusing if the user got `HederaError(OK)` but we're already doing that in the Java SDK, plus the fewer error types the user has to think about the better.

PR #119, opened 2019-12-10
Re-factor QueryBuilder and fix a few incorrect assumptions

mehcode:
- Receipt Status is a separate thing that needs checked separately
 - Payments were being generated for receipt queries


src/QueryBuilder.ts
```diff
@@ -6,219 +8,223 @@ import {
...
-        throwIfExceptional(responseHeader.getNodetransactionprecheckcode());
+                    await this._generatePaymentTransaction(client, node, this._paymentAmount);
+                } else if (this._maxPaymentAmount != null || client.maxQueryPayment != null) {
+                    node = client._randomNode();
 
-        return Hbar.fromTinybar(responseHeader.getCost());
-    }
+                    // TODO[2020-01-01]: Use `??` when we can
+                    const maxPaymentAmount: Hbar = this._maxPaymentAmount ||
+                        client.maxQueryPayment!;
```
abonander: I think we made `client.maxQueryPayment` default to 1 hbar and then the user can set it to 0 to disable payments.

src/util.ts
```diff
@@ -85,33 +84,24 @@ export function setTimeoutAwaitable(timeoutMs: number): Promise<null> {
...
-            "missing body from Response"
-        );
+export function timeoutPromise<T>(
+    ms: number,
+    promise: Promise<T>,
+    timedOutCallback: (reject: (reason: Error) => void) => void
+): Promise<T> {
+    const timeout = new Promise<T>((resolve, reject) => {
+        const timerId = setTimeout(() => {
+            clearTimeout(timerId);
```
abonander: `clearTimeout()` call shouldn't be necessary.

src/QueryBuilder.ts
mehcode: It does actually now that I double check the Client class.

PR #146, opened 2019-12-31
TransactionId Parity

danielakhterov:


src/Time.ts
```diff
@@ -0,0 +1,13 @@
+export class Time {
```
mehcode: Make sure to export this from exports

PR #154, opened 2020-01-04
Parity

danielakhterov:
- Make `TransactionReceipt.exchangeRateSet` be private
    - Make `Status` a class with static fields instead of enum
    - Implement `Status.toString()`
    - Fix encoding of `Int*Array` and `AddressArray`
    - Add test for encoding `Int*Array and `AddressArray`


src/Status.ts
```diff
@@ -1,108 +1,226 @@
...
+    public static readonly InvalidExchangeRateFile = new Status(98);
+    public static readonly InsufficientLocalCallGas = new Status(99);
+    public static readonly EntityNotAllowedToDelete = new Status(100);
+    public static readonly AuthorizationFailed = new Status(101);
+    public static readonly FileUploadedProtoInvalid = new Status(102);
+    public static readonly FileUploadedProtoNotSavedToDisk = new Status(103);
+    public static readonly FeeScheduleFilePartUploaded = new Status(104);
+    public static readonly ExchangeRateChangeLimitExceeded = new Status(105);
+
+    public readonly value: number;
```
mehcode: @abonander Thought on a name here? `code` perhaps?

src/Status.ts
```diff
@@ -1,108 +1,226 @@
...
+
+    public constructor(value: number) {
```
mehcode: I'd mark this constructor as private.

src/Status.ts
```diff
@@ -1,108 +1,226 @@
...
+            case Status.InvalidFeeFile.value: return "INVALID_FEE_FILE";
+            case Status.InvalidExchangeRateFile.value: return "INVALID_EXCHANGE_RATE_FILE";
+            case Status.InsufficientLocalCallGas.value: return "INSUFFICIENT_LOCAL_CALL_GAS";
+            case Status.EntityNotAllowedToDelete.value: return "ENTITY_NOT_ALLOWED_TO_DELETE";
+            case Status.AuthorizationFailed.value: return "AUTHORIZATION_FAILED";
+            case Status.FileUploadedProtoInvalid.value: return "FILE_UPLOADED_PROTO_INVALID";
+            case Status.FileUploadedProtoNotSavedToDisk.value: return "FILE_UPLOADED_PROTO_NOT_SAVED_TO_DISK";
+            case Status.FeeScheduleFilePartUploaded.value: return "FEE_SCHEDULE_FILE_PART_UPLOADED";
+            case Status.ExchangeRateChangeLimitExceeded.value: return "EXCHANGE_RATE_CHANGE_LIMIT_EXCEEDED";
+            default: return "UNKNOWN STATUS CODE";
```
mehcode:
```
UNKNOWN STATUS CODE (4120)
```

src/Status.ts
danielakhterov: Constructor is used in other file which is why it's public.

mehcode: Should still add some kind of doc there stating it's "unstable" or otherwise not intended for public use.

PR #157, opened 2020-01-08
Fix `FileInfoQuery.getCost()` returning 0 on deleted file

danielakhterov:


mehcode:
Please add comments around this stating _why_ we're doing this

Issue #134, opened 2019-12-26
Improve examples structure and coverage


Cooper-Kunz:
**Description** 

Currently, if you visited the examples, you wouldn't know what's supported given that there's nothing for smart contracts, files, threshold keys (e.g. #126), etc. 

Screenshot below of the current `/examples` folder:

<img width="1011" alt="Screen Shot 2019-12-26 at 12 48 39 PM" src="https://user-images.githubusercontent.com/11776938/71488506-237f7500-27de-11ea-81f7-22df5e67010b.png"> 

Based on #103 & #125 I think there are a few examples that need to be updated, also.

---

**Suggestion**

Create the following structure within `/examples`: 

- hello-hedera.js <- contains a "Hello World" example
- /cryptocurrency
- /files
- /smart-contracts
- /miscellaneous

Short term goal (scope of this issue): Implement popular / required APIs for each service.
Long term goal: implement each API in priority, based on usage and availability within the SDK.

---

I'm happy to work on this in my free time.

I've been using [v0.8.0-beta.0](https://github.com/hashgraph/hedera-sdk-js/tree/v0.8.0-beta.0) but can target `beta.3`, if it's further updated & stable.

mehcode:
@Cooper-Kunz On "Supported" I think we should adopt a matrix of sorts for SDK support in the Readme. I'm not sure how that would look. Perhaps just:

- [x] Crypto-Currency
   - [x] Create Account
   - [x] Update Account

... etc.

---

On examples, I agree that we need a good deal more of them. Java has a good amount and we could transfer those over here to showcase more of the SDK as low hanging fruit.

Cooper-Kunz:
A 'supported features' matrix makes sense to me - perhaps even a consistent one across each SDK.

How do you feel about the following structure, breaking it out by service? (fake content, obviously)

| Cryptocurrency | Support | References |
| --- | --- | --- |
| Create Account | ✅ | [docs](), [1]()
| Delete Account | 🚫 | [docs](), [1](), [2]()
| Get Account Balance | ✅ | [docs](), [1]()
| Get Stakers | 🚫 | [docs]()

e.g...

| File Service | Support | References |
| --- | --- | --- |
| Create File | ✅ | [docs]()
| Delete File | ✅  | [docs](), [1](), [2]()
| Get File Contents | ✅ | [docs](), [1]()

---

[...]

PR #176, opened 2020-01-28
Add new examples 

Cooper-Kunz:
I've added the following examples as part of #134 

- create-file.js
- delete-file.js
- multi-party-transfer.js
- confirmation-mechanisms.js
- create-consensus-topic.js
- create-threshold-account.js 

Please review these when you get a chance, @mehcode & @SimiHunjan 👍  

Also, let me know if there are any I'm missing which would be helpful.


examples/confirmation-mechanisms.js
```diff
@@ -0,0 +1,43 @@
...
+  const operatorPrivateKey = process.env.OPERATOR_KEY;
+  const operatorAccount = process.env.OPERATOR_ID;
+
+  if (operatorPrivateKey == null || operatorAccount == null) {
+    throw new Error(
+      "environment variables OPERATOR_KEY and OPERATOR_ID must be present"
+    );
+  }
+
+  const client = new Client({
```
mehcode: @Cooper-Kunz 

How do you feel about the "new style" of making Clients?

```js
const client = Client.forTestnet()
    .setOperator(operatorAccount, operatorPrivateKey);
```

I'd like to discourage people from using a single node for their network configuration.

examples/confirmation-mechanisms.js
```diff
@@ -0,0 +1,43 @@
...
+      account: operatorAccount,
+      privateKey: operatorPrivateKey
+    }
+  });
+
+  // Confirmation 1 - Executing will run a "precheck"
+  const transactionId = await new CryptoTransferTransaction()
+    .addSender(operatorAccount, 0)
+    .addRecipient("0.0.1", 0)
+    .setMemo("testing :)")
```
mehcode: `setMemo` is now `setTransactionMemo` as of v1 ( given that there are now 3 types of memos for transactions )

examples/confirmation-mechanisms.js
```diff
@@ -0,0 +1,43 @@
...
+    network: { "0.testnet.hedera.com:50211": "0.0.3" },
+    operator: {
+      account: operatorAccount,
+      privateKey: operatorPrivateKey
+    }
+  });
+
+  // Confirmation 1 - Executing will run a "precheck"
+  const transactionId = await new CryptoTransferTransaction()
+    .addSender(operatorAccount, 0)
```
mehcode: Note that as of v1 you _can_ do this:

```js
  .addSender(operatorAccount, new Hbar(1))
  .addRecipient("0.0.1", new Hbar(1))
```

Just a plain number is still tinybar

examples/confirmation-mechanisms.js
```diff
@@ -0,0 +1,43 @@
...
+    .addRecipient("0.0.1", 0)
+    .setMemo("testing :)")
+    .execute(client);
+
+  // Confirmation 2 - Ask for a receipt
+  console.log("attempting to get receipt for transaction id = " + transactionId + '\n');
+  const receipt = await transactionId.getReceipt(client);
+  console.log('transaction ' + transactionId + ' receipt = ' + JSON.stringify(receipt) + '\n');
+
+  // Confirmation 3 - Ask for a record
```
mehcode: Asking for a receipt when asking for a record is redundant. Perhaps explain that a **receipt** is what is necessary for ensure that consensus was reached. A **record** is useful only if you need the properties it provides. It has no confirmation guarantees above a **receipt**. Additionally, I would be sure to note that a **record** does cost to fetch. I don't remember the exact cost though.

examples/create-consensus-topic.js
```diff
@@ -0,0 +1,29 @@
...
+    operator: {
+      account: operatorAccount,
+      privateKey: operatorPrivateKey
+    }
+  });
+
+  const tx = await new ConsensusTopicCreateTransaction().execute(client);
+  console.log("tx:", tx);
+
+  const record = await tx.getRecord(client);
```
mehcode: Like I mentioned earlier, records are only useful if you need the additional properties. As they are non-free, always show using `getReceipt` in our examples unless we need the record itself.

examples/create-consensus-topic.js
```diff
@@ -0,0 +1,29 @@
...
+  const newTopicId = record.receipt._topicId;
```
mehcode: `_topicId` should be `.getTopicId()`

---

For context this changed to match Java _and_ to provide a tiny bit of debugging help so if you call `.getFileId()` where the was no file ID it will throw an exception with more information besides just emitting "undefined".

examples/create-file.js
```diff
@@ -0,0 +1,33 @@
+const { Client, FileCreateTransaction, Ed25519PublicKey, Hbar } = require("@hashgraph/sdk");
+
+async function main() {
+  
+  const operatorAccount = process.env.OPERATOR_ID;
+  const operatorPrivateKey = process.env.OPERATOR_KEY;
+  const operatorPublicKey = Ed25519PublicKey.fromString(process.env.OPERATOR_PUB_KEY);
```
mehcode: Please do

```js

  const operatorPrivateKey = Ed25519PrivateKey.fromString(process.env.OPERATOR_KEY);
  const operatorPublicKey = operatorPrivateKey.publicKey;
```

examples/create-file.js
```diff
@@ -0,0 +1,33 @@
...
+    operator: {
+      account: operatorAccount,
+      privateKey: operatorPrivateKey
+    }
+  });
+
+  const transactionId = await new FileCreateTransaction()
+    .setContents("Hello, Hedera's file service!")
+    .addKey(operatorPublicKey) // Defines the "admin" of this file
+    .setMaxTransactionFee(Hbar.of(1000))
```
mehcode: `Hbar.of` is deprecated (indefinitely) and replaced with `new Hbar` to better match other languages

examples/create-file.js
```diff
@@ -0,0 +1,33 @@
...
+    .execute(client);
+
+  const receipt = await transactionId.getReceipt(client);  
+  console.log("new file id = ", receipt._fileId);
```
mehcode: `getFileId` over `_fileId`

examples/create-threshold-account.js
```diff
@@ -0,0 +1,54 @@
...
+
+  // Generate our key lists
+  const privateKeyList = [];
+  const publicKeyList = [];
+  for (var i = 0; i < 4; i++) {
+    var privateKey = await Ed25519PrivateKey.generate();
+    var publicKey = privateKey.publicKey;
+    privateKeyList.push(privateKey);
+    publicKeyList.push(publicKey);
+    console.log(i + ": pub key:" + publicKey);
```
mehcode: you probably meant `priv key` here

examples/delete-file.js
```diff
@@ -0,0 +1,51 @@
...
+    operator: {
+      account: operatorAccount,
+      privateKey: operatorPrivateKey
+    }
+  });
+
+  // First, we'll create a file with our operator as an admin
+  const transactionId = await new FileCreateTransaction()
+    .setContents("creating a file to test deletion")
+    .setMaxTransactionFee(Hbar.of(100))
```
mehcode: It should only cost ~2 hbar to create a file. For exposition we should try and keep the max declared fee close to what it costs.

examples/confirmation-mechanisms.js
Cooper-Kunz: I'm a fan. It feels more consistent with the rest of the SDK's usage.

Is this available in recent releases?

examples/confirmation-mechanisms.js
Cooper-Kunz: I think the code example was just to demonstrate how they would be fetched clearly - in order of detail that they provide. Hence the duplication 😄 

But I will definitely explain the context, why you would get one vs. others, and the fees. Perhaps it would just make more sense to link to those items though? e.g. fees in docs, and a blog post?

examples/create-file.js
Cooper-Kunz: That's a much nicer usage 🎊

examples/confirmation-mechanisms.js
mehcode: Yes.

Cooper-Kunz: What network information does this default to for testnet & mainnet respectively? All of the address book at random, via the myhbarwallet proxy?

mehcode: @Cooper-Kunz Depends. 

On Node.js, `Client.forTestnet()` is the same as:

```js
new Client({ network: {
  "0.testnet.hedera.com:50211": "0.0.3",
  "1.testnet.hedera.com:50211": "0.0.4",
  "2.testnet.hedera.com:50211": "0.0.5",
  "3.testnet.hedera.com:50211": "0.0.6"
}});
```

On the Browser, it is:


```js
new Client({ network: {
  "grpc-web.myhbarwallet.com": "0.0.3",
}});
```

---

When making transactions or queries, a random node is picked from your local address book each time for client side load balancing.

Cooper-Kunz: Great clarification, thanks! 😄

Cooper-Kunz: bf418042bf409817a43a9011523c5511cbe0e072 should have updated _most_ of the examples to use this pattern @mehcode

Cooper-Kunz:
Hello @mehcode 👋  Are you able to give the updates a review when you get a chance? 

@SimiHunjan are there any additional examples that would be helpful for yourself/docs?

examples/create-threshold-account.js
```diff
@@ -0,0 +1,49 @@
...
+    thresholdKey.add(publicKeyList[i]);
+  }
+
+  // Create a new account with this threshold key
+  const accountCreateTransaction = await new AccountCreateTransaction()
+    .setKey(thresholdKey)
+    .setInitialBalance(0)
+    .execute(client);
+
+  console.log(await accountCreateTransaction.getReceipt(client)._accountId);
```
danielakhterov: Should  use `accountCreateTransaction.getReceipt(client).getAccountId()` since `TransactionReceipt._accountId` is meant to be private and for internal use only.

Cooper-Kunz: Hey @danielakhterov no reason for the client config being on two lines, I was just trying to match [this example](https://github.com/hashgraph/hedera-sdk-js/blob/c7d5a38a2ebc8a66e84d248a4baae5fa0d42e70b/examples/get-account-balance.js#L11) which was added by someone on your team. Do you want me to update? I don't mind 😄 

c4f66eabad1901ba5df52604b5be360a69608e4a should resolve the `.getAccountId()` issue. Thanks for the QA!

mehcode:
This was mostly good minus some nit picks but the existing examples have some of the same issues. This is definitely a large :+1: though to what we had ( and the PR was getting huge.. ) so merged. :smile: Thanks @Cooper-Kunz

Cooper-Kunz:
No problems at all! Thank you and @danielakhterov for the reviews and ensuring it's consistent. 🎊

PR #198, opened 2020-02-14
add fromPem() to Ed25519PrivateKey, modify fromBytes to accept 48 length Uint8Arrays

Sheng-Long:


src/crypto/Ed25519PrivateKey.ts
```diff
@@ -2,10 +2,37 @@ import * as nacl from "tweetnacl";
...
 import { Ed25519PublicKey } from "./Ed25519PublicKey";
 import { Mnemonic } from "./Mnemonic";
-import { decodeHex, deriveChildKey, ed25519PrivKeyPrefix, encodeHex, pbkdf2, randomBytes } from "./util";
+import { decodeHex, deriveChildKey, ed25519PrivKeyPrefix, encodeHex, pbkdf2, randomBytes, findSubarray, arraysEqual } from "./util";
 import { RawKeyPair } from "./RawKeyPair";
 import { createKeystore, loadKeystore } from "./Keystore";
 import { BadKeyError } from "../errors/BadKeyError";
+import { BadPemFileError } from "../errors/BadPemFileError";
+
+const privateKeyUint8Array: Uint8Array =
```
danielakhterov: Make sure this is "-----BEGIN PRIVATE KEY-----\n"
Also what about using "\n-----END PRIVATE KEY-----\n"

src/crypto/Ed25519PrivateKey.ts
```diff
@@ -204,4 +217,32 @@ export class Ed25519PrivateKey {
...
+     */
+    public static fromPem(pemArray: Uint8Array): Ed25519PrivateKey {
+        const index = findSubarray(pemArray, privateKeyUint8Array);
+
+        if (index === -1) {
+            throw new BadPemFileError();
+        }
+
+        const newPemArray = pemArray.slice(index);
+        const firstNLIndex = newPemArray.findIndex((el) => el === 10);
```
danielakhterov: Perhaps use `Uint8Array.indexOf()` method instead of `Uint8Array.findIndex()` because you're looking for a specific value.

src/crypto/util.ts
```diff
@@ -60,3 +60,30 @@ export function deriveChildKey(
 
     return { keyBytes: digest.subarray(0, 32), chainCode: digest.subarray(32) };
 }
+
+export function findSubarray(array: Uint8Array, subArray: Uint8Array): number {
```
danielakhterov: Add some tests for this, please.

src/errors/BadPemFileError.ts
```diff
@@ -0,0 +1,6 @@
+export class BadPemFileError extends Error {
+    public constructor() {
+        super("Failed to parse .pem file");
```
danielakhterov: Don't use `super(message)` use 
```typescript
super();
this.message = "blah";
```

src/crypto/Ed25519PrivateKey.ts
Sheng-Long: no need, our findSubarrray refactor and fixing the subarray definitions removes the need to find Indexes

src/errors/BadPemFileError.ts
Sheng-Long: changed for each *Error.ts file

src/crypto/util.ts
Sheng-Long: Added a test, do I need multiple?

PR #206, opened 2020-02-20
implement rudimentary DER decoding

abonander:
initial support for encrypted private keys in PEMs since we have to introspect the PKCS#8 `EncryptedPrivateKeyInfo` struct (an example of which I have added to the test)

### DIFF
--- a/src/crypto/der.ts
+++ b/src/crypto/der.ts
@@ -1,0 +1,118 @@
+export type AsnSeq = { seq: AsnType[] }
+export type AsnInt = { int: number }
+export type AsnBytes = { bytes: Uint8Array }
+export type AsnIdent = { ident: string }
+export type AsnNull = null;
+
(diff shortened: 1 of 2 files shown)


src/crypto/der.ts
```diff
@@ -0,0 +1,117 @@
...
+
+export function decodeIncremental(bytes: Uint8Array): [AsnType, Uint8Array] {
+    // slice off the initial tag byte, `decodeLength` returns a slice of the remaining data
+    const [ len, rem ] = decodeLength(bytes.subarray(1));
+    const data = rem.subarray(0, len);
+    const tail = rem.subarray(len);
+
+    switch (bytes[ 0 ]) {
+        case 2:
+            return [{ int: decodeInt(data) }, tail ];
```
abonander: The asymmetry in the spacing rules here really bugs me and there's no way to configure IntelliJ's autoformat to fix this.

abonander:
CC #208

PR #211, opened 2020-02-25
Mnemonic: provide `validate()` method

### LINKED ISSUE #210: Mnemonic constructor should ensure that words are in the BIP-39 standard list
Since generating a key from a mnemonic involves simply passing the phrase through PBKDF2, if an end-user typoes a word in their mnemonic phrase, it will still generate a valid key that they can then use to create an account which they will not likely be able to access again.

As a simple check in `Mnemonic`'s constructor, we can throw an error on any strings that don't appear in the BIP-39 English word list.

abonander:

TODO:
- [x] move validation to a `validate()` method as some clients may be using keys with "bad" mnemonics already
- [x] verify the checksum in the mnemonic: https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki#generating-the-mnemonic

closes #210


src/crypto/MnemonicError.ts
```diff
@@ -0,0 +1,12 @@
+export class MnemonicError extends Error {
```
mehcode: As we do `BadKeyError` maybe `BadMnemonicError` instead?

abonander: We went with `BadKeyError` instead of `KeyError` because the latter seemed too generic. I don't really feel the same way about `MnemonicError`, however it's also not exactly a hill I'm willing to die on.

PR #218, opened 2020-02-27
Make ContractFunctionResult.get* methods default to 0

Sheng-Long:


src/contract/ContractFunctionResult.ts
```diff
@@ -65,7 +65,11 @@ export class ContractFunctionResult {
         return this.bytes;
     }
 
-    public getString(index: number): string {
+    public getString(index?: number): string {
+        if (index == null) {
+            return Buffer.from(this.getBytes(0)).toString("utf-8");
+        }
+
         return Buffer.from(this.getBytes(index)).toString("utf-8");
```
mehcode: Do `index ?? 0` instead here

src/contract/ContractFunctionResult.ts
```diff
@@ -65,7 +65,11 @@ export class ContractFunctionResult {
         return this.bytes;
     }
 
-    public getString(index: number): string {
+    public getString(index?: number): string {
```
mehcode: We should make this change for all accessors

Sheng-Long: Should this be spun out as an issue, or should it be included in this pr?

mehcode: this PR, we shouldn't merge in a one-off change like this

PR #222, opened 2020-03-10
Use `@stablelib` for encoding and use `SubtleCrypto` if needed.

danielakhterov:


mehcode:
- [ ] Rename `Ed25519PrivateKey.derive` to `Ed25519PrivateKey.derive2` and document that it will eventually replace `derive` in a v2; then, use _just_ node-crypto or stablelibs in derive to avoid the async requirement
 - [ ] Replace `crypto.randomBytes` with tweetnacl
 - [ ] (On browser) Replace `crypto.pbkdf2` with SubtleCrypto or StableLibs ( depending on support for current browser )

Issue #265, opened 2020-08-03
Unused dependencies?


ChALkeR:
E.g. the following don't seem to be imported:

* `@improbable-eng/grpc-web-node-http-transport`
* `@stablelib/hash`
* `browser-headers`
* `fastestsmallesttextencoderdecoder`

PR #266, opened 2020-08-05
chore: remove unused dependencies

ChALkeR:
Fixes: https://github.com/hashgraph/hedera-sdk-js/issues/265

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,9 +50,7 @@
   "dependencies": {
     "@grpc/grpc-js": "^1.1.3",
     "@improbable-eng/grpc-web": "^0.13.0",
-    "@improbable-eng/grpc-web-node-http-transport": "^0.13.0",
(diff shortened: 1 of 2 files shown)


mehcode:
Can you rebase this so we can see if CI will pass with them gone?

ChALkeR:
Rebased.

ChALkeR:
@mehcode Seems like e2e tests require env configuration which is missing here, so they can't possibly pass.
Other tests are passing.

mehcode:
> Seems like e2e tests require env configuration which is missing here, so they can't possibly pass.
Other tests are passing.

Ah. That makes a lot of sense. They are using secret configuration and couldn't pass because this incoming branch is not in the same organization. We should probably separate out the CI. 

Thanks for the catch here regardless.

Issue #268, opened 2020-08-05
Consider removing usage of google/protobuf/wrappers and using an in-repo protocol for that


ChALkeR:
Currently, this library uses one proto file not present in the repo itself: `google/protobuf/wrappers`.
That protocol is completely trivial: https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/wrappers.proto

It's used directly as `google-protobuf/google/protobuf/wrappers_pb.js` in a number of places here, but there is a problem with that -- that file has been compiled in `commonjs` mode instead of `commonjs_strict`, which brings the same issues with globals pollution and CSP incompatibility. https://github.com/protocolbuffers/protobuf/issues/7778 is related.

#267 is related, and even after the fixes mentioned in #267, it would be still blocked against this one if `wrappers.proto` is not also recompiled in `commonjs_strict` mode.

A possible solution would be to replace all direct usage of that compiled js file with a proto file inside this repo, compiled in the same mode as all the other proto files. Perhaps even a subset of that file if not all imports are used.

mehcode:
It's a little unfortunate that the google provided library isn't compiled in `strict` mode but I agree a trivial fix is to inline the proto source.

mehcode:
Done in `develop`

Issue #267, opened 2020-08-05
Global scope pollution and CSP incompatibility


ChALkeR:
_Filing as a single issue because there is a clear path forward, equivalent for both issues, see below._

There are two major related problems with the current protobuf usage:

### Currently, this library pollutes global scope

To reproduce:
```console
> Object.keys(global)
[
  'global', 'clearInterval', 'clearTimeout', 'setInterval', 'setTimeout', 'queueMicrotask',
  'clearImmediate', 'setImmediate', 'proto'
]
> Object.keys(global.proto)
[
  'google', 'proto', 'Timestamp', 'TimestampSeconds', 'AccountID',
  'ContractID', 'CurrentAndNextFeeSchedule', 'FeeComponents',
  'FeeData', 'FeeSchedule', 'FileID', 'HederaFunctionality', 'Key',
  ...
```

This is problematic for a number of reasons, including — what if other library or a fork uses the exact same names for their protocols? (`Timestamp`/`ContractID` don't seem specific).

Also, this makes things bad even _without_ any other libraries that do that, because once this gets in a project, it would be very hard to tell in the future if there is a problem or not. The best way to eliminate such conflicts would be to allow _no_ library to pollute global scope.

### Content Security Policy incompatibility

Currently, this library uses the `Function` constructor which allows generate JS code dynamically in runtime, which requires a `unsafe-eval` CSP rule, which basically makes CSP useless, and CSP is a crucial part of web security, i.e. a safeguard against XSS vulnerabilities.

Afaik, this library uses it _just_ to get the `global` variable as `Function('return this')()` for the above case (to pollute globals), hence this is basically the same issue.

### Solution

Protobuf has a `commonjs_strict` mode as opposed to `commonjs`:

```diff
diff --git a/gulpfile.js b/gulpfile.js
index e04a4300..5e79d197 100644
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -43,7 +43,7 @@ gulp.task("build:proto", (cb) => {
     const cmd = "protoc " +
         `--plugin="protoc-gen-ts=${plugin}" ` +
         `--ts_out="service=grpc-web:${out}" ` +
-        `--js_out="import_style=commonjs,binary:${out}" ` +
+        `--js_out="import_style=commonjs_strict,binary:${out}" ` +
         `-I "${include}" ` +
         ` ${files}`;
```

Which brings the following differences in generated code (full diff on a single generated file, no need to apply manually):
```diff
[...]

PR #269, opened 2020-08-11
Build in commonjs_strict mode

ChALkeR:
`package` statements removed from protocols, since in strict mode the whole namespace is exported instead of the package, and keeping the packages would have required changes in all js files that import protocols.

Had to rename one `ConsensusService` -> `MirrorConsensusService`, as there were two different entities named `ConsensusService`.

This mostly solves #267, apart from #268. I.e. with this patch, the only thing that remains polluting the global scope is:
```js
> proto
{
  google: {
    protobuf: {
      BoolValue: [Function],
      BytesValue: [Function],
      DoubleValue: [Function],
      FloatValue: [Function],
      Int32Value: [Function],
      Int64Value: [Function],
      StringValue: [Function],
      UInt32Value: [Function],
      UInt64Value: [Function]
    }
  }
}
```
which comes from`google/protobuf/wrappers.proto` (see #268).

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -43,7 +43,7 @@ gulp.task("build:proto", (cb) => {
     const cmd = "protoc " +
         `--plugin="protoc-gen-ts=${plugin}" ` +
         `--ts_out="service=grpc-web:${out}" ` +
-        `--js_out="import_style=commonjs,binary:${out}" ` +
+        `--js_out="import_style=commonjs_strict,binary:${out}" ` +
         `-I "${include}" ` +
         ` ${files}`;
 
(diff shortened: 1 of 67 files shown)


mehcode:
Thanks for the change here. I completely agree that `commonjs_strict` is the right way to do this. I was unaware it existed myself.

mehcode:
Interestingly enough, this _did_ break everything. 

```
UNIMPLEMENTED: Method not found: CryptoService/cryptoGetBalance
```

All transactions report `UNIMPLEMENTED` now.

ChALkeR:
Hm. This is interesting, as this was loaded correctly and `test:unit` passed. I'll check what's going on there.

PR #272, opened 2020-08-13
Drop stablelib/base64 as there is already a decoder impl present here

ChALkeR:
There is already a slim decoder implementation here and there seems to be no reason for pulling in a custom base64 decoder when a two-liner with `atob`/`Buffer.from` works just as fine.

The point of `@stablelib/base64`, it seems, was to provide a decoder which delays decoding errors until the end so that it can be close to constant time on valid and invalid input of the same length. If that is needed here, this PR should be rejected (and perhaps instead the removed here code which is currently commented out should be fixed).

Refs: 87ed8eb9e72c4507a38280550753aa43bc2acd16

---

One thing that confuses me though: _what_ were those strings that `@stablelib/base64` failed to decode?
Buffer base64 decoder differs in these aspects:
 1. It supports both default ([RFC3548, section 3](https://tools.ietf.org/html/rfc3548#section-3) and url-safe ([RFC3548, section 4](https://tools.ietf.org/html/rfc3548#section-4)) base64 decoding.
 _Note: stablelib also supports both, but under separate functions, `decode` and `decodeURLSafe`._
 2. It skips garbage (in 7-bit plane), e.g. whitespace.

The potential problem is that if the reason why decoding with stablelib decoder failed in 87ed8eb9e72c4507a38280550753aa43bc2acd16, then it will also fail with atob which is used in browser version here.

Both of those could be solved with a trivial regex on `text` here: https://github.com/hashgraph/hedera-sdk-js/blob/cfb607bd49ee937ecea572ec7debf0bae610f6a9/src/encoding/base64.ts#L2-L4
replacing `-` with `+`, `_` with `/`, and whitespace (and/or other chars in 7-bit plane) with an empty string.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,6 @@
   "dependencies": {
     "@grpc/grpc-js": "^1.1.3",
     "@improbable-eng/grpc-web": "^0.13.0",
-    "@stablelib/base64": "^1.0.0",
     "@stablelib/hex": "^1.0.0",
     "@stablelib/hmac": "^1.0.0",
     "@stablelib/pbkdf2": "^1.0.0",
@@ -67,6 +66,7 @@
   },
   "devDependencies": {
     "@launchbadge/eslint-config": "^0.13.10",
+    "@stablelib/base64": "^1.0.0",
     "@types/jest": "^26.0.7",
     "@types/node": "^14.0.27",
     "eslint": "^7.5.0",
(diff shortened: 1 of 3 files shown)


mehcode:
Fair.

mehcode:
`master` is now failing the end-to-end tests in CI now but I think its unrelated. I'm going to ask someone from the SDK team to take a look.

PR #273, opened 2020-08-13
Remove untested stablelib crypto fallback

ChALkeR:
It was used just as a fallback in web browsers without `window.crypto.subtle`, and seems untested.

Moreover, the ecosystem usage of those packages seems low, and it doesn't seem like they are tested/reviewed enough:
1. All the downloads are coming from `@hashgraph/sdk` itself ([1](https://www.npmjs.com/browse/depended/@stablelib/pbkdf2), [2](https://www.npmjs.com/browse/depended/@stablelib/scrypt), also compare [@stablelib/pbkdf2](https://www.npmjs.com/package/@stablelib/pbkdf2) and [@hashgraph/sdk](https://www.npmjs.com/package/@hashgraph/sdk) downloads).
2. And `@hashgraph/sdk` doesn't even use those libs in node or modern browsers (as can be seen here).

Afaik, the same methods could be polyfilled using [crypto-browserify](https://www.npmjs.com/package/crypto-browserify) by downstream packages, if required.
Which is also how bundlers resolve `crypto` module already for browsers.

`@stablelib/sha384` is still used in one place.


mehcode:
> Afaik, the same methods could be polyfilled using crypto-browserify by downstream packages, if required.
Which is also how bundlers resolve crypto module already for browsers.

This is highly controversial and already removed in webpack 5 and not included at all in other bundlers. `crypto-browserify`, while far more used than `stablelibs`, is an ancient library with no security audits or even maintenance.  

---

To be honest, I'm pretty unhappy with `stablelibs` as well. In a v2 of the SDK, we plan to fully separate several pieces of the core into Node and Browser versions. The cryptography in Node will directly use the node `crypto` module and the cryptography in the browser will only use web crypto. 

There are maintained polyfills available for web crypto if an environment needs that.

---

With that said, I'd rather not remove this in v1 without strong motivation. It would be a breaking change because there might be people depending on it working. 

---

> Moreover, the ecosystem usage of those packages seems low, and it doesn't seem like they are tested/reviewed enough:

The tests in this package are quite lacking in a browser context, but there are extensive browser tests in myhbarwallet.com that uses this library.

mehcode:
I'm going to close this as its a breaking change that we can't directly accept. I think we're mostly on the same page for a v2, they just can't reasonably be removed in v1.

ChALkeR:
Yeah, this is definitely a breaking change so it could be only a part of semver-major.

> In a v2 of the SDK, we plan to fully separate several pieces of the core into Node and Browser versions. The cryptography in Node will directly use the node crypto module and the cryptography in the browser will only use web crypto.

This sounds great!

> `crypto-browserify`, while far more used than stablelibs, is an ancient library with no security audits or even maintenance.

That's not entirely correct, I think. `crypto-browserify` is rather small wrapper which imports implementations for different methods and re-exports them. E.g. last [pbkdf2](https://www.npmjs.com/package/pbkdf2) (gh: [crypto-browserify/pbkdf2](https://github.com/crypto-browserify/pbkdf2)) was published 2 months ago.

PR #277, opened 2020-08-20
Revert "Drop stablelib/base64 as there is already a decoder impl present here"

mehcode:
Reverts hashgraph/hedera-sdk-js#272


ChALkeR:
@mehcode Hm. Was there some kind of incompatibility there?

PR #300, opened 2020-11-20
Example creating a multi-sig account, and signing a transaction offline

gregscullard:
* creates a multi-sig account
* creates a crypto transfer transaction
* requests signatures from other key holders
* signs the transaction with all signatures
* executes


danielakhterov:
Also, thank you for this PR as well. Again, my apologies for not looking at it for this long.

Issue #332, opened 2021-02-06
Transaction.fromBytes error


cmdhema:
When run 'TokenAssociateTransaction.fromBytes(tx.toBytes())' got below error.

TypeError: Cannot read property 'tokens' of null
    at _fromProtobuf (/Users/kjwook/Desktop/kjwook/sources/adb-marketplace/node_modules/@hashgraph/sdk/lib/token/TokenAssociateTransaction.cjs:88:32)
    at Function.fromBytes (/Users/kjwook/Desktop/kjwook/sources/adb-marketplace/node_modules/@hashgraph/sdk/lib/transaction/Transaction.cjs:221:12)

andrix10:
Thanks for bringing this up, it has been fixed and will be reflected in the next patch.

PR #333, opened 2021-02-06
fix bug token transaction fromBytes error

cmdhema:
I raised issue 
https://github.com/hashgraph/hedera-sdk-js/issues/332

I think it is typo error. Maybe copy&paste.

Because All the transaction body implemented tokenCreate, so there are Transaction.fromBytes() error
Except TokenDissociateTransaction.js (It is well).


danielakhterov:
Thanks for the fix, should go out soon in the next update

PR #357, opened 2021-05-07
feat: implemented new test client environment

andrix10:


test/integration/client/index.js
```diff
@@ -2,61 +2,77 @@ import {
...
-export default async function newIntegrationClient(createNewAccount = false) {
-    let client;
-
-    if (
-        process.env.HEDERA_NETWORK != null &&
-        process.env.HEDERA_NETWORK == "previewnet"
-    ) {
-        client = Client.forPreviewnet();
-    } else {
+export default class newIntegrationTestEnv {
```
danielakhterov, line 13: This isn't Go :) You should use capital PascalCase for class names without the `new` prefix in the name. I recommend you do two things.
1) Create a `IntegrationTestEnv` class which is identical to what you have minus the name change.
2) Add an `async static new() { ... }` method that does all the initialization for you and returns a valid `IntegrationTestEnv`. This remove the need for `await client.initialize();` in each test case.

test/integration/AccountBalanceIntegrationTest.js
```diff
@@ -15,11 +15,13 @@ describe("AccountBalanceQuery", function () {
     it("account 0.0.3 should have a balance higher than 1 tinybar", async function () {
         this.timeout(15000);
 
-        const client = await newClient();
+        const client = new newClient();
```
danielakhterov, line 18: This should probably be called `env` or `testEnv` instead of `client`

danielakhterov (changes requested):

PR #366, opened 2021-05-10
Example scheduled transaction

derdis14:
When I was exploring the new scheduled transactions feature, I also looked at the existing SDK example "schedule-multi-sig-transaction.js". Unfortunately, in my opinion, the existing example is not very helpful and it also contains a bug. So I tried to create a better example. I hope you like it and consider merging it.


examples/scheduled-transaction-multi-sig-threshold.js
```diff
@@ -0,0 +1,109 @@
...
+  AccountBalanceQuery,
+  TransferTransaction,
+  ScheduleSignTransaction,
+  ScheduleInfoQuery,
+  TransactionRecordQuery,
+} = require("@hashgraph/sdk");
+require("dotenv").config();
+
+async function main() {
+  // set up operator
```
danielakhterov, line 16: Can you copy and paste the client setup from other examples. We want the examples to be able to be ran with _either_ `OPERATOR_KEY`, `OPERATOR_ID`, `HEDERA_NETWORK` environment variables, a `CONFIG_FILE` environment variable, or a mixture of both.

examples/scheduled-transaction-multi-sig-threshold.js
```diff
@@ -0,0 +1,109 @@
+const {
+  Client,
```
danielakhterov, line 2: Can you update this file to use 4 spaces at tab. We have an `.editorconfig` file at the root of the project, it might be easier for you to add some plugin to your editor that automatically reads this file, and reformat that way.

danielakhterov (changes requested):
Some small changes to make, but otherwise the PR LGTM.

derdis14:
Thank you for your feedback and explanations. I have updated the commit accordingly.

PR #378, opened 2021-05-15
fix: added missing module in example

derdis14:


danielakhterov:
Since you're already doing this, can you add this code to examples `consensus_pub_sub` and `get_account_balance` just for consistency.

PR #398, opened 2021-06-08
feat: use TLS to connect to Hedera network nodes where possible

mehcode:


src/channel/NodeChannel.js
```diff
@@ -13,15 +14,36 @@ export default class NodeChannel extends Channel {
...
         super();
 
+        let clientCredentials;
+        let clientAddress;
+
+        if (certificate != null) {
+            clientCredentials = credentials.createSsl(
+                Buffer.from(utf8.encode(certificate))
+            );
+            clientAddress = `${address.split(":")[0]}:50212`;
```
danielakhterov, line 29: I don't think we should be doing this. If someone is running their own version of the network, and they've enabled TLS on a different port then this will fail for no reason.

src/channel/NodeChannel.js
```diff
@@ -13,15 +14,36 @@ export default class NodeChannel extends Channel {
...
+        } else if (!address.endsWith(":50212")) {
+            clientCredentials = credentials.createInsecure();
+            clientAddress = `${address.split(":")[0]}:50211`;
```
danielakhterov, line 32: Same as before for non-TLS version. Specifically when I run the tests against Hedera Services master the network is `127.0.0.1:50213..4` , this would make those tests fail.

src/client/NodeClient.js
```diff
@@ -34,40 +35,40 @@ export const Network = {
...
-        "35.204.86.32:50211": new AccountId(12),
-        "35.234.132.107:50211": new AccountId(13),
-        "35.236.2.27:50211": new AccountId(14),
-        "35.228.11.53:50211": new AccountId(15),
-        "34.91.181.183:50211": new AccountId(16),
-        "34.86.212.247:50211": new AccountId(17),
-        "172.105.247.67:50211": new AccountId(18),
-        "34.89.87.138:50211": new AccountId(19),
-        "34.82.78.255:50211": new AccountId(20),
+        "35.237.200.180": new AccountId(3),
```
danielakhterov, line 38: Instead of having a port-less network, we should have two networks, one with TLS and one without. This way if someone calls `Client.useTls()` or whatever we end up calling it, we can simply call `setNetwork(MAINNET_TLS)`

PR #399, opened 2021-06-09
refactor: `TopicMessageQuery`

danielakhterov:


andrix10 (approved):

src/topic/TopicMessageQuery.js
```diff
@@ -191,83 +322,109 @@ export default class TopicMessageQuery {
...
+            this._counter += 1;
+
+            if (this._listener != null) {
+                this._listener(topicMessage);
+            } else {
+                throw new Error("(BUG) listener is unexpectedly not set");
+            }
+
+            if (this._limit != null && this._limit.comp(this._counter)) {
+                this._completionHandler();
```
steven-sheehy, line 424: This is incorrect. You need to hook up the completion handler to grpc's oncomplete or end signal that signifies the stream was completed successfully. This logic is wrong. It should not looked at grouped chunked messages but each individual consensus topci response. It also is complete when endTime is exceeded. But you shouldn't duplicate this logic that server already does. Just hook into the end signal.

src/topic/TopicMessageQuery.js
```diff
@@ -191,83 +322,109 @@ export default class TopicMessageQuery {
...
+            }
+        } catch (error) {
+            this._errorHandler(topicMessage, error);
```
steven-sheehy, line 427: I removed calling error handler in Java when the application onnext throws an error.

src/topic/TopicMessageQuery.js
```diff
@@ -191,83 +322,109 @@ export default class TopicMessageQuery {
...
-                        list.delete(transactionId);
-
-                        try {
-                            listener(topicMessage);
-                        } catch (error) {
-                            errorHandler(topicMessage, error);
-                        }
+                        setTimeout(() => {
+                            this._makeServerStreamRequest(client);
+                        }, 250 * 2 ** this._attempt);
```
steven-sheehy, line 391: This does not set a max backoff

src/topic/TopicMessageQuery.js
```diff
@@ -191,83 +322,109 @@ export default class TopicMessageQuery {
                 this._startTime != null ? this._startTime._toProtobuf() : null,
             consensusEndTime:
                 this._endTime != null ? this._endTime._toProtobuf() : null,
-            limit: this._limit != null ? this._limit : null,
+            limit: this._limit != null ? this._limit.sub(this._counter) : null,
```
steven-sheehy, line 325: This does not update the starttime from the last message received for retries. It does not even store the last message.

src/topic/TopicMessageQuery.js
```diff
@@ -191,83 +322,109 @@ export default class TopicMessageQuery {
...
+                        this._attempt += 1;
 
-                        list.delete(transactionId);
-
-                        try {
-                            listener(topicMessage);
-                        } catch (error) {
-                            errorHandler(topicMessage, error);
-                        }
+                        setTimeout(() => {
```
steven-sheehy, line 389: This doesn't produce any log output that a retry is occurring.

steven-sheehy (changes requested):
There's a lot of bugs with your refactoring and a lot of the items I addressed in Java were not addressed here in Javascript. Please test the scenarios listed in the original [ticket](https://github.com/hashgraph/hedera-sdk-java/issues/502) with your implementation.

src/topic/TopicMessageQuery.js
mehcode: This was reverted. It's a large breaking change.

src/topic/TopicMessageQuery.js
danielakhterov: I completely agree that the SDK shouldn't be trying to replicate something that the mirror already does, but from my testing the gRPC library we use calls the `end` callback for errors and the completion of the stream. This was a bit confusing, but I've tested this a bit more. I've just did some more testing, and apparently I need to branch within `on("status")` rather than connecting the completion handler to `.on("end")`.

src/topic/TopicMessageQuery.js
danielakhterov: You're right, this should keep track of `startTime` of the last message, but I don't see why we should keep the last message completely. From what I can tell all that matters is the `startTime` and making sure we decrement `limit` when needed.

src/topic/TopicMessageQuery.js
```diff
@@ -191,83 +322,109 @@ export default class TopicMessageQuery {
...
+    }
 
-        handle._setCall(() => cancel());
+    /**
+     * @private
+     * @param {TopicMessage} topicMessage
+     */
+    _passTopicMessage(topicMessage) {
+        try {
+            this._counter += 1;
```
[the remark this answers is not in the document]
steven-sheehy: The mirror node doesn't care about chunks besides storing the fields in the database. It's purely a client side concern. Also, the mirror node APIs are stateless and it's up to the client to track their progress. Similar to how Kafka topics work.

src/topic/TopicMessageQuery.js
steven-sheehy: True, you only need consensus start time

PR #400, opened 2021-06-09
fix: string to entity conversion inconsistencies

### LINKED ISSUE #397: String to entity conversion inconsistencies
Using v 2.03 of the SDK I experienced inconsistent handling of entities by the SDK. 

For instance, when importing an account ID from a `.env` file: 

`const myAccountId = process.env.MY_ACCOUNT_ID`

And then using `myAccountId` to configure the client:

`client.setOperator(myAccountId, myPrivateKey);`

`setOperator` seems to convert the string to an Account ID entity. 

However, when performing other calls, specifically `TopicMessageSubmitTransaction()` the SDK manages it differently; producing a type error when passing a string into the following: 

`const myTopicId = process.env.MY_TOPIC_ID`

```
const MessageSubmit = await new TopicMessageSubmitTransaction({
      topicId: myTopicId,
      message: "Hello there!",
    }).execute(client);
  }
```

This produces a type error while I anticipated the string to entity conversion being consistent across all SDK calls.

andrix10:

closes #397

### DIFF
--- a/src/account/AccountBalanceQuery.js
+++ b/src/account/AccountBalanceQuery.js
@@ -120,9 +120,9 @@ export default class AccountBalanceQuery extends Query {
      */
     setContractId(contractId) {
         this._contractId =
-            contractId instanceof ContractId
-                ? contractId
-                : ContractId.fromString(contractId);
+            typeof contractId === "string"
+                ? ContractId.fromString(contractId)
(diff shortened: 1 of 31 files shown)


danielakhterov:
Can you update the constructors to support `ScheduleId | string` as well.

danielakhterov (approved):

PR #407, opened 2021-06-11
Fix: `TopicMessageQuery` refactor issues

danielakhterov:


cacampbell (approved):

danielakhterov:
Merge as this is an important fix that needs to be released ASAP. However, I'd still like any reviewers to thoroughly review my changes in case I've made more silly mistakes.
Note: CI is failing on node, but passing on chrome. I've ran locally and it seems to be passing as well. I'm guessing simply increasing the timeout for the particular test case that is failing should resolve the issue, but that's for another PR.

src/topic/TopicMessageQuery.js
```diff
@@ -332,6 +341,16 @@ export default class TopicMessageQuery {
                 (data) => {
                     const message = proto.ConsensusTopicResponse.decode(data);
 
+                    if (this._limit != null && this._limit.gt(0)) {
+                        this._limit = this._limit.sub(1);
+                    }
+
+                    this._startTime = Timestamp._fromProtobuf(
```
steven-sheehy, line 348: Need to add one nanosecond

src/topic/TopicMessageQuery.js
```diff
@@ -388,9 +403,10 @@ export default class TopicMessageQuery {
 
                         setTimeout(() => {
```
steven-sheehy, line 404: There's still no log that indicates a retry is in progress

src/topic/TopicMessageQuery.js
danielakhterov: Damn, good catch. Fixed this in https://github.com/hashgraph/hedera-sdk-js/pull/411

src/topic/TopicMessageQuery.js
danielakhterov: I believe I also forgot to add logging in Go; logging just isn't second nature to me yet. Fixed here though https://github.com/hashgraph/hedera-sdk-js/pull/411

PR #411, opened 2021-06-16
fix: `TopicMessageQuery`

danielakhterov:
Make sure `startTime` is one nanosecond in the future when saving the `consensusTimestamp` of a message.
Add a log statement when retrying `TopicMessageQuery`

### DIFF
--- a/src/Timestamp.js
+++ b/src/Timestamp.js
@@ -5,6 +5,8 @@ import Long from "long";
  * @typedef {import("@hashgraph/proto").ITimestamp} proto.ITimestamp
  */
 
+const MAX_NS = Long.fromNumber(999999999);
+
 export default class Timestamp {
     /**
      * @param {Long | number} seconds
@@ -74,6 +76,16 @@ export default class Timestamp {
             this.seconds.toInt() * 1000 +
                 Math.floor(this.nanos.toInt() / 1000000)
         );
+    }
+
+    /**
+     * @param {Long | number} nanos
+     * @returns {Timestamp}
+     */
+    plusNanos(nanos) {
+        const ns = this.nanos.add(nanos);
+
+        return new Timestamp(this.seconds.add(ns.div(MAX_NS)), ns.mod(MAX_NS));
     }
 
     /**
--- a/src/Timestamp.js
+++ b/src/Timestamp.js
@@ -5,7 +5,7 @@ import Long from "long";
  * @typedef {import("@hashgraph/proto").ITimestamp} proto.ITimestamp
  */
 
-const MAX_NS = Long.fromNumber(999999999);
+const MAX_NS = Long.fromNumber(1000000000);
 
 export default class Timestamp {
     /**
--- a/src/topic/TopicMessageQuery.js
+++ b/src/topic/TopicMessageQuery.js
@@ -347,18 +347,11 @@ export default class TopicMessageQuery {
                         this._limit = this._limit.sub(1);
                     }
 
-                    const consensusTimestamp = /** @type {proto.ITimestamp} */ (
-                        message.consensusTimestamp
-                    );
-
-                    const nanos = /** @type {number} */ (
-                        consensusTimestamp.nanos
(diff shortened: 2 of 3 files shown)


src/topic/TopicMessageQuery.js
```diff
@@ -401,11 +408,25 @@ export default class TopicMessageQuery {
                         this._attempt < this._maxAttempts &&
                         this._retryHandler(error)
                     ) {
+                        const delay = Math.min(
+                            250 * 2 ** this._attempt,
+                            this._maxBackoff
+                        );
+                        console.log(
```
steven-sheehy, line 415: This should probably be a `console.error()`. If you were using a logging framework it would be a warn level.

src/topic/TopicMessageQuery.js
```diff
@@ -401,11 +408,25 @@ export default class TopicMessageQuery {
...
+                            `Error subscribing to topic ${
+                                this._topicId != null
+                                    ? this._topicId.toString()
+                                    : "UNKNOWN"
+                            } during attempt ${
+                                this._attempt
+                            }. Waiting ${delay} ms before next attempt`
```
steven-sheehy, line 422: We should add the error message (but not stacktrace) at the end like we do for Java.

steven-sheehy (changes requested):

src/topic/TopicMessageQuery.js
danielakhterov: That sounds like it should be `console.warn()` then; on another note we should really update to use a proper logging library.

src/topic/TopicMessageQuery.js
```diff
@@ -347,12 +347,19 @@ export default class TopicMessageQuery {
...
                     );
 
+                    const nanos = /** @type {number} */ (
+                        consensusTimestamp.nanos
+                    );
+
+                    this._startTime = Timestamp._fromProtobuf({
+                        seconds: consensusTimestamp.seconds,
+                        nanos: nanos + 1,
+                    });
```
steven-sheehy, line 361: This is quite a verbose way to increment a timestamp. But more importantly, as Xin noted to me it can overflow the nanos when nanos is `999,999,999` and not increment the seconds. Recommend you create a helper method in `Timestamp.plusNanos()` similar to how `java.time.Instant.plusNanos()` works that handles nanos larger than 9 digits.

```javascript
this._startTime = Timestamp._fromProtobuf((message.consensusTimestamp)).plusNanos(1)
```

danielakhterov: I think it's fair for the SDK to implement `Timestamp.plusNanos()`, but one questions comes to mind is how do the services work here? If I pass nanos > 999,999,999, does Hedera err, cap the nanos, or automatically combine nanos and seconds? Perhaps it shouldn't matter what the SDK does if the services already fix?

steven-sheehy: HAPI assigns the consensus timestamp, so it will ensure it's with 0-999999999. It's not user input to HAPI. But it doesn't matter if the server validates it or not as this is not user supplied input but SDK calculated input. Since the SDK is wanting to increment by one nanos it should handle setting a correct value as per the contract that there's always at most 1B nanos in a second.

src/Timestamp.js
```diff
@@ -5,6 +5,8 @@ import Long from "long";
  * @typedef {import("@hashgraph/proto").ITimestamp} proto.ITimestamp
  */
 
+const MAX_NS = Long.fromNumber(999999999);
```
steven-sheehy, line 8: You have an off by one error. If the timestamp is sec=0,ns=999999998, incrementing by one nano should give you sec=0,ns=999999999 but it gives you sec=1,ns=0. Please add a test so you can verify.

danielakhterov: You're right, fixed and added regression test.

steven-sheehy (approved):
LGTM

PR #416, opened 2021-06-18
feat: adding missing TokenInfo andTopicInfo [To|From]Bytes

andrix10:


src/token/TokenInfo.js
```diff
@@ -306,4 +308,22 @@ export default class TokenInfo {
...
+    static fromBytes(bytes) {
+        return TokenInfo._fromProtobuf(prot.TokenInfo.decode(bytes));
+    }
+
+
+    /**
+     * @returns {Uint8Array}
+     */
+    toBytes() {
+        return prot.TokenInfo.encode(this._toProtobuf()).finish();
```
danielakhterov, line 325: Can you make this `proto` not `prot` to be consistent with the rest of the SDK?

src/topic/TopicInfo.js
```diff
@@ -172,4 +174,23 @@ export default class TopicInfo {
...
         };
     }
+
+    /**
+     * @param {Uint8Array} bytes
+     * @returns {TopicInfo}
+     */
+    static fromBytes(bytes) {
+        const temp = {
+            topicInfo: prot.ConsensusTopicInfo.decode(bytes)
```
danielakhterov, line 184: Can you make this `proto` to be consistent with the rest of the SDK?

danielakhterov (changes requested):
Can you also add these changes to the changelog
```markdown
## Unreleased

### Added

 * Implemented to and from bytes for `TopicInfo` and `TokenInfo`
```

src/topic/TopicInfo.js
```diff
@@ -178,4 +172,23 @@ export default class TopicInfo {
...
+
+    /**
+     * @param {Uint8Array} bytes
+     * @returns {TopicInfo}
+     */
+    static fromBytes(bytes) {
+        const temp = {
+            topicInfo: proto.ConsensusTopicInfo.decode(bytes)
+        }
+        return TopicInfo._fromProtobuf(/** @type {proto.IConsensusGetTopicInfoResponse} */ (temp));
```
danielakhterov, line 184: Why don't you just do instead of `temp`
```javascript
return TopicInfo._fromProtogbuf({ topicInfo: proto.ConsensusTopicInfo.decode(bytes) });`
```

danielakhterov (changes requested):

danielakhterov (approved):

PR #421, opened 2021-06-24
fix: add basic client side TLS support for mirror node connections

danielakhterov:
Closes: #423


danielakhterov:
Test is not failing due to any change in this PR.

PR #430, opened 2021-06-29
Update retry handler's to retry on gRPC error `INTERNAL` only if it matches `RST_STREAM` regex

danielakhterov:
Resolves: #430 

Warning: This is untested since all the nodes seem to work fine right now.

### DIFF
--- a/src/Executable.js
+++ b/src/Executable.js
@@ -16,7 +16,7 @@ export const ExecutionState = {
     Error: "Error",
 };
 
-export const RST_STREAM = new RegExp(".*RST_STREAM.*");
+export const RST_STREAM = /\brst[^0-9a-zA-Z]stream\b/gi;
 
 /**
  * @abstract
--- a/src/Executable.js
+++ b/src/Executable.js
@@ -16,6 +16,8 @@ export const ExecutionState = {
     Error: "Error",
 };
 
+export const RST_STREAM = new RegExp(".*RST_STREAM.*");
+
 /**
  * @abstract
  * @internal
@@ -193,7 +195,8 @@ export default class Executable {
         return (
             error.status === GrpcStatus.Unavailable ||
             error.status === GrpcStatus.ResourceExhausted ||
-            error.status === GrpcStatus.Internal
+            (error.status === GrpcStatus.Internal &&
+                RST_STREAM.test(error.message))
         );
     }
 
@@ -237,9 +240,11 @@ export default class Executable {
             try {
                 response = await this._execute(channel, request);
             } catch (err) {
+                const error = GrpcServiceError._fromResponse(err);
+
                 if (
-                    err instanceof GrpcServiceError &&
-                    this._shouldRetryExceptionally(err) &&
+                    error instanceof GrpcServiceError &&
+                    this._shouldRetryExceptionally(error) &&
                     attempt <= this._maxRetries
                 ) {
                     node.increaseDelay();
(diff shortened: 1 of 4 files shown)


src/topic/TopicMessageQuery.js
```diff
@@ -115,16 +116,18 @@ export default class TopicMessageQuery {
...
                     );
                 } else {
                     switch (error.code) {
-                        // NOT_FOUND
-                        // eslint-disable-next-line no-fallthrough
-                        case 5:
                         // RESOURCE_EXHAUSTED
                         // eslint-disable-next-line no-fallthrough
                         case 8:
+                            return RST_STREAM.test(error.details.toString());
```
steven-sheehy, line 127: This should be for case 13 INTERNAL

src/topic/TopicMessageQuery.js
```diff
@@ -115,16 +116,18 @@ export default class TopicMessageQuery {
                         error.toString().includes("NOT_FOUND") ||
                         error.toString().includes("UNAVAILABLE") ||
                         error.toString().includes("RESOURCE_EXHAUSTED") ||
-                        error.toString().includes("INTERNAL")
+                        (error.toString().includes("INTERNAL") &&
```
steven-sheehy, line 119: Why would there be a non-gRPC `StatusObject` that includes gRPC status codes? It seems like errors in this if statement would more likely to be lower level HTTP2 errors, right? In which case you might want to just retry all if so.

steven-sheehy (changes requested):

steven-sheehy:
> Warning: This is untested since all the nodes seem to work fine right now.

I'm not sure how you can say that. Testnet node 0.0.6 has this issue right now. Your CI workflow is failing right now with this exact error. It seems like that regex is not enough to catch the proxy issue. You can probably make it more generic and just search for `\brst[^0-9a-zA-Z]stream\b`.

src/topic/TopicMessageQuery.js
mehcode: This really isn't part of this PR but I do agree that this whole part of the default retry handler should likely be thrown out and we should test for a "gRPC error" more reliably somehow.

src/Executable.js
```diff
@@ -16,6 +16,8 @@ export const ExecutionState = {
     Error: "Error",
 };
 
+export const RST_STREAM = new RegExp(".*RST_STREAM.*");
```
mehcode, line 19: You want to use literals here - `/foo/u` - in JS.

src/Executable.js
```diff
@@ -193,7 +195,8 @@ export default class Executable {
         return (
             error.status === GrpcStatus.Unavailable ||
             error.status === GrpcStatus.ResourceExhausted ||
-            error.status === GrpcStatus.Internal
+            (error.status === GrpcStatus.Internal &&
+                RST_STREAM.test(error.message))
```
steven-sheehy, line 199: Should have a unit test that at least tests the retry handler with various errors to ensure it produces the desired boolean.

steven-sheehy (changes requested):

src/topic/TopicMessageQuery.js
danielakhterov: Not entirely sure on what you're suggesting. I did remove that block of code and just `return true`, but I could be misunderstanding what you want me to do here.

steven-sheehy (approved):

PR #441, opened 2021-07-10
fix: legacy derive, and mnemonic 3derive, added tests

andrix10:

Closes: #442

### DIFF
--- a/src/crypto/util.ts
+++ b/src/crypto/util.ts
@@ -27,8 +27,13 @@ export function legacyDeriveChildKey(
         password.byteOffset + seed.length,
         8
     );
-    view.setInt32(0, index);
-    view.setInt32(4, index >= 0 ? 0 : 0xFF);
+    if (index === 0xFFFFFFFFFF) {
+        view.setInt32(0, 0xFF);
+        view.setInt32(4, index >>> 32);
+    } else {
+        view.setInt32(0, index < 0 ? -1 : 0);
+        view.setInt32(4, index);
+    }
 
     const salt = Uint8Array.from([ 0xFF ]);
     return Pbkdf2.deriveKey(
--- a/__tests__/unit/Mnemonic.test.ts
+++ b/__tests__/unit/Mnemonic.test.ts
@@ -110,37 +110,35 @@ describe("Mnemonic", () => {
         expect(result.isOk()).toBeFalsy();
     });
 
-    it("legacy2 mnemonic should work", async () => {
-        const words = [
-            "obvious",
-            "favorite",
-            "remain",
-            "caution",
-            "remove",
-            "laptop",
-            "base",
-            "vacant",
-            "increase",
-            "video",
-            "erase",
-            "pass",
-            "sniff",
-            "sausage",
-            "knock",
-            "grid",
-            "argue",
-            "salt",
-            "romance",
-            "way",
-            "alone",
-            "fever",
-            "slush",
-            "dune",
-        ];
+    it("should match MyHbarWallet v1", async() => {
+        const mnemonic = await Mnemonic.fromString("jolly kidnap Tom lawn drunk chick optic lust mutter mole bride galley dense member sage neural widow decide curb aboard margin manure");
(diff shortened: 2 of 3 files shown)


src/crypto/util.ts
```diff
@@ -27,8 +27,13 @@ export function legacyDeriveChildKey(
         password.byteOffset + seed.length,
         8
     );
-    view.setInt32(0, index);
-    view.setInt32(4, index >= 0 ? 0 : 0xFF);
+    if (index === 0xFFFFFFFFFF) {
+        view.setInt32(0, 0xFF);
+        view.setInt32(4, index >>> 32);
```
danielakhterov, line 32: I don't think this does what you think it does. `index >>> 32` would be `0xFF`. If `index` is `0xFFFFFFFFFF` then why not just set the value directly `view.setInt32(4, 0xFFFFFFFF)`

src/crypto/util.ts
```diff
@@ -27,8 +27,13 @@ export function legacyDeriveChildKey(
         password.byteOffset + seed.length,
         8
     );
-    view.setInt32(0, index);
-    view.setInt32(4, index >= 0 ? 0 : 0xFF);
+    if (index === 0xFFFFFFFFFF) {
```
danielakhterov, line 30: After the discussion we had on Slack, can you update this to
1) Hardcode `0xFF_FFFFFFFF` when the `index` is exactly that instead of using `index >>> 32`
2) Add a if branch where if `index > 0xFF_FFFFFFFF` we throw an error saying we don't support indexes of that size.

danielakhterov (approved):

src/crypto/util.ts
ChALkeR: `index >>> 32` would be `0xFFFFFFFF` here.

1. Shifts operate on 32-bit numbers
2. Only 5 bits are effective, so `>>> 32` is equivalent to  `>>> 0`.

This way of writing down `0xFFFFFFFF` is confusing, as that's not what one might expect for `0xFFFFFFFFFF >>> 32` to be.

Refs: https://262.ecma-international.org/11.0/#sec-numeric-types-number-unsignedRightShift
Refs: https://262.ecma-international.org/11.0/#table-numeric-type-ops


@danielakhterov @andrix10 this likely should be fixed somehow, I just don't know what's the correct value there.

ChALkeR: Which is the same as `view.setInt32(4, -1);` btw.

danielakhterov: @ChALkeR I forget when, but we did eventually stumble onto this issue when writing the v2 SDK (this PR is for v1). The current implementation in v2 is
```javascript
    if (index === 0xffffffffff) {
        view.setInt32(seed.length + 0, 0xff);
        view.setInt32(seed.length + 4, index >>> 32);
    } else {
        view.setInt32(seed.length + 0, index < 0 ? -1 : 0);
        view.setInt32(seed.length + 4, index);
    }
```
which tries to maintain backwards compatibility. Also, this is _legacy_ derive. Users are encouraged not to use these methods unless they created keys using the old v1 SDK or created keys using Hedera's legacy mobile application.

ChALkeR: @danielakhterov Yes, this is the same code chunk.
For `index === 0xffffffffff`, `index >>> 32` is `0xffffffff`, which is `-1`.
Not `0xff`, as one might think while reading this code.

danielakhterov: So you're suggesting we use `-1` instead of `index >>> 32` since it's more clear with `-1`? That sounds reasonable to me.

ChALkeR: Yes.

danielakhterov: Can you make a PR for this?

ChALkeR: I will. https://github.com/hashgraph/hedera-sdk-js/pull/784 is more important though I think )

ChALkeR: Filed https://github.com/hashgraph/hedera-sdk-js/pull/804

PR #467, opened 2021-07-23
feat: reversible Hbar.[from|to]String()

andvara-naut:

**Problem**
Hbar.toString() produces a string, but you cannot use that string within Hbar.fromString().

**Solution**
Update Hbar.fromString() to correctly decode the HbarUnit type from the string and use that for the Hbar construction.

Use: https://regex101.com/r/fKXLJT/1


src/Hbar.js
```diff
@@ -65,7 +65,37 @@ export default class Hbar {
      * @returns {Hbar}
      */
     static fromString(str, unit = HbarUnit.Hbar) {
-        return new Hbar(new BigNumber(str), unit);
+        try {
+            let [amount, symbol] = str.split(" ");
+            switch (symbol) {
```
danielakhterov, line 70: Can you move the parsing of `HbarUnit` into `HbarUnit.fromString()`

danielakhterov (changes requested):

test/unit/Hbar.js
```diff
@@ -18,4 +18,84 @@ describe("Hbar", function () {
 
         expect(tinybar).to.eql(hbar.toTinybars().toString());
     });
+
```
danielakhterov, line 21: Can you add this:
```javascript
    it("[from|to]String() should be reversable", function () {
        expect(Hbar.fromString("1").toTinybars().toString()).to.be.equal("100000000");
        expect(Hbar.fromString("1.151").toTinybars().toString()).to.be.equal("115100000");
        expect(Hbar.fromString("1 tℏ").toTinybars().toString()).to.be.equal("1");
        expect(Hbar.fromString("1 μℏ").toTinybars().toString()).to.be.equal("100");
        expect(Hbar.fromString("1 mℏ").toTinybars().toString()).to.be.equal("100000");
        expect(Hbar.fromString("1 ℏ").toTinybars().toString()).to.be.equal("100000000");
        expect(Hbar.fromString("1 kℏ").toTinybars().toString()).to.be.equal("100000000000");
        expect(Hbar.fromString("1 Mℏ").toTinybars().toString()).to.be.equal("100000000000000");
        expect(Hbar.fromString("1 Gℏ").toTinybars().toString()).to.be.equal("100000000000000000");
        expect(Hbar.fromString("1.151 mℏ").toTinybars().toString()).to.be.equal("115100");
        expect(Hbar.fromString("1.151 ℏ").toTinybars().toString()).to.be.equal("115100000");
        expect(Hbar.fromString("1.151 kℏ").toTinybars().toString()).to.be.equal("115100000000");
        expect(Hbar.fromString("1.151 Mℏ").toTinybars().toString()).to.be.equal("115100000000000");
        expect(Hbar.fromString("1.151 Gℏ").toTinybars().toString()).to.be.equal("115100000000000000");
        expect(Hbar.fromString("+1").toTinybars().toString()).to.be.equal("100000000");
        expect(Hbar.fromString("+1.151").toTinybars().toString()).to.be.equal("115100000");
        expect(Hbar.fromString("+1 Gℏ").toTinybars().toString()).to.be.equal("100000000000000000");
        expect(Hbar.fromString("+1.151 Gℏ").toTinybars().toString()).to.be.equal("115100000000000000");
        expect(Hbar.fromString("-1").toTinybars().toString()).to.be.equal("-100000000");
        expect(Hbar.fromString("-1.151").toTinybars().toString()).to.be.equal("-115100000");
        expect(Hbar.fromString("-1 Gℏ").toTinybars().toString()).to.be.equal("-100000000000000000");
        expect(Hbar.fromString("-1.151 Gℏ").toTinybars().toString()).to.be.equal("-115100000000000000");

        let err = false;

        try {
            Hbar.fromString("1.151 tℏ");
        } catch (error) {
            err = error.toString().includes("Hbar in tinybars contains decimals");
        }

        if (!err) {
            throw new Error("`Hbar.fromString()` did not error");
        }

        try {
            Hbar.fromString("1.151 μℏ");
        } catch (error) {
            err = error.toString().includes("Hbar in tinybars contains decimals");
        }

        if (!err) {
            throw new Error("`Hbar.fromString()` did not error");
        }

        try {
            Hbar.fromString("1.151.")
        } catch (error) {
            err = error.toString().includes("invalid string provided");
        }

        if (!err) {
            throw new Error("`Hbar.fromString()` did not error");
        }

        try {
            Hbar.fromString("1.151 uℏ")
        } catch (error) {
            err = error.toString().includes("invalid string provided");
        }

        if (!err) {
            throw new Error("`Hbar.fromString()` did not error");
        }
    });
```

test/unit/Hbar.js
```diff
@@ -18,4 +18,84 @@ describe("Hbar", function () {
...
+    it("should reverse Hbar units using [to|from]String()", function () {
+        let hbar = new Hbar("100.00", HbarUnit.Hbar);
+        let toString = String;
```
danielakhterov, line 24: Not sure why you're assigning `String` here. Should simply do `let toString`;

test/unit/Hbar.js
```diff
@@ -18,4 +18,84 @@ describe("Hbar", function () {
...
+        fromString = Hbar.fromString(toString).toString();
+        expect(toString).to.eql(fromString);
+    });
+
+    it("should not pass regex check", function () {
+        let check = String;
+
+        check = "";
+        let fromString = Hbar.fromString(check);
+        console.log(fromString)
```
danielakhterov, line 97: I think you forgot to remove this.

src/Hbar.js
```diff
@@ -65,7 +65,13 @@ export default class Hbar {
      * @returns {Hbar}
      */
     static fromString(str, unit = HbarUnit.Hbar) {
-        return new Hbar(new BigNumber(str), unit);
+        const pattern = /^((?:\+|-)?\d+(?:\.\d+)?)(?: (tℏ|μℏ|mℏ|ℏ|kℏ|Mℏ|Gℏ))?$/;
+        if (pattern.test(str)) {
+            let [amount, symbol] = str.split(" ");
+            return new Hbar(new BigNumber(amount), HbarUnit.fromString(symbol));
+        } else {
+            throw new Error("Variable failed regex: "+str);
```
danielakhterov, line 73: Be sure to run `pnpm format` and `pnpm lint`

src/Hbar.js
```diff
@@ -65,7 +65,13 @@ export default class Hbar {
      * @returns {Hbar}
      */
     static fromString(str, unit = HbarUnit.Hbar) {
-        return new Hbar(new BigNumber(str), unit);
+        const pattern = /^((?:\+|-)?\d+(?:\.\d+)?)(?: (tℏ|μℏ|mℏ|ℏ|kℏ|Mℏ|Gℏ))?$/;
+        if (pattern.test(str)) {
+            let [amount, symbol] = str.split(" ");
+            return new Hbar(new BigNumber(amount), HbarUnit.fromString(symbol));
```
danielakhterov, line 71: Do something like ```
        let [amount, symbol] = str.split(" ");

        if (symbol != null) {
            unit = HbarUnit.fromString(symbol);
        }

        return new Hbar(new BigNumber(amount), unit);
``` to maintain backwards compatibility with `unit`.

File: test/unit/Hbar.js (hunk 2 above)
Comment (danielakhterov) [reply test/unit/Hbar.js:21]: Not exactly related to `Hbar.[from|to]String()`, but can you add
```
        if (!this._valueInTinybar.isInteger()) {
            throw new Error("Hbar in tinybars contains decimals");
        }
``` to the end of the `Hbar` constructor.
This is part of the test I added above.

File: src/Hbar.js (hunk 5 above)
Comment (danielakhterov) [inline src/Hbar.js:73]: For the error just make it a string constant; something like `invalid string provided` or `invalid argument` or something else if you think of one better.

File: test/unit/Hbar.js [hunk 7]
```diff
@@ -18,4 +18,84 @@ describe("Hbar", function () {
...
+        let toString = String;
+        let fromString = String;
+
+        toString = gigabar.toString();
+        fromString = Hbar.fromString(toString).toString();
+        expect(toString).to.eql(fromString);
+    });
+
+    it("should not pass regex check", function () {
+        let check = String;
```
danielakhterov, line 93: If you're going to set `check` to `""` right after the declaration you might as well set it here
`let check = ""`

test/unit/Hbar.js
```diff
@@ -18,4 +18,84 @@ describe("Hbar", function () {
 
         expect(tinybar).to.eql(hbar.toTinybars().toString());
     });
+
+    it("should reverse Hbar units using [to|from]String()", function () {
+        let hbar = new Hbar("100.00", HbarUnit.Hbar);
+        let toString = String;
+        let fromString = String;
```
danielakhterov, line 25: I'm not sure if this is the case, but you setting the value equal to `String` is not the same thing as defining the type of this variable. To define a variable of type `string | undefined` you'd do 
```javascript
/** @type {string=} */
let fromString;
```

danielakhterov (changes requested):

src/HbarUnit.js
```diff
@@ -28,6 +28,31 @@ export default class HbarUnit {
...
+            case HbarUnit.Millibar._symbol:
+                return HbarUnit.Millibar;
+            case HbarUnit.Kilobar._symbol:
+                return HbarUnit.Kilobar;
+            case HbarUnit.Megabar._symbol:
+                return HbarUnit.Megabar;
+            case HbarUnit.Gigabar._symbol:
+                return HbarUnit.Gigabar;
+            default:
+                return HbarUnit.Hbar;
```
danielakhterov, line 53: We should be error if the symbol is unrecognized.

test/unit/Hbar.js
```diff
@@ -2,20 +2,169 @@ import Hbar from "../src/Hbar.js";
...
+        for (let j = 0; j < check.length; j++) {
+            try {
+                Hbar.fromString(check[j]);
+            } catch (error) {
+                if (error.toString().includes("invalid argument provided")) {
+                    k++;
+                }
+            }
+        }
+        expect(k).to.equal(i);
```
danielakhterov, line 69: Although this does check to make sure all these test cases fail, we probably should show which one passed, but was expected to fail.

test/unit/Hbar.js
```diff
@@ -2,20 +2,169 @@ import Hbar from "../src/Hbar.js";
...
+            "1.151 h",
+        ];
+
+        let i = check.length;
+
+        for (let j = 0; j < check.length; j++) {
+            try {
+                Hbar.fromString(check[j]);
+            } catch (error) {
+                if (error.toString().includes("invalid argument provided")) {
```
danielakhterov, line 64: Perhaps change this to simply be `expect(error.toString().includes("invalid argument provided")).to.be.true;` You'd then be able to remove `i`, `j`, and `k` completely and use a for each loop instead.

test/unit/Hbar.js
```diff
@@ -2,20 +2,169 @@ import Hbar from "../src/Hbar.js";
...
 
-    it("should stringify to Microbar", function () {
+    it("should append default unit Hbar", function () {
+        /**
+         * fromString strips + and should append the default Hbar unit when none are present
+         */
+        const unit = HbarUnit.Hbar._symbol;
+        let check = ["1", "-1", "+1", "1.151", "-1.151", "+1.151"];
+
+        for (let i = 0; i < check.length; i++) {
```
danielakhterov, line 79: Should just use a for each loop instead
```javascript
for (const c of checks) {
   ...
}
```

test/unit/Hbar.js
```diff
@@ -2,20 +2,169 @@ import Hbar from "../src/Hbar.js";
...
+            expect(check[i].replace("+", "") + " " + unit).to.equal(
+                Hbar.fromString(check[i]).toString()
+            );
+        }
+    });
+
+    it("should pass regex, convert to tinybar, and reverse [to|from]String", function () {
```
danielakhterov, line 86: This is not needed since you already added the same tests above.

test/unit/Hbar.js
```diff
@@ -2,20 +2,169 @@ import Hbar from "../src/Hbar.js";
...
     });
 
-    it("should convert from hbars to tinybars", function () {
-        const tinybar = "98327923153";
-        const hbar = new Hbar("983.27923153");
+    it('should throw error "Hbar in tinybars contains decimals"', function () {
+        let check = ["1.151 tℏ", "1.151 μℏ", "1.151", "-1.151", "+1.151"];
 
-        expect(tinybar).to.eql(hbar.toTinybars().toString());
+        for (let i = 0; i < check.length; i++) {
```
danielakhterov, line 158: Should use a for each loop here instead. Also, isn't this test the same as the one above where you're using `i`, `j`, and `k`?

danielakhterov (changes requested):

test/unit/Hbar.js
andvara-naut: I'll look into for each. The tests are not the same. This one tests the specific "Hbar in tinybars contains decimals". The i,j,k tests the regex.

test/unit/Hbar.js
andvara-naut: they arent the same tests but ill remove it anyway

danielakhterov (approved):

Issue #446, opened 2021-07-13
Free queries shouldn't created signed payment transactions


danielakhterov:
### Description

Right now, for `AccountBalanceQuery` still creates a signed `paymentTransaction` and `paymentTransactionId` even though those aren't needed or used for the query itself since it's free. This usually isn't an issue, but when used with a hardware wallet the wallet will be asked to sign for what should be a free query.

### Steps to reproduce

Query account balance with a client that has an operator set.

### Hedera network

mainnet, testnet, previewnet

### Version

v2.0.24

### Operating system

Linux

PR #472, opened 2021-07-30
fix: free queries should not require payment transactions to be signed

danielakhterov:

**Description**:
Free queries currently require signing even though the payment transaction is a stub with accounts and transfers being `0`. Since the original code was already checking for `operator == null` to determine if a real payment transaction was being generated I updated the parameter to be `null` whenever `this._isPaymentRequired()` is `false`.

**Related issue(s)**:

Closes: #446 

**Notes for reviewer**:

**Checklist**


andvara-naut (approved):
don't see anything breaking here.

PR #480, opened 2021-08-06
Implement helper methods on CustomFixedFee

andvara-naut:
**Problem**
When denominating_token_id is null, a fixed fee will be charged in Hbar, but currently the amount of the fee is a long which specifies the Hbar fee in tinybars.

Additionally, when denominating_token_id is 0.0.0, the fee should be charged in the same token. The user should not have to know about this in order to set the fixed fee denominating token to the same token.

**Solution**
- Add CustomFixedFee.[set|get]HbarAmount(), which sets denominating_token_id to null and has an Hbar parameter.
- Add CustomFixedFee.setDenominatingTokenToSameToken(), which sets denominating_token_id to 0.0.0


src/token/CustomFixedFee.js
```diff
@@ -38,6 +39,35 @@ export default class CustomFixedFee extends CustomFee {
...
+    setHbarAmount(amount) {
+        this._amount = amount.toTinybars();
+        this._denominatingTokenId = null;
+        return this;
+    }
+
+    /**
+     * @returns {Hbar} 
+     */
+    get hbarAmount() {
```
danielakhterov, line 55: Can we update this implementation to return `null` if `this._denominatingTokenId` is not `null`? Otherwise if a user does `CustomFixedFee.setDenominatingTokenId("0.0.10").setAmount(10).hbarAmount` they'd get a non-null value implying the `CustomFixedFee` is a Hbar amount.
Note: Your implementation is in line with Java's, and I'll make an issue to update Java as well.

danielakhterov (changes requested):

danielakhterov (approved):

PR #533, opened 2021-08-23
Client.[min|max]Backoff

andvara-naut:
**Problem**
Currently each request has a default min backoff of 250ms and ma backoff of 8s. A user should be able to set these values themselves.

**Solution**
Implement Client.[min|max]Backoff


src/Executable.js
```diff
@@ -51,6 +51,12 @@ export default class Executable {
          * @type {AccountId[]}
          */
         this._nodeIds = [];
+
+        /** @type {number} */
+        this._minBackoff = 250;
```
danielakhterov, line 56: These fields should be null-able and default to `null`

src/Executable.js
```diff
@@ -279,7 +315,7 @@ export default class Executable {
 
             switch (this._shouldRetry(request, response)) {
                 case ExecutionState.Retry:
-                    await delayForAttempt(attempt);
+                    await delayForAttempt(attempt,this._minBackoff,this._maxBackoff);
```
danielakhterov, line 318:
```java
        if (this._maxBackoff == null) {
            this._maxBackoff = client.getMaxBackoff();
        }

        if (this._minBackoff == null) {
            this._minBackoff = client.getMinBackoff();
        }
```
Needs to be done at beginning of this method.

src/Executable.js
```diff
@@ -296,9 +332,11 @@ export default class Executable {
 
 /**
  * @param {number} attempt
+ * @param {number} minBackoff
+ * @param {number} maxBackoff
  * @returns {Promise<void>}
  */
-function delayForAttempt(attempt) {
+function delayForAttempt(attempt, minBackoff, maxBackoff) {
```
steven-sheehy, line 339: The new parameters are unused

steven-sheehy (changes requested):

src/Executable.js
```diff
@@ -105,6 +111,63 @@ export default class Executable {
         return this;
     }
 
+    /**
+     * @param {number} minBackoff
+     */
+    setMinBackoff(minBackoff) {
+        if (this._maxBackoff != null && minBackoff >= this._maxBackoff) {
```
danielakhterov, line 118: I believe this should be `minBackoff > this._maxBackoff` since those two fields being equal would indicate backoff is linear instead of exponential.

src/Executable.js
```diff
@@ -105,6 +111,63 @@ export default class Executable {
...
+     */
+    get minBackoff() {
+        return this._minBackoff;
+    }
+
+    /**
+     * @param {number} maxBackoff
+     */
+    setMaxBackoff(maxBackoff) {
+        if (this._minBackoff != null && maxBackoff <= this._minBackoff) {
```
danielakhterov, line 137: Similarly this should use `<` instead of `<=`.

src/Executable.js
```diff
@@ -105,6 +111,63 @@ export default class Executable {
...
+     */
+    get maxBackoff() {
+        return this._maxBackoff;
+    }
+
+    /**
+     * @param {number} minBackoff
+     * @param {number} maxBackoff
+     */
+    setBackoff(minBackoff, maxBackoff) {
```
danielakhterov, line 156: Although I do like this method, can we make this private for the time being. I'd like to get this added to Java before we add this method here.

src/Executable.js
```diff
@@ -105,6 +111,63 @@ export default class Executable {
...
+        }
+        this._minBackoff = minBackoff;
+        this._maxAttempts = maxBackoff;
+    }
+
+    /**
+     * @returns {number | null}
+     */
+    get backoff() {
+        return this._maxBackoff, this._maxBackoff;
```
danielakhterov, line 168: The return value doesn't match the return type. Perhaps we should return an object
```javascript
{
    min: number;
    max: number;
}
```

src/Executable.js
```diff
@@ -279,7 +342,17 @@ export default class Executable {
 
             switch (this._shouldRetry(request, response)) {
                 case ExecutionState.Retry:
-                    await delayForAttempt(attempt);
+                    if (this._maxBackoff == null) {
```
danielakhterov, line 345: This conditional should be moved to the beginning of this method since it essentially no-ops after the first iteration.

danielakhterov (changes requested):

src/Executable.js
```diff
@@ -105,6 +111,63 @@ export default class Executable {
...
+    /**
+     * @param {number} minBackoff
+     * @param {number} maxBackoff
+     */
+    setBackoff(minBackoff, maxBackoff) {
+        if (minBackoff <= maxBackoff) {
+            throw new Error("minBackoff cannot be larger than maxBackoff.");
+        }
+        this._minBackoff = minBackoff;
+        this._maxAttempts = maxBackoff;
```
steven-sheehy, line 161: Bad assignment

src/client/Client.js
```diff
@@ -337,6 +343,36 @@ export default class Client {
         return this;
     }
 
+    /**
+     * @param {number} minBackoff
+     */
+    setMinBackoff(minBackoff) {
+        this._minBackoff = minBackoff;
```
steven-sheehy, line 350: This should do the same validations as `Executable.setMinBackoff()`. Right now this allows setting minBackoff to be greater than maxBackoff.

src/client/Client.js
```diff
@@ -337,6 +343,36 @@ export default class Client {
...
+     */
+    get minBackoff() {
+        return this._minBackoff;
+    }
+
+    /**
+     * @param {number} maxBackoff
+     */
+    setMaxBackoff(maxBackoff) {
+        if (maxBackoff >= this.minBackoff) {
```
steven-sheehy, line 364: This should do the same validations as Executable.setMaxBackoff()

src/Executable.js
steven-sheehy, line 137: This should throw an error if `maxBackoff` is null. Have a default of null is fine for `maxBackoff` but if the user explicitly sets it it should be with a non-null value.

src/Executable.js
steven-sheehy, line 118: This should throw an error if `minBackoff ` is null.

steven-sheehy (changes requested):

src/Executable.js
danielakhterov: I meant the beginning of the `execute` method. The point is that by having the conditions within the switch statement they'll be executed each time the request is retried which doesn't make much sense. It'd be better to simply execute them ones ahead of time. You're not wrong that by having them within the switch statement they might never be executed making it even more efficient, but the likeliness of that happening is quite slim; speaking from experience.

src/client/Client.js
```diff
@@ -337,6 +343,83 @@ export default class Client {
...
+        return this._maxBackoff;
+    }
+
+    /**
+     * @private
+     * @param {?number} minBackoff
+     * @param {?number} maxBackoff
+     * @returns {this}
+     */
+    // @ts-ignore
```
danielakhterov, line 395: We shouldn't be using `ts-ignore`. 
Also, unfortunately adding `@private` is not all you have to do to make something look private. You'll need to add `_` prefix to the method name. Something like `_setBackoff(...)`. If someone uses this SDK in JS they won't realize `setBackoff()` is private, but adding an `_` in the beginning is a recognized convention for private methods/fields.

src/client/Client.js
```diff
@@ -337,6 +343,83 @@ export default class Client {
...
+    }
+
+    /**
+     * @private
+     * @typedef {Object} Backoff
+     * @property {number | null} minBackoff
+     * @property {number | null} maxBackoff
+     * @returns {Backoff}
+     */
+    // @ts-ignore
```
danielakhterov, line 418: Same case as above.

src/Executable.js
```diff
@@ -105,6 +111,82 @@ export default class Executable {
...
+    }
+
+    /**
+     * @private
+     * @typedef {Object} Backoff
+     * @property {number | null} minBackoff
+     * @property {number | null} maxBackoff
+     * @returns {Backoff}
+     */
+    // @ts-ignore
```
danielakhterov, line 185: Same case as in `Client`.

src/Executable.js
```diff
@@ -105,6 +111,82 @@ export default class Executable {
...
+        return this._maxBackoff;
+    }
+
+    /**
+     * @private
+     * @param {?number} minBackoff
+     * @param {?number} maxBackoff
+     * @returns {this}
+     */
+    // @ts-ignore
```
danielakhterov, line 162: Same case as in `Client`

danielakhterov (changes requested):

src/Executable.js
```diff
@@ -105,6 +111,80 @@ export default class Executable {
...
+        return this._maxBackoff;
+    }
+
+    /**
+     * @private
+     * @param {?number} minBackoff
+     * @param {?number} maxBackoff
+     * @returns {this}
+     */
+    _setBackoff(minBackoff, maxBackoff) {
```
danielakhterov, line 162: Damn, I see why you added the `@ts-ignore` :( This is definitely my fault for forgoting that it would error from `tsc` instead of `eslint`. Unfortunately, I really don't want us to add `@ts-ignore` anywhere, so can we remove `[Executable|Client]._setBackoff()` and `[Executable|Client]._backoff()`. We might add those back later, but for this PR lets just keep it as in-sync with the other SDKs as we can.

danielakhterov (changes requested):

src/Executable.js
andvara-naut: i swear

danielakhterov (approved):

PR #538, opened 2021-08-26
fix: update `Executable._shouldRetryExceptionally()` to use `.code` f…

### LINKED ISSUE #494: Executable retry logic uses incorrect object equality comparison
### Description

The retry logic in [_shouldRetryExceptionally(error)](https://github.com/hashgraph/hedera-sdk-js/blob/main/src/Executable.js#L210) uses object referential equality which will return false even when the objects inner values are matching since the object instances are different.

I believe this is the cause for the lack of retries on expected/occasional errors.

Instead the code should either 
1. Do a deep comparison e.g. using `util.isDeepStrictEqual(val1, val2)`
or
2. Simplify the comparison by comparing the actual `_code` value from each object instead

### Steps to reproduce

1. Run the SDK against an endpoint that is not available to mimic GrpcStatus.Unavailable case
Should be the same for others

### Additional context

_No response_

### Hedera network

other

### Version

v02.0.28

### Operating system

macOS

danielakhterov:
…or comparing error statuses instead of comparing the instances themselves

**Description**:

**Related issue(s)**:

Fixes #494 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/src/Executable.js
+++ b/src/Executable.js
@@ -209,9 +209,9 @@ export default class Executable {
      */
     _shouldRetryExceptionally(error) {
         return (
-            error.status === GrpcStatus.Unavailable ||
-            error.status === GrpcStatus.ResourceExhausted ||
-            (error.status === GrpcStatus.Internal &&
+            error.status.code === GrpcStatus.Unavailable.code ||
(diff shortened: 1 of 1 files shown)


andvara-naut (dismissed):
lgtm

PR #584, opened 2021-09-08
fix: renamed consensusTimestampstamp to consensusTimestamp

derdis14:


derdis14:
Locally, the integration tests were successful. However, I had to uncomment lines 14 and 57 regarding dotenv in the IntegrationTestEnv class. Why are these commented?

danielakhterov (approved):

danielakhterov:
The reason `dotenv` was commented out was because it was giving me sort of trouble awhile ago. Should uncomment it now though.

PR #606, opened 2021-09-15
fix/TransactionTransfer

valtyr-naut:

### DIFF
--- a/src/account/TransferTransaction.js
+++ b/src/account/TransferTransaction.js
@@ -236,7 +236,7 @@ export default class TransferTransaction extends Transaction {
         this._requireNotFrozen();
 
         if (this.hbarTransfers.get(accountId) != null) {
-            amount = this.squashTransfers(accountId, amount);
+            amount = this._squashTransfers(accountId, amount);
         }
 
         this._hbarTransfers._set(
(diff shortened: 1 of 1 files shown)


src/account/TransferTransaction.js
```diff
@@ -244,6 +249,24 @@ export default class TransferTransaction extends Transaction {
         return this;
     }
 
+    /**
+     * @internal
+     * @param {AccountId | string} accountId
+     * @param {number | string | Long | LongObject | BigNumber | Hbar} amount
+     * @returns {number | Long}
+     */
+    squashTransfers(accountId, amount) {
```
danielakhterov, line 258: This should have an `_` prefix. Also, I recommend you return `Hbar` instead of `number | Long`

src/account/TransferTransaction.js
```diff
@@ -244,6 +249,24 @@ export default class TransferTransaction extends Transaction {
...
+        let currentValue = this.hbarTransfers.get(accountId);
+        return Number(
+            Hbar.fromTinybars(
+                (currentValue == null
+                    ? Long.ZERO
+                    : currentValue.toTinybars()
+                ).add(new Hbar(Number(amount)).toTinybars())
```
danielakhterov, line 265: Instead of casting `amount` to `Number` and passing that into `new Hbar()` you should do `amount instanceof Hbar ? amount : new Hbar(amount)`

danielakhterov (changes requested):

danielakhterov (approved):

Issue #624, opened 2021-09-22
TopicMessageSubmitTransaction.setMessage throws "silent" exception


gregscullard:
### Description

When the input to .setMessage is incorrect, the JS code continues onto the next instruction when it should throw an exception. In other words, something will go wrong later in the code and the root cause of the issue won't be surfaced.

### Steps to reproduce

```
        let hcsMessageJson = {
            "message": "message json : " + i.toString()
        };

        const mst = await new TopicMessageSubmitTransaction()
            .setTopicId(newTopicId)
            .setMessage(hcsMessageJson)
            .freezeWith(myClient);

        mst.sign(rootKey1);
```

Throws an exception on `mst.sign` indicating the transaction wasn't frozen.

Switching to `.setMessage(JSON.stringify(hcsMessageJson))` works.

I would have expected that `.setMessage` fails rather than silently continue

### Hedera network

testnet

### Version

v2.0.30

### Operating system

macOS

PR #642, opened 2021-09-29
fix: silent set message error

valtyr-naut:
TopicMessageSubmitTransaction.setMessage() silent error changed to throw when passed non string - uint8array or undefined.

Linked Issue: #624 

### DIFF
--- a/src/topic/TopicMessageSubmitTransaction.js
+++ b/src/topic/TopicMessageSubmitTransaction.js
@@ -165,7 +165,7 @@ export default class TopicMessageSubmitTransaction extends Transaction {
     setMessage(message) {
         this._requireNotFrozen();
 
-        if (util.REQUIRE_STRING(message) || util.REQUIRE_UINT8ARRAY(message)) {
+        if (util.requireString(message) || util.requireUint8Array(message)) {
             this._message =
                 typeof message === "string" ? utf8.encode(message) : message;
         }
--- a/src/topic/TopicMessageSubmitTransaction.js
+++ b/src/topic/TopicMessageSubmitTransaction.js
@@ -6,7 +6,7 @@ import TopicId from "./TopicId.js";
 import * as utf8 from "../encoding/utf8.js";
 import TransactionId from "../transaction/TransactionId.js";
 import Timestamp from "../Timestamp.js";
-import { util } from "../util.js";
+import * as util from "../util.js";
 
 /**
  * @namespace proto
@@ -165,7 +165,7 @@ export default class TopicMessageSubmitTransaction extends Transaction {
     setMessage(message) {
         this._requireNotFrozen();
 
-        if (util.requireString(message) || util.requireUint8Array(message)) {
+        if (util.REQUIRE_STRING(message) || util.REQUIRE_UINT8ARRAY(message)) {
             this._message =
                 typeof message === "string" ? utf8.encode(message) : message;
         }
--- a/src/topic/TopicMessageSubmitTransaction.js
+++ b/src/topic/TopicMessageSubmitTransaction.js
@@ -163,9 +163,15 @@ export default class TopicMessageSubmitTransaction extends Transaction {
      */
     setMessage(message) {
         this._requireNotFrozen();
-        this._message =
-            typeof message === "string" ? utf8.encode(message) : message;
-
+        if (message == undefined) {
+            throw new Error("TopicSubmitMessageTransaction.setMessage() given undefined message.");
+        }
+        if (typeof message === "string" || message instanceof Uint8Array) {
+            this._message = typeof message === "string" ? utf8.encode(message) : message;
+        }
+        else{
+            throw new Error("TopicSubmitMessageTransaction.setMessage() given non string/Uint8Array message.");
+        }
         return this;
     }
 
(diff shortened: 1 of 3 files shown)


src/topic/TopicMessageSubmitTransaction.js
```diff
@@ -163,9 +163,19 @@ export default class TopicMessageSubmitTransaction extends Transaction {
      */
     setMessage(message) {
         this._requireNotFrozen();
-        this._message =
-            typeof message === "string" ? utf8.encode(message) : message;
-
+        if (message == undefined) {
```
danielakhterov, line 166:
```suggestion
        if (message == null) {
```

src/topic/TopicMessageSubmitTransaction.js
```diff
@@ -163,9 +163,19 @@ export default class TopicMessageSubmitTransaction extends Transaction {
...
+            );
+        }
+        if (typeof message === "string" || message instanceof Uint8Array) {
+            this._message =
+                typeof message === "string" ? utf8.encode(message) : message;
+        } else {
+            throw new Error(
+                "TopicSubmitMessageTransaction.setMessage() given non string/Uint8Array message."
+            );
+        }
```
danielakhterov, line 178: Since you're already made an condition to exit early with an error I think it'd be a good idea to be consistent and make another condition to exit early if the types don't match.
```suggestion
        if (typeof message !== "string" && !(message instanceof Uint8Array)) {
            throw new Error(
                "TopicSubmitMessageTransaction.setMessage() given non string/Uint8Array message."
            );
        }

        this._message =
            typeof message === "string" ? utf8.encode(message) : message;
```

src/topic/TopicMessageSubmitTransaction.js
danielakhterov: Since you're creating a
```javascript
if (value == null) {
    throw new Error("value cannot be null");
}
```
Maybe it'd be a good idea to move this into a separate function inside `./src/util.js` or some other file which does this exact check? Something like:
```javascript
/**
 * @template {*} T
 * @param {T | null | undefined} value
 * @returns {NonNullable<T>}
 */
function requireNonNull(value = undefined) {
    if (value == null) {
        throw new Error("value cannot be null");
    }

    return /** @type {NonNullable<T>} */ (value);
}
```

danielakhterov (changes requested):

danielakhterov:
If you use `Fixed #<issue number>` instead of `Linked Issue: #<issue number>` it will be automatically closed when this PR is merged.

src/util.js
```diff
@@ -0,0 +1,46 @@
+export class util {
```
danielakhterov, line 1: Shouldn't be a class.This file should just contain a collection of helper methods.

src/util.js
```diff
@@ -0,0 +1,46 @@
...
+    /**
+     * Utility Error Messages
+     */
+    static requireNonNullError = "This value cannot be null nor undefined.";
```
danielakhterov, line 5: Move these outside the class and change their name to be `SCREAMING_SNAKE_CASE`

src/util.js
```diff
@@ -0,0 +1,46 @@
...
+            return true;
+        }
+    }
+
+    /**
+     * @param {any | null | undefined} variable
+     * @returns {boolean}
+     */
+    static requireUint8Array(variable) {
+        this.requireNonNull(variable);
```
danielakhterov, line 39: Since this is a static method `this` would not work here.

src/topic/TopicMessageSubmitTransaction.js
```diff
@@ -6,6 +6,7 @@ import TopicId from "./TopicId.js";
 import * as utf8 from "../encoding/utf8.js";
 import TransactionId from "../transaction/TransactionId.js";
 import Timestamp from "../Timestamp.js";
+import { util } from "../util.js";
```
danielakhterov, line 9: After updating `util` to be collection of methods instead of a class.
```suggestion
import * as util from "../util.js";
```

test/unit/TopicMessageSubmitTransaction.js
```diff
@@ -0,0 +1,44 @@
+import TopicMessageSubmitTransaction from "../src/topic/TopicMessageSubmitTransaction.js";
+import * as utf8 from "../src/encoding/utf8.js";
+import { util } from "../src/util.js";
```
danielakhterov, line 3:
```suggestion
import * as util from "../src/util.js";
```

danielakhterov (changes requested):

src/util.js
```diff
@@ -0,0 +1,44 @@
...
+ */
+export const REQUIRE_NON_NULL_ERROR = "This value cannot be null | undefined.";
+export const REQUIRE_STRING_ERROR = "This value must be a string.";
+export const REQUIRE_UINT8ARRAY_ERROR = "This value must be a Uint8Array.";
+
+/**
+ * @param {any | null | undefined} variable
+ * @returns {boolean}
+ */
+export function REQUIRE_NON_NULL(variable) {
```
danielakhterov, line 12: Heh, should of clarified. The functions should still be `camelCase`, but the strings should be `SCREAMING_SNAKE_CASE` since they're non-changing variables.

danielakhterov (changes requested):

danielakhterov (approved):

PR #671, opened 2021-10-14
feat: timeout grpc's connection manually since it doesn't actually us…

### LINKED ISSUE #670: gRPC connection does not have a timeout
### Description

Since the gRPC connection doesn't use a timeout a connection to a server that isn't responding doesn't close for a long time transaction requests on a retry will fail with `TRANSACTION_EXPIRED`.

### Steps to reproduce

Use 
```javascript
    'grpc.keepalive_timeout_ms'?: number;
```
to set a timeout on a gRPC client

### Additional context

_No response_

### Hedera network

other

### Version

v2.1.0

### Operating system

Linux

janaakhterov:
…e the timeout_ms option

**Description**:

**Related issue(s)**:

Fixes #670 

**Notes for reviewer**:

Unfortunately grpc-node doesn't seem to support timeouts right now according to https://github.com/grpc/grpc-node/issues/1593 https://github.com/grpc/grpc-node/issues/1545 so instead I added a manual "timeout" using `setTimeout()` This makes nodes which do not respond act as if they were to respond with `UNAVAILABLE`

**Checklist**

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -8,7 +8,7 @@ specifiers:
   '@babel/plugin-transform-modules-commonjs': ^7.15.4
   '@babel/register': ^7.15.3
   '@grpc/grpc-js': ^1.3.4
-  '@hashgraph/cryptography': ^1.0.20
+  '@hashgraph/cryptography': ^1.0.21
   '@hashgraph/proto': 2.0.0
   '@types/chai': ^4.2.22
   '@types/crypto-js': ^4.0.2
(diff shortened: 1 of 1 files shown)


gregscullard:
> @gregscullard Currently the JS SDK doesn't do any mocking (we will in the future but the infrastructure just isn't there yet), so I'll add a test that uses nodes 10 and 3 in that order on mainnet. This test will not work all the time though since node 10 can start working again.

OK, that makes sense. It may be worth raising an issue wrt mocking so that this scenario is captured are a requirement for the mock tests when you get to it.

PR #732, opened 2021-11-02
fix: util entityId import

valtyr-naut:
Signed-off-by: valtyr-naut <dalton.kaua@launchbadge.com>

removed util.js import from EntityId.


danielakhterov:
Superseded by: #733

PR #779, opened 2021-11-30
chore: drop direct protobufjs dep, it's unused

ChALkeR:
**Description**:

This removes direct `protobufjs` from `@hashgraph/sdk` itself.

Only `@hashgraph/proto` depends on `protobufjs`, and the version specified in `@hashgraph/sdk` is ineffective -- the code uses the `protobufjs` version from `@hashgraph/proto` deps either way, as all the imports of `protobufjs` are done by `@hashgraph/proto`.

There are no direct `protobufjs` imports in `@hashgraph/sdk`.

**Checklist**

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,6 @@
         "crypto-js": "^4.0.0",
         "js-base64": "^3.6.1",
         "long": "^4.0.0",
-        "protobufjs": "^6.11.2",
         "utf8": "^3.0.0"
     },
     "devDependencies": {


danielakhterov:
We don't use `protobufjs` directly, but we do use the types defined by it. `@types/protobufjs` is deprecated and we are encouraged to use `protobufjs` as a dev dependency instead. The CI is current failing because of dependabot updating package without updating `pnpm-lock.yml`, but when I pulled your branch and ran `pnpm i && pnpm format && pnpm lint` it erred.

ChALkeR:
Ah, I see why that was done. Sorry for the confusion.

PR #780, opened 2021-11-30
feat: export encoding methods from cryptography package

ChALkeR:
**Description**:

This adds encoding methods to exports of `@hashgraph/cryptography`, so that the code could be reused in `@hashgraph/sdk` instead of shipping it copy-pasted.

Currently, both `@hashgraph/cryptography` and `@hashgraph/sdk` have the same code:
* https://github.com/hashgraph/hedera-sdk-js/tree/main/src/encoding
* https://github.com/hashgraph/hedera-sdk-js/tree/main/packages/cryptography/src/encoding

This will allow to remove that duplication in a separate PR.

**Checklist**

### DIFF
--- a/packages/cryptography/src/encoding.js
+++ b/packages/cryptography/src/encoding.js
@@ -1,0 +1,14 @@
+import * as base64 from "./encoding/base64.js";
+export { base64 };
+
+import * as der from "./encoding/der.js";
+export { der };
+
+import * as hex from "./encoding/hex.js";
+export { hex };
+
+import * as pem from "./encoding/pem.js";
+export { pem };
+
+import * as utf8 from "./encoding/utf8.js";
+export { utf8 };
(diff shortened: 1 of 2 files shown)


danielakhterov:
The code was duplicated deliberately. `@hashgraph/cryptography` nor `@hashgraph/sdk` are not encoding libraries and those APIs are not meant to be use by users. By exporting them we'd be adding them to the public API which means we'd need to support them and prevent breaking changes which is not something we want. Perhaps we can deduplicate them using links instead, but exporting them from `@hashgraph/cryptography` is not an option.

ChALkeR:
Perhaps moving that into a separate utils package could resolve the code duplication?

danielakhterov:
I'm on the edge about having a separate package simply because some other users might try to use it and expect it to be stable even if it's a <1.0 release. Nonetheless, definitely something to consider if we really want to de-duplicate it.

PR #782, opened 2021-12-01
fix: use arr.buffer safely

ChALkeR:
**Description**:

This SDK both
1. Uses `Buffer.from` in some places and treats them as `Uint8Array` instances:
   ```
    src/channel/NodeMirrorChannel.js
    src/channel/NodeChannel.js
    src/encoding/hex.js
    src/encoding/utf8.js
    packages/cryptography/src/encoding/base64.js
    packages/cryptography/src/encoding/hex.js
    packages/cryptography/src/encoding/utf8.js
    ```
    Note that `Buffer.from` isn't the only source of this, as this SDK also creates pooled `Uint8Array` instances manually, e.g.: 
  https://github.com/hashgraph/hedera-sdk-js/blob/9c9ace7ddb67f46f72105066d2be2fbc37991d5c/src/channel/Channel.js#L252

2. Treats `Uint8Array` instances as non-pooled in most places across the codebase (see the diff of this PR for exact locations).

This seems very fragile, as accidentally treating a pooled `Uint8Array` instance as a non-pooled one might result in a catastrophic failure, e.g. leaking unrelated data over the network.

E.g.:
```console
> Buffer.from('x')
<Buffer 78>
> Buffer.from('x').buffer
ArrayBuffer {
  [Uint8Contents]: <2f 00 00 00 00 00 00 00 2f 00 00 00 00 00 00 00 78 00 00 00 00 00 00 00 78 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 ... 8092 more bytes>,
  byteLength: 8192
}
```

This PR fixes that by ensuring that each time `arr.buffer` is used, we use appropriate `arr.byteOffset` and `arr.byteLength`.

**Checklist**

### DIFF
--- a/src/channel/Channel.js
+++ b/src/channel/Channel.js
@@ -233,9 +233,12 @@ export function encodeRequest(data) {
 
 /**
  * @param {ArrayBuffer} data
+ * @param {number} byteOffset
+ * @param {number} byteLength
  * @returns {Uint8Array}
  */
-export function decodeUnaryResponse(data) {
+export function decodeUnaryResponse(data, byteOffset = 0, byteLength = data.byteLength) {
+    const dataView = new DataView(data, byteOffset, byteLength);
     let dataOffset = 0;
 
     /** @type {?Uint8Array} */
@@ -244,11 +247,10 @@ export function decodeUnaryResponse(data) {
     // 0 = successful
     let status = 0;
 
-    while (dataOffset < data.byteLength) {
-        const dataView = new DataView(data, dataOffset);
-        const frameByte = dataView.getUint8(0);
+    while (dataOffset < dataView.byteLength) {
+        const frameByte = dataView.getUint8(dataOffset + 0);
         const frameType = frameByte >> 7;
-        const frameByteLength = dataView.getUint32(1);
+        const frameByteLength = dataView.getUint32(dataOffset + 1);
         const frameData = new Uint8Array(data, dataOffset + 5, frameByteLength);
 
         if (frameType === 0) {
--- a/src/channel/Channel.js
+++ b/src/channel/Channel.js
@@ -237,7 +237,11 @@ export function encodeRequest(data) {
  * @param {number} byteLength
  * @returns {Uint8Array}
  */
-export function decodeUnaryResponse(data, byteOffset = 0, byteLength = data.byteLength) {
+export function decodeUnaryResponse(
+    data,
+    byteOffset = 0,
+    byteLength = data.byteLength
+) {
     const dataView = new DataView(data, byteOffset, byteLength);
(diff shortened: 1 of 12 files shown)


ChALkeR:
Lint fails, will fix that shortly.

src/channel/Channel.js
```diff
@@ -235,16 +235,16 @@ export function encodeRequest(data) {
...
+    let dataOffset = offset;
 
     /** @type {?Uint8Array} */
     let unaryResponse = null;
 
     // 0 = successful
     let status = 0;
 
-    while (dataOffset < data.byteLength) {
+    while (dataOffset < byteLength) {
```
ChALkeR, line 247: still looks unsafe as this can read past the `byteLength`

ChALkeR: fixed

danielakhterov (approved):

src/channel/Channel.js
```diff
@@ -244,11 +251,10 @@ export function decodeUnaryResponse(data) {
...
 
-    while (dataOffset < data.byteLength) {
-        const dataView = new DataView(data, dataOffset);
-        const frameByte = dataView.getUint8(0);
+    while (dataOffset < dataView.byteLength) {
+        const frameByte = dataView.getUint8(dataOffset + 0);
         const frameType = frameByte >> 7;
-        const frameByteLength = dataView.getUint32(1);
+        const frameByteLength = dataView.getUint32(dataOffset + 1);
         const frameData = new Uint8Array(data, dataOffset + 5, frameByteLength);
```
ChALkeR, line 258: Upon a recheck, this is still not guarded here on this line, as `data` is an `ArrayBuffer` instance.
I'll file a follow-up PR.

src/util.js
```diff
@@ -297,3 +297,23 @@ export function convertToNumber(variable) {
...
+ *
+ * @param {Uint8Array} arr
+ * @param {number | undefined} offset
+ * @param {number | undefined} length
+ * @returns {DataView}
+ */
+export function safeView(arr, offset = 0, length = arr.byteLength) {
+    if (!(Number.isInteger(offset) && offset >= 0))
+        throw new Error("Invalid offset!");
+    if (!(Number.isInteger(length) && length > 0))
```
ChALkeR, line 312: Likely should use `>=` to be sure this works on empty arrays in case if those are needed.

PR #783, opened 2021-12-01
fix: serialization before signing

feri42:
**Description**:

Serializing a Transaction object with  `toBytes()`  before signing it will corrupt the object in a way which will prevent any further signing. Also serializing after signing and then attempting a second signature fails for the same reason. Here I submit test cases that should all be successful and a proposed solution. Maybe there is a better fix for this.

It is not only `toBytes()` that corrupts the transaction object, it is also `getSignatures()` and `getTransactionhash()`. Which is not expected, as they are get-functions.

The use case for mixed signing and serializing is when we want to send a transaction through IPC for signing by a different process. This used to work in the 1.x verion of the Javascript SDK.

To reproduce the issue revert the commit that modifies `Transaction.js`, then run:
```
npm run test:unit:single test/unit/Serialize.js 
```
Some test will fail, after enabling the fix, all will succeed which is the expected behaviour.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,6 @@
         "test": "run-s test:node test:browser",
         "test:node": "env HEDERA_SDK_TEST='' run-s test:unit:node test:integration:node",
         "test:unit:node": "env HEDERA_SDK_TEST='' nyc mocha -r @babel/register -r chai/register-expect 'test/unit/*.js'",
-        "test:unit:single": "env HEDERA_SDK_TEST='' nyc mocha -r @babel/register -r chai/register-expect",
         "test:integration:node": "env HEDERA_SDK_TEST='' nyc mocha -r @babel/register -r chai/register-expect 'test/integration/*.js'",
         "test:browser": "run-s test:unit:browser test:integration:browser",
         "test:browser:chrome": "run-s test:unit:browser:chrome test:integration:browser:chrome",
--- a/package.json
+++ b/package.json
@@ -64,6 +64,7 @@
         "test": "run-s test:node test:browser",
         "test:node": "env HEDERA_SDK_TEST='' run-s test:unit:node test:integration:node",
         "test:unit:node": "env HEDERA_SDK_TEST='' nyc mocha -r @babel/register -r chai/register-expect 'test/unit/*.js'",
+        "test:unit:single": "env HEDERA_SDK_TEST='' nyc mocha -r @babel/register -r chai/register-expect",
         "test:integration:node": "env HEDERA_SDK_TEST='' nyc mocha -r @babel/register -r chai/register-expect 'test/integration/*.js'",
         "test:browser": "run-s test:unit:browser test:integration:browser",
         "test:browser:chrome": "run-s test:unit:browser:chrome test:integration:browser:chrome",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -64,6 +64,7 @@
         "test": "run-s test:node test:browser",
         "test:node": "env HEDERA_SDK_TEST='' run-s test:unit:node test:integration:node",
         "test:unit:node": "env HEDERA_SDK_TEST='' nyc mocha -r @babel/register -r chai/register-expect 'test/unit/*.js'",
+        "test:unit:single": "env HEDERA_SDK_TEST='' nyc mocha -r @babel/register -r chai/register-expect",
```
danielakhterov, line 67: Not entirely sure why this was added to be honest. If you'd like to test a single unit test you should be able to use
```
pnpm test:unit:node -- -g "Mix signing and serialization.*"
```

test/unit/Serialize.js
```diff
@@ -0,0 +1,98 @@
...
+    operatorId: AccountId.fromString("0.0.666"),
+    senderId: AccountId.fromString("0.0.123456"),
+    recipientId: AccountId.fromString("0.0.654321"),
+    validStart: Timestamp.fromDate(new Date(1640000000000)),
+    amount: Hbar.fromTinybars("50505050505"),
+    fee: Hbar.fromTinybars("100000000"),
+    memo: "Transaction (de)serialization test",
+};
+
+const serialize = (tx) => tx.toBytes().toString("hex");
```
danielakhterov, line 30: For consistencies sake can we declare these as functions instead. Same goes for `deserialize`, `sign` and `buildTx`
```suggestion
function serialize(tx) {
    return tx.toBytes().toString("hex");
}
```

test/unit/Serialize.js
```diff
@@ -0,0 +1,98 @@
...
+};
+
+describe("Mix signing and serialization", function () {
+    it("Sign then serialize", function () {
+        const tx = buildTx(params);
+        sign(tx, PRIVATE_KEY1);
+        sign(tx, PRIVATE_KEY2);
+        const serialized = serialize(tx);
+        expect(serialized).equals(SERIALIZED);
+    });
```
danielakhterov, line 57: Can we add a new line between each `it()` to better separate them visually.

danielakhterov (changes requested):
Changes themselves look good, but nitpicked your test a little bit.

package.json
feri42: Thanks wasn't aware of that. Will remove.

feri42:
I didn't test this, but `_buildTransactionAsync()` may have a similar issue. Not sure though how the `_signTranscation()` call would work with the same fix.

danielakhterov (approved):

danielakhterov:
Make sure to sign off on all the previous commits to pass DCO

danielakhterov:
Make sure to sign off on all the previous commits to pass DCO

danielakhterov:
It does look like `buildTransactionAsync` would have the same issue. Also just noticed I spelt `Transaction` wrong there :laughing:

feri42:
Lol. And I just copied without noticing.

danielakhterov (approved):

PR #831, opened 2021-12-22
feat: HIP-33

valtyr-naut:
Signed-off-by: valtyr-naut <dalton.kaua@launchbadge.com>

NetworkName class

LedgerId
- [x] Constants
- [x] fromString
- [x] toString
- [x] fromBytes
- [x] toBytes
- [x] fromNetworkName
- [x] toNetworkName
- [x] isMainnet
- [x] isTestnet
- [x] isPreviewnet
- [x] Test
- [x] lazy deprecate NetworkName

https://github.com/hashgraph/hedera-sdk/pull/19/files


src/NetworkName.js
```diff
@@ -1,60 +1,175 @@
...
+        return new Uint8Array(
+            typeof networkId != "string" ? [networkId] : [parseInt(networkId)]
+        );
+    }
+
+    /**
+     * @returns {Uint8Array}
+     */
+    toProtobuf() {
+        // TODO
```
valtyr-naut, line 81: thoughts? @danielakhterov

src/NetworkName.js
```diff
@@ -1,60 +1,175 @@
...
+        // rename?
+        return new Uint8Array([this._networkId]);
+    }
+
+    /**
+     * @param {Uint8Array} networkName
+     * @returns {NetworkName}
+     */
+    static fromProtobuf(networkName) {
+        // TODO
```
valtyr-naut, line 91: thoughts? @danielakhterov

src/client/ManagedNetwork.js
```diff
@@ -111,11 +107,11 @@ export default class MangedNetwork {
     }
 
     /**
-     * @param {NetworkName} networkName
+     * @param {string} networkName
      * @returns {this}
      */
     setNetworkName(networkName) {
-        this._ledgerId = _ledgerIdToLedgerId(networkName);
+        this._ledgerId = NetworkName.networkIdFromName(networkName).toString();
```
danielakhterov, line 114: Change `this._ledgerId` to be of type `NetworkName` instead and stringify it only when needed.

src/Node.js
```diff
@@ -85,7 +85,7 @@ export default class Node extends ManagedNode {
      * @returns {this}
      */
     setCert(ledgerId) {
```
danielakhterov, line 87: Perhaps change this to require `ledgerId` to be of type `NetworkName` instead?

danielakhterov (changes requested):

src/NetworkName.js
danielakhterov: I believe this was before we talked, but we don't really need `NetworkName.[to|from]Protobuf()` However, if you'd like to implement it it's fine. The type this should encode/decode to should be `Uint8Array`.

src/Node.js
valtyr-naut: changed to use networkName only not ledgerId to convert to networkName

src/LedgerId.js
```diff
@@ -0,0 +1,123 @@
...
+        return this._ledgerId;
+    }
+
+    /**
+     * @param {string} ledgerId
+     * @returns {LedgerId}
+     */
+    static fromString(ledgerId) {
+        ledgerId =
+            hex.decode(ledgerId).toString().length >= 1
```
danielakhterov, line 34: Probably should put `hex.decode(ledgerId)` into a separate variable since you're doing this twice. I haven't read through your tests yet, but this looks like it'd throw an error if the string is `mainnet`.

src/NetworkName.js
```diff
@@ -1,60 +1,136 @@
...
-    Testnet: "testnet",
-    Previewnet: "previewnet",
-};
+export default class NetworkName {
+    /**
+     * @hideconstructor
+     * @internal
+     * @param {string} networkName
+     */
+    constructor(networkName) {
```
danielakhterov, line 14: Might be better to require the parameter to be of type `number` where 0 = mainnet, 1 = testnet, 2 = previewnet, and all else is treated as other.

src/client/ManagedNetwork.js
```diff
@@ -111,11 +107,11 @@ export default class MangedNetwork {
     }
 
     /**
-     * @param {NetworkName} networkName
+     * @param {string} networkName
      * @returns {this}
      */
     setNetworkName(networkName) {
-        this._ledgerId = _ledgerIdToLedgerId(networkName);
+        this._ledgerId = NetworkName.toId(networkName).toString();
```
danielakhterov, line 114: This should probably be `NetworkName.fromString(networkname)`.

src/client/NativeClient.js
```diff
@@ -56,17 +56,17 @@ export default class NativeClient extends Client {
                 switch (props.network) {
                     case "mainnet":
                         this.setNetwork(Network.MAINNET);
-                        this.setNetworkName("mainnet");
+                        this.setNetworkName(NetworkName.NETNAMES[0]);
```
danielakhterov, line 59: I think if you use `NetworkName.MAINNET` it might be better since `NETNAME[0]` isn't very descriptive.

src/client/Network.js
```diff
@@ -112,7 +108,7 @@ export default class Network extends ManagedNetwork {
      */
     get networkName() {
         return this._ledgerId != null
```
danielakhterov, line 110: If you change `this._ledgerId` to be of type `NetworkName` you wouldn't need to convert it in the conditional.

test/unit/LedgerId.js
```diff
@@ -0,0 +1,86 @@
...
+    const MAINNET = "mainnet";
+    const TESTNET = "testnet";
+    const PREVIEWNET = "previewnet";
+
+    it('fromString returns LedgerId from strings/hex "mainnet", "testnet", "previewnet"', function () {
+        let ledgerId = LedgerId.fromString("0");
+        expect(ledgerId.ledgerId).to.eql(new Uint8Array([0]));
+        ledgerId = LedgerId.fromString(MAINNET);
+        expect(ledgerId.ledgerId).to.eql(new Uint8Array([0]));
+        ledgerId = LedgerId.fromString(hex.encode(MAINNET));
```
danielakhterov, line 14: Um, this should not work. `hex.encode()` requires `Uint8Array` to be provided not `string`. If this does not error that's unfortunate.

src/LedgerId.js
```diff
@@ -0,0 +1,123 @@
...
+         */
+        this._ledgerId = ledgerId;
+
+        Object.freeze(this);
+    }
+
+    /**
+     * @returns {Uint8Array}
+     */
+    get ledgerId() {
```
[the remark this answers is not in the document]
valtyr-naut: removed

src/LedgerId.js
valtyr-naut: it doesnt error. its also memory vs time thing. since both are miniscule i went with making the call twice instead of making a new var.

src/LedgerId.js
```diff
@@ -0,0 +1,125 @@
...
+     * @returns {boolean}
+     */
+    isPreviewnet() {
+        return this.toString() == LedgerId.NETNAMES[2];
+    }
+
+    /**
+     * @returns {boolean}
+     */
+    isOther() {
```
danielakhterov, line 112: Should not be present.

src/LedgerId.js
```diff
@@ -0,0 +1,125 @@
...
+    }
+}
+
+LedgerId.MAINNET = new LedgerId(new Uint8Array([0]));
+
+LedgerId.TESTNET = new LedgerId(new Uint8Array([1]));
+
+LedgerId.PREVIEWNET = new LedgerId(new Uint8Array([2]));
+
+LedgerId.OTHER = new LedgerId(new Uint8Array([3]));
```
danielakhterov, line 123: This and `NETNAMES` should not be part of the public API. I'd commend using variables to store these.

danielakhterov (changes requested):

src/LedgerId.js
danielakhterov: Not necessarily for performance reasons, but for code cleanness.

src/NetworkName.js
```diff
@@ -1,60 +1,134 @@
...
 }
 
-export default NetworkName;
+const NETNAMES = ["mainnet", "testnet", "previewnet", "other"];
+
+const MAINNET = new NetworkName("mainnet");
+
+const TESTNET = new NetworkName("testnet");
+
+const PREVIEWNET = new NetworkName("previewnet");
```
danielakhterov, line 132: The comment made before was only for `NETNAMES`, you'd still want to declare this and the other `*NET` types as `NetworkName.MAINNET = new NetworkName("mainnet");`

src/client/NodeClient.js
```diff
@@ -114,19 +113,19 @@ export default class NodeClient extends Client {
                     case "mainnet":
                         this.setNetwork(Network.MAINNET);
                         this.setMirrorNetwork(MirrorNetwork.MAINNET);
-                        this.setNetworkName(NetworkName.Mainnet);
+                        this.setNetworkName("mainnet");
```
danielakhterov, line 116: Probably should keep the former after you've made `NetworkName.MAINNET` again.

danielakhterov (changes requested):

src/LedgerId.js
```diff
@@ -0,0 +1,134 @@
...
+
+    /**
+     * If the ledger ID is a known value such as `[0]`, `[1]`, `[2]` this method
+     * will instead return "mainnet", "testnet", or "previewnet", otherwise it will
+     * hex encode the bytes.
+     *
+     * @returns {string}
+     */
+    toString() {
+        if (this._ledgerId.length > 1) {
```
danielakhterov, line 53: I believe this is supposed to be `this._ledgerId.length === 0` since that's the only reason `this._ledgerId[0]` return `undefined`

src/LedgerId.js
```diff
@@ -0,0 +1,134 @@
...
+                return hex.encode(this._ledgerId);
+        }
+    }
+
+    /**
+     * @returns {string}
+     */
+    _toStringForChecksum() {
+        //legacy checksum logic from NetworkName
+        if (this._ledgerId.length > 1) {
```
danielakhterov, line 73: Same for this

src/client/Client.js
```diff
@@ -108,24 +109,49 @@ export default class Client {
...
     /**
+     * @deprecated
      * @param {NetworkName} networkName
      * @returns {this}
      */
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
     setNetworkName(networkName) {
-        this._network.setNetworkName(networkName);
-        return this;
+        //uses custom NetworkName type
```
danielakhterov, line 117: Not going to block this PR for this, but in the future can we use `//<space><comment>` instead of `//<comment>` for consistency. That or use JSDoc comments.

src/exports.js
```diff
@@ -47,7 +47,6 @@ export { default as LiveHash } from "./account/LiveHash.js";
 export { default as LiveHashAddTransaction } from "./account/LiveHashAddTransaction.js";
 export { default as LiveHashDeleteTransaction } from "./account/LiveHashDeleteTransaction.js";
 export { default as LiveHashQuery } from "./account/LiveHashQuery.js";
-export { default as NetworkName } from "./NetworkName.js";
```
danielakhterov, line 50: Can't remove this unfortunately, it'd be a breaking change. If you'd like you can just export whatever this was exporting here directly without re-creating the file.

src/token/TokenNftInfo.js
```diff
@@ -24,10 +25,11 @@ export default class TokenNftInfo {
     /**
      * @private
      * @param {object} props
-     * @param {NftId} props.nftId;
-     * @param {AccountId} props.accountId;
-     * @param {Timestamp} props.creationTime;
-     * @param {Uint8Array | null} props.metadata;
```
danielakhterov, line 30: Heh, good catch with the semicolons.

test/unit/LedgerId.js
```diff
@@ -0,0 +1,117 @@
+import * as hex from "../src/encoding/hex.js";
+import LedgerId from "../src/LedgerId.js";
+
+const MAINNET = "mainnet";
+const HEX_MAINNET = hex.encode(new Uint8Array([0]));
```
danielakhterov, line 5: Not blocking this PR for this, but I'm curious why you use `hex.encode(...)` here instead of just constructing the hex version yourself `"00"`

test/unit/LedgerId.js
```diff
@@ -0,0 +1,117 @@
...
+
+const PREVIEWNET = "previewnet";
+const HEX_PREVIEWNET = hex.encode(new Uint8Array([2]));
+
+const OTHER = "other";
+const HEX_OTHER = hex.encode(new Uint8Array([3]));
+
+describe("LedgerId", function () {
+    it('fromString returns LedgerId from strings/hex "mainnet", "testnet", "previewnet"', function () {
+        expect(LedgerId.fromString(MAINNET).toString()).to.eql(MAINNET);
```
danielakhterov, line 18: Not blocking the PR for this, but I might be better to save `LedgerId.fromString(MAINNET)` into a variable and compare it. I'm gong to assume that calling `LedgerId.fromString()` is slower. That shouldn't matter too much here as I think it'll still be fast, but just for future.

test/unit/LedgerId.js
```diff
@@ -0,0 +1,117 @@
...
+        expect(ledgerId.isTestnet()).to.eql(false);
+        expect(ledgerId.isMainnet()).to.eql(false);
+
+        ledgerId = LedgerId.fromBytes(new Uint8Array([3]));
+        expect(ledgerId.isPreviewnet()).to.eql(false);
+        expect(ledgerId.isTestnet()).to.eql(false);
+        expect(ledgerId.isMainnet()).to.eql(false);
+    });
+
+    it("toString throws error when length > 1", function () {
```
danielakhterov, line 108: This should not be the case. `LedgerId.toString()` should work just fine if the bytes are longer than 1, but if it is exactly 1 we should check to see if the value is `0`, `1`, or `2` in which case we'd print the the network's name instead of hex: `0` => `"mainnet"`, etc.

danielakhterov (changes requested):
Pretty much complete except some small misunderstandings.

test/unit/LedgerId.js
valtyr-naut: 🤷

src/exports.js
```diff
@@ -47,7 +47,7 @@ export { default as LiveHash } from "./account/LiveHash.js";
...
+export { default as NetworkName } from "./LedgerId.js";
```
danielakhterov, line 50:
```suggestion
/**
 * @typedef {object} NetworkNameType
 * @property {ClientNetworkName} Mainnet
 * @property {ClientNetworkName} Testnet
 * @property {ClientNetworkName} Previewnet
 */

/**
 * @type {NetworkNameType}
 */
export const NetworkName = {
    Mainnet: "mainnet",
    Testnet: "testnet",
    Previewnet: "previewnet",
};
```

danielakhterov (changes requested):

danielakhterov (approved):

PR #953, opened 2022-02-14
Fix Transaction.freezeWith followed by PrivateKey.signTransaction exception scenario 

3Nigma:
**Description**:
When freezing a `Transaction` with a `Client` we also try to [set the node-account-ids for that transaction](https://github.com/hashgraph/hedera-sdk-js/blob/b9d8f1c97c3d201e1f2b56e366a04b388c5ba005/src/transaction/Transaction.js#L732). 

If the transaction doesn't have them set, [eg via `setNodeAccountIds`](https://github.com/hashgraph/hedera-sdk-js/blob/b9d8f1c97c3d201e1f2b56e366a04b388c5ba005/examples/sign-transaction.js#L49):
```js
const transferTransaction = new TransferTransaction()
    .setNodeAccountIds([new AccountId(3)])
    .addHbarTransfer(receipt.accountId, -1)
    .addHbarTransfer("0.0.3", 1)
    .freezeWith(client);
```
then [we take them from the client network](https://github.com/hashgraph/hedera-sdk-js/blob/b9d8f1c97c3d201e1f2b56e366a04b388c5ba005/src/transaction/Transaction.js#L683):
```js
this._nodeAccountIds.setList(
    client._network.getNodeAccountIdsForExecute()
);
```
which, in term, [looks at the `Network.getNumberOfNodesForTransaction()`](https://github.com/hashgraph/hedera-sdk-js/blob/b9d8f1c97c3d201e1f2b56e366a04b388c5ba005/src/client/Network.js#L237) and [_should_ take only the `count` nodes from the current network](https://github.com/hashgraph/hedera-sdk-js/blob/b9d8f1c97c3d201e1f2b56e366a04b388c5ba005/src/client/ManagedNetwork.js#L187). 

This doesn't happen (basically, `count` is neglected and the full list of nodes is returned).

If the developer then goes and [does a `PrivateKey.signTransaction(transaction)`](https://github.com/hashgraph/hedera-sdk-js/blob/b9d8f1c97c3d201e1f2b56e366a04b388c5ba005/src/PrivateKey.js#L260) which, in term, calls `Transaction.addSignature`, [the first thing that's checked](https://github.com/hashgraph/hedera-sdk-js/blob/b9d8f1c97c3d201e1f2b56e366a04b388c5ba005/src/transaction/Transaction.js#L581) is that only one `this._nodeAccountIds` is present. This is, of course, false and the operation ends up throwing a `transaction did not have exactly one node ID set` message.

The fix is a minor one concerning [a forgotten set-variable](https://github.com/hashgraph/hedera-sdk-js/blob/b9d8f1c97c3d201e1f2b56e366a04b388c5ba005/src/client/ManagedNetwork.js#L198) not being updated in the `Network._getNumberOfMostHealthyNodes` function.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


danielakhterov (approved):
Good catch, and fantastic write up. Thank you for the PR.

Issue #967, opened 2022-02-24
Add ExchangeRateSet to the SDK so that file 0.0.112 can be parsed


gregscullard:
### Problem

The file `0.0.112` contains exchange rate information, it's not possible to parse it with the SDK.

### Solution

Add an `ExchangeRateSet` class to enable parsing of the file

PR #968, opened 2022-02-24
Adds ExchangeRateSet.js

gregscullard:
**Description**:
ExchangeRateSet is missing from the SDK making it impossible to parse file `0.0.112` for exchange rate information.

Fixes #
#967 

**Notes for reviewer**:
Only adds the `ExchangeRateSet` class.

**Checklist**


danielakhterov (commented):
I believe there is a file added by mistake `dist/assets/*` However I have the same question as in Java which is should we rename the set class to make it easier to understand?

danielakhterov:
Also can you target develop instead of main. We want all new feature to hit develop first.

gregscullard:
> Also can you target develop instead of main. We want all new feature to hit develop first.

Done

gregscullard:
> I believe there is a file added by mistake `dist/assets/*` However I have the same question as in Java which is should we rename the set class to make it easier to understand?

Removed

gregscullard:
> I believe there is a file added by mistake `dist/assets/*` However I have the same question as in Java which is should we rename the set class to make it easier to understand?

> I believe there is a file added by mistake `dist/assets/*` However I have the same question as in Java which is should we rename the set class to make it easier to understand?

Removed (maybe the dist folder should be added to .gitIgnore ?)

gregscullard:
> Also can you target develop instead of main. We want all new feature to hit develop first.

Done

danielakhterov (approved):

Issue #966, opened 2022-02-24
ExchangeRate returns incorrect expirationTime


gregscullard:
### Description

The `expirationTime` of the `exchangeRate` in a receipt is incorrect.

For example, the expiration time returned from a receipt is
`{
  "low": 1645714800,
  "high": 0,
  "unsigned": false
}`

which should resolve to `Thu Feb 24 2022 15:00:00`

is in fact `1970-01-20T01:08:34.800Z` in the receipt

### Steps to reproduce

Run a transaction, get a receipt, print the `receipt.exchangeRate.expirationTime`

### Hedera network

other

### Version

v2.9.1

danielakhterov:
Fixed: here: https://github.com/hashgraph/hedera-sdk-js/commit/257b21e6590ffceb784378dd59b7adc238c9ff3d

PR #980, opened 2022-03-03
Fix for incorrect dates returned in receipts from ExchangeRate.js

keirabee:
**Description**:
This should fix the incorrect receipt problem with Exchange.js.  

This PR modifies exchange.js' expiration time to return time in seconds (default time for getTime() is in millis, so this is divided by 1000) instead of just the seconds field. Added a unit test to demonstrate the fix.
**Related issue(s)**:
https://github.com/hashgraph/hedera-sdk-js/issues/966

**Notes for reviewer**:

**Checklist**

- [n/a ] Documented (Code comments, README, etc.)
- [ ✓] Tested (unit, integration, etc.)

### DIFF
--- a/test/unit/ExchangeRates.js
+++ b/test/unit/ExchangeRates.js
@@ -1,7 +1,7 @@
 import "mocha";
 import { expect } from "chai";
-
-import { ExchangeRates } from "../../src/exports.js";
+import Long from "long";
+import { ExchangeRates, ExchangeRate } from "../../src/exports.js";
 
 const exchangeRateSetBytes = Buffer.from(
     "0a1008b0ea0110b6b4231a0608f0bade9006121008b0ea01108cef231a060880d7de9006",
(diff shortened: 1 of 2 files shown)


test/unit/ExchangeRates.js
```diff
@@ -37,3 +37,14 @@ describe("ExchangeRates", function () {
...
 });
+
+describe("ExchangeRate", function () {
+    it("fromBytes", function () {
+         const expirationTime = new Date("February 24, 2022 15:00:00 UTC");
+         const props = {
+             expirationTime: expirationTime,
+           }
+         const exchangeRate = new ExchangeRate(props);
+         expect(exchangeRate._toProtobuf().expirationTime.seconds).to.deep.equal(Long.fromValue({ low: 1645714800, high: 0, unsigned: false }));
```
danielakhterov, line 48:
```suggestion
        const exchangeRate = ExchangeRate._fromProtobuf(new ExchangeRate({ expirationTime })._toProtobu());
        expect(exchangeRate.expirationTime).to.be.equal(expirationTime);
```
I think it's important to do both `ExchangeRate._[to|from]Protobuf()` to confirm it the round trip works as expected. I believe you comparing to `1645714800` is actually comparing to the wrong value according to the issue; unless this test was to confirm the we're producing the wrong value in which case you can ignore this sentence.

keira-black: I think you're right. I'll update the tests and recheck.

danielakhterov:
Going to just merge, and fix directly on `develop`

PR #1060, opened 2022-04-08
1055 hip 336 update approvalallowance apis

### LINKED ISSUE #1055: HIP-336: Update approval/allowance APIs
### Problem

Changes have been made to the API that the SDK needs to support.

- We are removing the `CryptoAdjustAllowance` transaction API. Users will use the `CryptoApproveAllowance` transaction to add hbar, fungible, and non-fungible tokens. This transaction will also be used to remove or adjust hbar and fungible token allowances from a spender account
- We are removing the `CryptoDeleteAllowance` transaction for hbar and fungible token allowances. This transaction will be used to remove NFT allowances from a spender account
- The two transactions that will be available to users are `CryptoApproveAllowance` and `CryptoDeleteAllowance`
- This [PR](https://github.com/hashgraph/hedera-protobufs/pull/181) covers the protobuf updates required for this change

Please update the example in the SDK accordingly.

### Solution

Add support for the specified changes

### Alternatives

_No response_

easpaas:
**Description**:

This PR modifies the existing `AccountAllowanceAdjustTransaction` by removing the functionality from within the SDK. 

**Related issue(s)**:

Fixes #1055 

**Notes for reviewer**:

**Checklist**


src/account/AccountAllowanceAdjustTransaction.js
```diff
@@ -1,534 +1,534 @@
-import Transaction, {
```
danielakhterov, line 1: Unfortunately, because this is already on `main` and has been released we still need to keep this type around. What we should do instead of commenting everything out, is remove all our internal methods. This would make it so executing this request would error, but users can still construct and update this request otherwise.

src/exports.js
```diff
@@ -9,7 +9,6 @@ export { default as KeyList } from "./KeyList.js";
 export { default as Key } from "./Key.js";
 export { default as Mnemonic } from "./Mnemonic.js";
 
-export { default as AccountAllowanceAdjustTransaction } from "./account/AccountAllowanceAdjustTransaction.js";
```
danielakhterov, line 12: Cannot remove the type :(

test/unit/AccountAllowanceAdjustTransaction.js
```diff
@@ -1,99 +1,99 @@
-import { expect } from "chai";
```
danielakhterov, line 1: Lets just delete this entire file.

danielakhterov (changes requested):

easpaas (commented):
@danielakhterov could you verify that I have removed the necessary internal methods for `AccountAllowanceAdjustTransactions`?

danielakhterov (changes requested):
What we need to do is make sure that `AccountAllowanceAdjustTransaction` works exactly as before until the user attempts to freeze or execute the request. This means `AccountAllowanceAdjustTransaction.grantHbarAllowance(ownerAccountId, spenderAccountId, amount)` should result in `AccountAllowanceTransaction.hbarAllowance` being populated with a list of one which contains an `HbarAllowance` with `ownerAccountId`, `spenderAccountId`, and `amount` equal to the parameters provided.

easpaas (commented):
@danielakhterov I removed _fromProtobuf and replaced the internal methods with throwable error messages. However, I am not sure about the best UX for the error messages?

danielakhterov (approved):

danielakhterov:
I'm ignoring CI as I'm working on fixing it in another branch :/

danielakhterov:
CI is being fixed in another PR #1072 so for now I will just run tests locally for confirmation and merge.

Issue #1032, opened 2022-04-01
Guide for using signature provider feature for wallet developers


SimiHunjan:
### Problem

We need a guide that outlines the basic steps of creating a provider.

### Solution

Create guide.

easpaas:
@SimiHunjan @danielakhterov Are there other guides that I can reference for this? At the moment I think this ticket is too ambiguous to make any progress.

SimiHunjan:
@easpaas this was to document a guide for wallet developers in how they would use or what steps they would take to take advantage of the signature provider feature in the SDK. This is the HIP for that feature https://hips.hedera.com/hip/hip-338. This can be a markdown file. We were planning to host it at docs.hedera.com. The markdown file can be shared with me to add to that site.

PR #1074, opened 2022-04-15
feat(wip): local signer example that talks to a server

danielakhterov:

**Description**:

**Related issue(s)**:

#1032 

**Notes for reviewer**:

**Checklist**


local_signer_example/src/client.js
```diff
@@ -0,0 +1,425 @@
...
+    }
+
+    /**
+     * @param {Transaction} transaction
+     * @returns {Promise<Transaction>}
+     */
+    async signTransaction(transaction) {
+        /** @type {LocalProviderResponse} */
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+        const response = (
```
danielakhterov, line 314: @sergmetelin I'm conflicted on this implementation; the issue here is that `LocalSigner.signTransaction()` communicates with the server without going through `LocalProvider`. `LocalSigner` should **only** use `this.provider` for communication, but because `Provider` doesn't expose any a `sign` method we need to implement the communication protocol in `LocalSigner.signTransaction()`. We could get around this by simply adding a `signTransaction` method to the `Provider` interface, but that sort of goes against ethers.io Speaking of which I'm not completely sure if ethers.io even supports remote signing, so maybe we shouldn't either?

cc @gregscullard in case you have any ideas for this

PR #1083, opened 2022-04-18
feat: implement `ContractEvmTransaction`

danielakhterov:

**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
     "dependencies": {
         "@grpc/grpc-js": "^1.6.3",
         "@hashgraph/cryptography": "^1.1.2",
-        "@hashgraph/proto": "2.4.0",
+        "@hashgraph/proto": "link:packages/proto",
         "bignumber.js": "^9.0.2",
         "crypto-js": "^4.1.1",
         "js-base64": "^3.7.2",
--- a/src/contract/ContractCreateTransaction.js
+++ b/src/contract/ContractCreateTransaction.js
@@ -53,6 +53,7 @@ export default class ContractCreateTransaction extends Transaction {
     /**
      * @param {object} [props]
      * @param {FileId | string} [props.bytecodeFileId]
+     * @param {Uint8Array} [props.initcode]
      * @param {Key} [props.adminKey]
      * @param {number | Long} [props.gas]
      * @param {number | string | Long | BigNumber | Hbar} [props.initialBalance]
@@ -69,6 +70,12 @@ export default class ContractCreateTransaction extends Transaction {
          * @type {?FileId}
          */
         this._bytecodeFileId = null;
+
+        /**
+         * @private
+         * @type {?Uint8Array}
+         */
+        this._initcode = null;
 
         /**
          * @private
@@ -116,6 +123,10 @@ export default class ContractCreateTransaction extends Transaction {
 
         if (props.bytecodeFileId != null) {
             this.setBytecodeFileId(props.bytecodeFileId);
+        }
+
+        if (props.initcode != null) {
+            this.setInitcode(props.initcode);
         }
 
         if (props.adminKey != null) {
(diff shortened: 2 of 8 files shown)


src/contract/ContractCreateTransaction.js
```diff
@@ -70,6 +71,12 @@ export default class ContractCreateTransaction extends Transaction {
          */
         this._bytecodeFileId = null;
 
+        /**
+         * @private
+         * @type {?Uint8Array}
+         */
+        this._initcode = null;
```
danielakhterov, line 78: Rename to `bytecode`

src/EthereumTransaction.js
```diff
@@ -0,0 +1,254 @@
...
+ * @typedef {import("./account/AccountId.js").default} AccountId
+ * @typedef {import("./channel/Channel.js").default} Channel
+ * @typedef {import("./client/Client.js").default<*, *>} Client
+ * @typedef {import("./Timestamp.js").default} Timestamp
+ * @typedef {import("./transaction/TransactionId.js").default} TransactionId
+ * @typedef {import("long").Long} Long
+ */
+
+/**
+ * Create a new Hedera™ crypto-currency account.
```
Nana-EC, line 49:
```suggestion
 * Create a new Hedera™ transaction wrapped ethereum transaction.
```

src/EthereumTransaction.js
```diff
@@ -0,0 +1,254 @@
...
+ */
+export default class EthereumTransaction extends Transaction {
+    /**
+     * @param {object} [props]
+     * @param {Uint8Array | FileId} [props.ethereumData]
```
Nana-EC, line 54: I'm not sure these are mutually exclusive.
Is it not the case that the RLP encoded `ethereumData.callData` will be set to empty in services where CallData is specified.
In that case a user should be able to specify both.
Which means you should have `ethereumData` and `callData` params

src/EthereumTransaction.js
```diff
@@ -0,0 +1,254 @@
...
+    ) {
+        const body = bodies[0];
+        const transaction =
+            /** @type {HashgraphProto.proto.IEthereumTransactionBody} */ (
+                body.ethereumTransaction
+            );
+
+        let ethereumData = undefined;
+        if (transaction.callData != null) {
+            ethereumData = FileId._fromProtobuf(transaction.callData);
```
Nana-EC, line 105: this should be setting a callData variable

src/EthereumTransaction.js
Nana-EC: Would then also need to add a setter and getter for callData

src/contract/ContractCreateTransaction.js
```diff
@@ -237,6 +248,24 @@ export default class ContractCreateTransaction extends Transaction {
...
+        return this._initcode;
+    }
+
+    /**
+     * @param {Uint8Array} initcode
+     * @returns {this}
+     */
+    setInitcode(initcode) {
+        this._requireNotFrozen();
+        this._initcode = initcode;
```
Nana-EC, line 264: Q: Does this require some kind of OneOf check to ensure either `initCode` or `fileID`) is set and not both

Nana-EC (changes requested):
I think we're missing some `callData` input logic here

src/contract/ContractCreateTransaction.js
danielakhterov: I believe this is handled by `protobufjs`. If you set both fields it only encodes one. Perhaps we should set the the other field to null manually, but that may not be obvious to the users?

src/EthereumTransaction.js
danielakhterov: Huh, I think you're right. I believe I misread the protobuf documentation when we went with this API. Will fix.

package.json
```diff
@@ -53,7 +53,7 @@
     "dependencies": {
         "@grpc/grpc-js": "^1.6.7",
         "@hashgraph/cryptography": "^1.1.2",
-        "@hashgraph/proto": "^2.4.1",
+        "@hashgraph/proto": "link:packages/proto",
```
Nana-EC, line 56: Is proto still behind or can this be updated?

Nana-EC (approved):
LGTM

package.json
danielakhterov: This can be updated, and it has been updated in the release branches. Will make sure to update this here as well.

danielakhterov:
Merged via cli because GitHub is being rude.

Issue #1130, opened 2022-05-19
`setCallDataFileId` used in Java SDK where JS is using `setCallData`


SimiHunjan:
### Description

The goal has always been to keep the method names consistent across all SDKs.

Java: https://github.com/hashgraph/hedera-sdk-java/blob/main/sdk/src/main/java/com/hedera/hashgraph/sdk/EthereumTransaction.java#L71

JS: https://github.com/hashgraph/hedera-sdk-js/blob/main/src/EthereumTransaction.js#L174

### Steps to reproduce

See above

### Hedera network

other

### Version

2.14.2

PR #1132, opened 2022-05-19
Updated symbol name per bug from setCallData to setCallDataFileId

keira-black:
Signed-off-by: Keira Black <keirablack@Keiras-MacBook-Pro.local>

**Description**:

**Related issue(s)**:

Fixes #
#1130 
**Notes for reviewer**:

**Checklist**

- [ n/a] Documented (Code comments, README, etc.)
- [ X] Tested (unit, integration, etc.)

### DIFF
--- a/src/EthereumTransaction.js
+++ b/src/EthereumTransaction.js
@@ -81,7 +81,7 @@ export default class EthereumTransaction extends Transaction {
         }
 
         if (props.callData != null) {
-            this.setCallData(props.callData);
+            this.setCallDataFileId(props.callData);
         }
 
         if (props.maxGasAllowance != null) {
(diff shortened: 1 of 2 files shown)


src/EthereumTransaction.js
```diff
@@ -171,7 +171,7 @@ export default class EthereumTransaction extends Transaction {
      * @param {FileId} callData
      * @returns {this}
      */
-    setCallData(callData) {
+    setCallDataFileId(callData) {
```
danielakhterov, line 174: Unfortunately we cannot just change the name here because it's already release. We need to create a new method and deprecate the old one.

danielakhterov (changes requested):

PR #1140, opened 2022-05-27
chore: add some more documentation to the SDK

danielakhterov:


danielakhterov:
Since this is just adding to the internal documentation, I'll just merge it. If we find issues with the documentations I'd still like to fix them, but I'll do that in another PR.

PR #1141, opened 2022-05-27
feat: implement time drift syncing

### LINKED ISSUE #1103: Transaction needs to be created using a third party time source
### Problem

By far the most user issues we field are from the INVALID_TRANSACTION_START error, and every single time it is because the users computer clock is wrong.

As a wallet, we do not always create the transactions, sometimes they come from external dapps. Please consider using some sort of third party clock rather than the local one.

### Solution

Use third party time source

### Alternatives

_No response_

danielakhterov:

**Description**:

**Related issue(s)**:

Fixes #1103

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/src/sync.js
+++ b/src/sync.js
@@ -1,14 +1,37 @@
 import axios from "axios";
 import Cache from "./Cache.js";
+import Logger from "js-logger";
 
 const sync = (async () => {
-    const response = await axios.get(
-        "http://worldtimeapi.org/api/timezone/gmt"
-    );
-    const currentTime = Math.round(Date.now() / 1000);
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment,@typescript-eslint/no-unsafe-member-access
-    const worldTime = response.data.unixtime;
-    Cache.timeDrift = worldTime - currentTime;
+    // http://time.google.com:80 doesn't actually give us an NTP response, instead it returns
+    // a 302 redirected response. However, it does contain a `date` header which we can use.
+    try {
+        const response = await axios.get("http://time.google.com", {
+            maxRedirects: 0,
+            
+            // By default a status of 302 is considered an erring status
+            validateStatus: function (status) {
+                return status < 500;
+            },
+        });
+
+        const currentTime = Math.round(Date.now() / 1000);
+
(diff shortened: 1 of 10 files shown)


src/sync.js
```diff
@@ -0,0 +1,14 @@
+import axios from "axios";
+import Cache from "./Cache.js";
+
+const sync = (async () => {
+    const response = await axios.get(
+        "http://worldtimeapi.org/api/timezone/gmt"
```
willparks, line 6: After looking into WorldTimeAPI, it looks like it is not recommended for production environments. ([source](http://worldtimeapi.org/pages/faqs#please-do-not))

It turns out that NTP servers [use the UDP communications protocol](https://datatracker.ietf.org/doc/html/rfc5905#page-18), which excludes being able to communicate with them via the browser, ruling direct NTP communication out as a solution.

Following our discussion, grabbing the `Date` header off a response from [time.google.com](time.google.com) looks to be a good workaround.

src/sync.js
```diff
@@ -0,0 +1,37 @@
...
+            validateStatus: function (status) {
+                return status < 500;
+            },
+        });
+
+        const currentTime = Math.round(Date.now() / 1000);
+
+        // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date
+        // strongly discourages us using `Date.parse()`, but I'm not sure what we should replace
+        // it with without adding any new deps.
```
willparks, line 22: The only other alternative (without resorting to moment or date-fns) I see is to [manually parse](https://jsfiddle.net/willparks/ber8f0hc/32/) the Date header, but it is a bit verbose. Otherwise, I'm not sure what else can be done here either.

willparks (approved):

src/sync.js
danielakhterov: I would really like to avoid hand rolling parsing tbh. If we end up having trouble with `Date.parse()` we might end up hand rolling it or adding a library for it, but for the time being I think continuing to use `Date.parse()` is fine?

willparks: I completely agree, `Date.parse()` is working perfectly fine - no sense in reinventing the wheel :)

PR #1152, opened 2022-06-13
feat(wip): TCK

danielakhterov:

**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


rbair23 (approved):
There is some work to do, which can be tackled in subsequent PRs, or in this one. We really do want real transactions that will pass, and we really do want to verify the results in the mirror node. We should also consider adding an example for Jest or whatnot. Maybe find out what some of the wallets are doing and then add examples for them to follow. We should also, if we haven't already, have some markdown files to describe how to run the tests.

PR #1187, opened 2022-07-07
Possible fix. changed 'freeze' to 'seal' in constructor.

### LINKED ISSUE #1186: Child transaction creation not working with the JS SDK
### Description

Documentation on parent/child transactions shows the method `.setNonce()` for `TransactionId`
https://docs.hedera.com/guides/docs/sdks/transactions/transaction-id

That method is not working as expected in the JS SDK. A user reported the issue and I'm also running into the same issue. 
Here's the error" `TypeError: Cannot assign to read only property 'nonce' of object '[object Object]'`

### Steps to reproduce

The following code is a starting point for debugging:

	const txId1 = TransactionId.generate(new AccountId(operatorId)).setNonce(0);
	// const txId2 = txId1.setNonce(1);
	const txId2 = TransactionId.fromString(txId1.toString()).setNonce(1);

	console.log(`${txId1} and ${txId2}`);

### Additional context

_No response_

### Hedera network

testnet

### Version

2.17 and 2.14

### Operating system

_No response_

keirabee:
**Description**:
setNonce isn't working on main because the constructor freezes the object during creation. I changed the line to seal the object instead which still allows writing to writeable fields.

This PR modifies TransactionID in order to support writeability generally

**Related issue(s)**:

Fixes #1186 

**Notes for reviewer**:

**Checklist**

- [ ] Documented (Code comments, README, etc.) - n/a just a one line change
- [X] Tested (unit, integration, etc.) - ran unit tests, they passed okay. The example code given to demo the error on 1186 passes after this change and there are no new 'bugs' as such.

### DIFF
--- a/src/transaction/TransactionId.js
+++ b/src/transaction/TransactionId.js
@@ -65,7 +65,7 @@ export default class TransactionId {
             this.setNonce(nonce);
         }
 
-        Object.freeze(this);
+        Object.seal(this);
     }
 
     /**


keirabee:
Just pushed up a unit test

danielakhterov (approved):

PR #1193, opened 2022-07-20
feat: support `Mnemonic` -> `EcdsaPrivateKey`

### LINKED ISSUE #1177: Add ECDSA key support to Mnemonic logic
### Problem

Currently [Mnemonic logic](https://github.com/hashgraph/hedera-sdk-js/blob/main/packages/cryptography/src/Mnemonic.js) inherently supports Ed25519 key. 
Usage example [generate-key.js](https://github.com/hashgraph/hedera-sdk-js/blob/main/examples/generate-key.js)

With the expansion and need for great ECDSA support this becomes a limiting factor on wallets

### Solution

Add configuration to support ECDSA keys also
Expand `Mnemonic.toPrivateKey(passphrase)` to take in a keyType that is used to affect the hash algorithm used and key type returned

### Alternatives

Add `Mnemonic.toEcdsaPrivateKey(passphrase)` and `Mnemonic.toEd25519PrivateKey(passphrase)` with `Mnemonic.toPrivateKey(passphrase)` defaulting to the latter

danielakhterov:
Signed-off-by: Daniel Akhterov <akhterovd@gmail.com>

**Description**:

**Related issue(s)**:

Fixes #1177 

**Notes for reviewer**:

**Checklist**


packages/cryptography/src/Mnemonic.js
```diff
@@ -6,11 +6,18 @@ import legacyWords from "./words/legacy.js";
...
-import * as pbkdf2 from "./primitive/pbkdf2.js";
 import * as hmac from "./primitive/hmac.js";
 import * as slip10 from "./primitive/slip10.js";
+import * as bip32 from "./primitive/bip32.js";
+import * as bip39 from "./primitive/bip39.js";
 import * as entropy from "./util/entropy.js";
 import * as random from "./primitive/random.js";
+import EcdsaPrivateKey from "./EcdsaPrivateKey.js";
+import * as ecdsa from "./primitive/ecdsa.js";
+// import * as hex from "./encoding/hex.js";
```
Nana-EC, line 17: nit: remove

packages/cryptography/src/Mnemonic.js
```diff
@@ -125,7 +144,61 @@ export default class Mnemonic {
             return this.toLegacyPrivateKey();
         }
 
-        return this._toPrivateKey(passphrase);
+        const seed = await bip39.toSeed(this.words, passphrase);
+        const digest = await hmac.hash(hmac.HashAlgorithm.Sha512, "ed25519 seed", seed);
```
Nana-EC, line 148: nit: set constant for `"ed25519 seed"`

packages/cryptography/src/Mnemonic.js
```diff
@@ -125,7 +144,61 @@ export default class Mnemonic {
...
+     * Recover an ECDSA private key from this mnemonic phrase, with an
+     * optional passphrase.
+     *
+     * @param {string} [passphrase]
+     * @param {number[]} [path]
+     * @returns {Promise<PrivateKey>}
+     */
+    async toEcdsaPrivateKey(passphrase = "", path = ECDSA_ETH_PATH) {
+        const seed = await bip39.toSeed(this.words, passphrase);
+        const digest = await hmac.hash(hmac.HashAlgorithm.Sha512, "Bitcoin seed", seed);
```
Nana-EC, line 182: nit: make const for "Bitcoin seed"

packages/cryptography/src/Mnemonic.js
```diff
@@ -125,7 +144,61 @@ export default class Mnemonic {
...
+
+        let keyData = digest.subarray(0, 32);
```
Nana-EC, line 184: `const seed` up to `let chainCode` is mostly dup code.
Could pull out into a shared method that take in `passphrase` and `seedText` and returns `keyData` and `chainCode`

packages/cryptography/src/PrivateKey.js
```diff
@@ -268,17 +270,25 @@ export default class PrivateKey extends Key {
...
+        if (this._key instanceof Ed25519PrivateKey) {
+            const { keyData, chainCode } = await slip10.derive(
+                this.toBytesRaw(),
+                this._key._chainCode,
+                index
+            );
 
-        /** @type {new (bytes: Uint8Array, chainCode?: Uint8Array) => Ed25519PrivateKey | EcdsaPrivateKey} */
-        const constructor = /** @type {any} */ (this._key.constructor);
+            return new PrivateKey(new Ed25519PrivateKey(keyData, chainCode));
```
Nana-EC, line 280: Q: In `mneumonic` you calculate the keyPair with `nacl.sign.keyPair.fromSeed(keyData)`.
Here you use keyData directly.
Should they be consistent.

src/Mnemonic.js
```diff
@@ -91,10 +91,45 @@ export default class Mnemonic {
...
     }
 
+    /**
+     * Recover an Ed25519 private key from this mnemonic phrase, with an
+     * optional passphrase.
+     *
+     * @param {string} [passphrase]
+     * @returns {Promise<PrivateKey>}
+     */
+    async toEd25519PrivateKey(passphrase = "") {
```
Nana-EC, line 106: Q: should the path be exposed in these methods.
Not too familiar with the structure. This is most devs will call correct?
Then if they want to customize do they have to depend on the cryptography package specifically?

packages/cryptography/test/unit/Mnemonic.js
```diff
@@ -53,11 +53,27 @@ describe("Mnemonic", function () {
...
+            "f9c31c38697c5bf26645d5521b576ca0369be0dbfec6156ef3c74d0130022dc7";
+
+        let key = await mnemonic.toEcdsaPrivateKey();
+        console.log((await key.derive(-1)).toStringRaw());
+        console.log((await key.derive(0)).toStringRaw());
+        console.log((await key.derive(1)).toStringRaw());
+
+        expect(key.toStringRaw()).to.eql(expectedKey);
+    });
+
```
Nana-EC, line 76: Missing test coverage with customized path.

Nana-EC (changes requested):
Looking good.
Som requests for clarification and missing test coverage

packages/cryptography/src/PrivateKey.js
danielakhterov: This is because we use different libraries for `Ed25519` and `ECDSA` support. `NACL` aka `libsodium` doesn't support `ECDSA`. The library we use for `ECDSA` though, just provides us `keyData` and expects us to pass that around instead of some more concrete type.

src/Mnemonic.js
danielakhterov: The path should be provided here, that's a mistake on my part. I already added the necessary support to `packages/cryptography/src/Mnemonic`, but forgot to update this file as well.

packages/cryptography/test/unit/Mnemonic.js
danielakhterov: Current this test does not pass, so yea we'll need to figure it out. However, the reason it doesn't pass is what's still confusing to me. I provided the path `44/60/0/0` for ECDSA but I'm still getting different keys. I've tried printing out the different common derivations `-1`, `0`, and `1`, but neither of those produce the expected key. I'm not entirely sure what I'm doing wrong though, so I was hoping another pair of eyes would help here.

packages/cryptography/src/PrivateKey.js
Nana-EC: Okay

packages/cryptography/src/Mnemonic.js
```diff
@@ -125,7 +146,77 @@ export default class Mnemonic {
...
+
+    /**
+     * Recover an ECDSA private key from this mnemonic phrase, with an
+     * optional passphrase.
+     *
+     * @param {string} [passphrase]
+     * @param {number[]} [path]
+     * @returns {Promise<PrivateKey>}
+     */
+    async toEcdsaPrivateKey(passphrase = "", path = ECDSA_ETH_PATH) {
```
shemnon, line 181: I think we've settled that all Hedera keys should use the `3030'` path, ED and EC.  So this default should change.
However, we can used the Eth path to validate key derivations to prove the framework is working properly.

Nana-EC (commented):
Looking good to me.
Just need to get the  path update now that we've settled and get tests working and updated.

packages/cryptography/src/Mnemonic.js
danielakhterov: Honestly, not following with the last statement there.

Nana-EC: Similar to my statement.
Basically even though we use the default Hedera one we should have a test that sets the path as the Ethereum path for coverage

packages/cryptography/src/Mnemonic.js
```diff
@@ -125,7 +165,78 @@ export default class Mnemonic {
...
+            ({ keyData, chainCode } = await slip10.derive(
+                keyData,
+                chainCode,
+                index
+            ));
+        }
+
+        const keyPair = nacl.sign.keyPair.fromSeed(keyData);
+
+        if (CACHE.privateKeyConstructor == null) {
```
Nana-EC, line 183: nit: could use another internal method here and below
```suggestion
        return cacheNewPrivateKey(new Ed25519PrivateKey(keyPair, chainCode))
```

Where
```
cacheNewPrivateKey(privateKey) {
        if (CACHE.privateKeyConstructor == null) {
            throw new Error("PrivateKey not found in cache");
        }

        return CACHE.privateKeyConstructor(privateKey);
    }
```

packages/cryptography/src/primitive/ecdsa.js
```diff
@@ -12,20 +11,43 @@ const secp256k1 = new elliptic.ec("secp256k1");
...
+            continue;
+        }
+
+        return {
+            privateKey,
+            publicKey,
+        };
+    }
+
+    throw new Error(`Couldn't generate valid data.`);
```
Nana-EC, line 50: nit:
```suggestion
    throw new Error(`Couldn't generate valid ecdsa key pair.`);
```

Nana-EC (approved):
LG
2 minor nit suggestions

packages/cryptography/src/Mnemonic.js
danielakhterov: Although I do agree with this change, I would rather do this in another PR as I believe I could instead update `CACHE` to be a class with methods per field instead of making a singular method just for `Cache.privateKeyConstructor`
Created tracking issue: https://github.com/hashgraph/hedera-sdk-js/issues/1225

packages/cryptography/src/primitive/ecdsa.js
danielakhterov: No longer applicable after I've migrated back to `elliptic` instead of `tiny-secp256k1`

danielakhterov:
I tried to fix RN CI, but RN hates links/file deps apparently. Will have to revisit after we publish `@hashgraph/cryptography`

packages/cryptography/test/browser.test.js
```diff
@@ -2,7 +2,7 @@ import { test, expect } from "@playwright/test";
 
 test("can execute @hashgraph/cryptography within browser", async function ({ page }) {
     page.on("pageerror", async (message) => {
-        await browser.close();
+        // await browser.close();
```
Nana-EC, line 5: nit: remove if not needed or restore

Nana-EC (dismissed):
Logic looks good to me.
Just need CI to agree

danielakhterov:
@Nana-EC note React Native CI will not work in this PR because it hates links. We're using link in this particular PR because `@hashgraph/cryptography` needs to be updated for `@hashgraph/sdk` to use the new API. We'll have to fix the CI in another PR after we've released the current updates to `@hashgraph/cryptography`

Nana-EC (approved):
LG

PR #1222, opened 2022-08-02
fix: `TokenUpdateTransaction.pauseKey`

### LINKED ISSUE #1221: `pauseKey` does not update with `TokenUpdateTransaction` because its missing from `_makeTransactionData`
### Description

Sent `TokenUpdateTransaction` with `setPauseKey` to a newly created token. The tx was `SUCCESSFUL` but the `pauseKey` kept the old value and did not update as expected.

### Steps to reproduce

image

Looking at the SDK, the `pauseKey` is missing from `_makeTransactionData` [here](https://github.com/hashgraph/hedera-sdk-js/blob/main/src/token/TokenUpdateTransaction.js#L653)

### Additional context

_No response_

### Hedera network

testnet

### Version

v2.17.1

### Operating system

macOS

danielakhterov:
Signed-off-by: Daniel Akhterov <akhterovd@gmail.com>

**Description**:

**Related issue(s)**:

Fixes #1221

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/test/unit/TokenUpdateTransaction.js
+++ b/test/unit/TokenUpdateTransaction.js
@@ -1,0 +1,108 @@
+import {
+    PrivateKey,
+    TokenUpdateTransaction,
+    TransactionId,
+    AccountId,
+    Timestamp,
+} from "../../src/index.js";
+import Long from "long";
+
+describe("TokenUpdateTransaction", function () {
+    it("encodes to correct protobuf", function () {
+        const key1 = PrivateKey.fromString("302e020100300506032b6570042204205fc37fbd55631722b7ab5ec8e31696f6d3f818a15c5258a1529de7d4a1def6e2");
+        const key2 = PrivateKey.fromString("302e020100300506032b657004220420b5e15b70109fe6e11d1d6d06b20d27b494aa05a28a8bc84c627d9be66e179391");
+        const key3 = PrivateKey.fromString("302e020100300506032b6570042204202b5dc9915d5b6829592f4562a3d099e7b1bdd48da347e351da8d31cd41653016");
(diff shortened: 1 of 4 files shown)


test/unit/TokenUpdateTransaction.js
```diff
@@ -0,0 +1,108 @@
...
+            )
+            .setTokenName("name")
+            .setTokenSymbol("symbol")
+            .setTokenMemo("memo")
+            .setTreasuryAccountId(treasuryAccountId)
+            .setAutoRenewAccountId(autoRenewAccountId)
+            .setAdminKey(key1)
+            .setKycKey(key2)
+            .setFreezeKey(key3)
+            .setPauseKey(key4)
```
mgarbs, line 37: Curious why we are using the same keys to update that we used to create? I get why the `adminKey` will be the same but.. If these are all the same keys, then we can't know for sure if their values updated. We could mix up all the keys and check for their updated values in expect to get a more encompassing test that `TokenUpdateTransaction` checks out.

danielakhterov: This is strictly a serialization test; it asserts that if a user sets keys using `set*Key()` then when we generate the protobuf the keys are set accordingly. For instance if a user calls `setPauseKey(key4)` then in the protobuf we should expect `tokenUpate.pauseKey.ed25519` to be set to the same bytes as `key4.publicKey.toBytesRaw()`. This test would catch still mistakes like missing `pauseKey` in the `_makeTransactionData()`

mgarbs: Okay got it that makes sense

PR #1237, opened 2022-08-15
chore: make react-native peer dependencies optional

mehcode:
Signed-off-by: Ryan Leckey <ryan@launchbadge.com>

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,5 +110,10 @@
     },
     "peerDependencies": {
         "expo": "^45.0.3"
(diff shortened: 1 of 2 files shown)


danielakhterov:
Btw, we did have this change at one point in the past and it has to be undone. I don't remember what the exact error was nor did I record it, but I do remember _some_ issue happening. Anyways, will merge this _after_ release as to not potentially explode this one.

PR #1240, opened 2022-08-16
fix typo in function name

Shungy:
Fix function name according to documentation.

[Docs](https://docs.hedera.com/guides/docs/sdks/smart-contracts/create-a-smart-contract) state ContractCreateFlow inherits names from ContractCreateTransaction:

> Note: Please refer to ContractCreateTransaction()  for a complete list of applicable methods.

Function's name in ContractCreateTransaction is `setMaxAutomaticTokenAssociations`. Therefore the same name must be used in ContractCreateFlow

### DIFF
--- a/src/contract/ContractCreateFlow.js
+++ b/src/contract/ContractCreateFlow.js
@@ -226,7 +226,7 @@ export default class ContractCreateFlow {
      * @param {number} maxAutomaticTokenAssociation
      * @returns {this}
      */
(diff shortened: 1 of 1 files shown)


danielakhterov:
Unfortunately we cannot simply fix this typo. We need to create a new method, and deprecate the old one.

danielakhterov:
Merging this, will add the deprecation on `main`

PR #1361, opened 2023-01-04
Return dependabot dependency checker

### LINKED ISSUE #1360: Fix dependency bump automation
### Problem

hedera-sdk-js used to use Dependabot but in https://github.com/hashgraph/hedera-sdk-js/commit/f065cb1568191a546e8b4e8a52b7867149badc53 it was switched to Renovate. In the recent commit history I don't see any sort of automated dependency PRs by bots so Renovate must not be working.

### Solution

Either fix renovate or revert back to Dependabot which actually worked.

### Alternatives

_No response_

petreze:
Signed-off-by: Petar Tonev <petar.tonev@limechain.tech>

**Description**:
Remove the configuration for `renovate` and return the configuration for `dependabot` dependency checker since it was previously working well. `Dependabot` is also used in all other SDKs and hedera-services, for this reason it would be better to keep the same convention here as well

**Related issue(s)**:
Fixes #1360

**Notes for reviewer**:

**Checklist**


.github/dependabot.yml
```diff
@@ -0,0 +1,28 @@
...
+    interval: daily
+    time: "15:15"
+    timezone: "Europe/Sofia"
+  open-pull-requests-limit: 10
+- package-ecosystem: npm
+  directory: "packages/cryptography"
+  schedule:
+    interval: weekly
+  open-pull-requests-limit: 10
+  directory: "packages/proto"
```
steven-sheehy, line 15: This seems to be missing a `- package-ecosystem: npm`

.github/dependabot.yml
```diff
@@ -0,0 +1,28 @@
...
+updates:
+- package-ecosystem: npm
+  directory: "/"
+  schedule:
+    interval: daily
+    time: "15:15"
+    timezone: "Europe/Sofia"
+  open-pull-requests-limit: 10
+- package-ecosystem: npm
+  directory: "packages/cryptography"
```
steven-sheehy, line 11: You probably want to do an npm up or manually bump up versions and commit that to avoid the flood of initial PRs.

.github/dependabot.yml
petreze: Good point, added :+1:

.github/dependabot.yml
steven-sheehy: Why was this resolved without comment and no code changed to address? It's fine if you prefer to not manually update package.json in this PR and let Dependabot handle it with dozens of PRs as soon as this is merged, but would appreciate a comment if that's your approach.

steven-sheehy (approved):

ochikov (approved):

.github/dependabot.yml
ochikov: That's because it has been disabled for a while. We are going to combine them into one and try to resolve on one PR. We are going to deal with this into the next days.

PR #1395, opened 2023-01-13
Fix web proxy

### LINKED ISSUE #1363: GRPC-WEB Proxies needs to be monitored and put as healthy/unhealthy
### Description

If we get some kind of an HTTP ERROR CODE from the grpc-web-proxy, currently the SDK is design to watch only for headers labeled: `grpc-status` and `grpc-message` and thus we cannot catch the empty headers with the empty response but with HTTP ERROR CODE errors, so we cannot mark this web-proxy as unhealthy.

### Steps to reproduce

1. Create react-app
2. Connect the client to MHW web-proxy
3. Try to execute transaction
4. Get 503

### Additional context

Here are the only two headers that we are currently watching:
https://github.com/hashgraph/hedera-sdk-js/blob/develop/src/channel/WebChannel.js#L71

### Hedera network

mainnet, testnet, previewnet

### Version

v2.19.1

### Operating system

macOS

petreze:
**Description**:
This PR fixes the issue where the SDK does not check for any HTTP errors when trying to connect to some of the node proxies. Also there is a fix for a corner case if the random mechanism chooses the same node twice in a row and puts it inside the healthy nodes list

**Related issue(s)**:

Fixes #1363 

**Checklist**

### DIFF
--- a/src/Executable.js
+++ b/src/Executable.js
@@ -23,6 +23,7 @@ import GrpcStatus from "./grpc/GrpcStatus.js";
 import List from "./transaction/List.js";
 import Logger from "js-logger";
 import * as hex from "./encoding/hex.js";
+import HttpError from "./http/HttpError.js";
 
 /**
  * @typedef {import("./account/AccountId.js").default} AccountId
@@ -133,8 +134,6 @@ export default class Executable {
      * @returns {?AccountId[]}
      */
     get nodeAccountIds() {
-        console.log(`in nodeAccountIds: ${this._nodeAccountIds.locked}`)
-        console.log(`in nodeAccountIds: ${this._nodeAccountIds.list}`)
         if (this._nodeAccountIds.isEmpty) {
             return null;
         } else {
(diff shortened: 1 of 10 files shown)


src/Executable.js
```diff
@@ -440,16 +441,22 @@ export default class Executable {
      * Unlike `shouldRetry` this method does in fact still return a boolean
      *
      * @protected
-     * @param {GrpcServiceError} error
+     * @param {Error} error
```
ochikov, line 444: Why we need to change here the type of the error? Could we specify it with GrpcServiceError | Error?

src/Executable.js
```diff
@@ -440,16 +441,22 @@ export default class Executable {
...
-                RST_STREAM.test(error.message))
-        );
+        if (error instanceof GrpcServiceError) {
+            return (
+                error.status._code === GrpcStatus.Unavailable._code ||
+                error.status._code === GrpcStatus.ResourceExhausted._code ||
+                (error.status._code === GrpcStatus.Internal._code &&
+                    RST_STREAM.test(error.message))
+            );
+        } else {
```
ochikov, line 455: Are we sure that here into the else statement, the error is http error?

src/client/NativeClient.js
```diff
@@ -50,6 +50,7 @@ export const Network = {
 
     MAINNET: {
         "https://grpc-web.myhbarwallet.com:443": new AccountId(3),
+        "https://node01-00-grpc.swirlds.com:443": new AccountId(4),
```
ochikov, line 53: We need to add the following GRPC proxies to Native and Web client as well:
https://node02.swirldslabs.com/
https://node03.swirldslabs.com/
https://node04.swirldslabs.com/
https://node05.swirldslabs.com/
https://node06.swirldslabs.com/
https://node07.swirldslabs.com/
https://node08.swirldslabs.com/
https://node09.swirldslabs.com/
https://node10.swirldslabs.com/
https://node11.swirldslabs.com/
https://node12.swirldslabs.com/
https://node13.swirldslabs.com/
https://node14.swirldslabs.com/
https://node16.swirldslabs.com/
https://node17.swirldslabs.com/
https://node18.swirldslabs.com/
https://node19.swirldslabs.com/
https://node20.swirldslabs.com/
https://node21.swirldslabs.com/
https://node22.swirldslabs.com/
https://node23.swirldslabs.com/
https://node24.swirldslabs.com/
https://node25.swirldslabs.com/
https://node26.swirldslabs.com/

The nodeAccountIds are here: https://status.hedera.com/

ochikov (changes requested):

src/client/NativeClient.js
ochikov: Please note that node01 and node15 are MISSING. This is on purpose!

src/Executable.js
ochikov: I get it. Now we are moving one level up, the GrpcServiceError is an instance of Error and now we've got also HttpError. We are passing the parent class.

src/Executable.js
ochikov: Yes, you are right. I've tested it.

src/client/NativeClient.js
petreze: Added

PR #2020, opened 2023-11-09
LocalProvider - close method

svetoslav-nikol0v:
**Description**:

- Adding a close method in LocalProvider class that closes the client

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

The idea is to close client when we have an instance of the Wallet which is using LocalProvider.

**Checklist**

### DIFF
--- a/src/LocalProvider.js
+++ b/src/LocalProvider.js
@@ -163,4 +163,14 @@ export default class LocalProvider {
     call(request) {
         return request.execute(this._client);
     }
+
+    /**
+     * @returns {void}
+     */
+    close() {
+        this._client._network.close();
+        this._client._mirrorNetwork.close();
+        this._client._isShutdown = true;
+        clearTimeout(this._client._timer);
+    }
 }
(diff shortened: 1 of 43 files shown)


src/LocalProvider.js
```diff
@@ -163,4 +163,14 @@ export default class LocalProvider {
...
     }
+
+    /**
+     * @returns {void}
+     */
+    close() {
+        this._client._network.close();
+        this._client._mirrorNetwork.close();
+        this._client._isShutdown = true;
+        clearTimeout(this._client._timer);
```
rokn, line 174: Can we just call `this._client.close()` ?

svetoslav-nikol0v: Yes, we can. For some reason it didn't work first time but now it's ok.

rokn (approved):

rokn (approved):

ochikov (approved):

PR #2195, opened 2024-03-13
fix: do not instantiate logger inside a method

### LINKED ISSUE #2194: A new logger is instantiated every time a method is invoked
### Description

A new logger is instantiated every time `ManagedNetwork:_getNumberOfMostHealthyNodes` is invoked: https://github.com/hashgraph/hedera-sdk-js/blob/develop/src/client/ManagedNetwork.js#L219

Therefore, I have experienced that my program doesn't terminate as logger keeps holding resources (PIPE). You can monitor this on a macOS by running: `lsof -c node | grep PIPE | wc -l` before running the program and then again when the program is not terminating after its execution.

As shown below, you can see although NodeJS was holding 46 sockets on my machine, the number of PIPE resources increased rapidly to 1394. The program then terminates when it comes down to 22 (in my case this was the total number of resources held by other NodeJS programs)

Screenshot 2024-03-13 at 4 37 37 PM

Here are some commands that I used to monitor the total number of sockets and PIPE resources held by NodeJS:
```
# pipe-total.sh                                                                                                              
#!/bin/bash
lsof -c node | grep PIPE | wc -l

# total-socket.sh                                                                                                              
#!/bin/bash
lsof -i -P -n | wc -l
```

### Steps to reproduce

I instantiated a node client for a managed network (locally deployed using [solo](https://github.com/hashgraph/solo?tab=readme-ov-file#example---1-deploy-a-standalone-test-network-version-0425)) like this: `Client.fromConfig({ network: nodes })` and then used that client to invoke ~700 account queries

### Additional context

After deleting that line locally the program terminates immediately and total number of PIPE resources were fixed: https://github.com/hashgraph/hedera-sdk-js/blob/develop/src/client/ManagedNetwork.js#L219

Screenshot 2024-03-13 at 9 18 17 PM
 

Here is the ticket that I was trying to resolve and found this issue: https://github.com/hashgraph/solo/issues/153

### Hedera network

other

### Version

v2.42.0

### Operating system

macOS

leninmehedy:
**Description**:
This PR removes the problematic instantiation of a logger inside a method that is invoked repeatedly.

**Related issue(s)**:

Fixes #2194 

**Notes for reviewer**:

I couldn't run the integration tests as there were no instruction in the README. I saw errors when I ran `task test` locally: `Error: Failed to construct client for IntegrationTestEnv`

However, please let me know if anything else I should test/modify.

**Checklist**

### DIFF
--- a/src/client/ManagedNetwork.js
+++ b/src/client/ManagedNetwork.js
@@ -20,7 +20,6 @@
 
 import LedgerId from "../LedgerId.js";
 import * as util from "../util.js";
-import Logger from "../logger/Logger.js"; // eslint-disable-line
 
 /**
  * @typedef {import("../channel/Channel.js").default} Channel
@@ -87,14 +86,6 @@ export default class ManagedNetwork {
         this._nodeMaxReadmitPeriod = this._maxBackoff;
 
         this._earliestReadmitTime = Date.now() + this._nodeMinReadmitPeriod;
-
-        /**
-         * Logger
-         *
-         * @external
-         * @type {Logger | null}
-         */
-        this._logger = null;
     }
 
     /**
@@ -227,11 +218,6 @@ export default class ManagedNetwork {
 
             // Check if the node exists
             if (!selectedNode) {
-                if (this._logger) {
-                    this._logger.debug(
-                        `Selected an undefined node at index ${nodeIndex}`,
-                    );
-                }
                 break; // Break out of the loop if undefined node is selected
             }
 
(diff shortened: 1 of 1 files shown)


svetoslav-nikol0v:
Hey [leninmehedy](https://github.com/leninmehedy),
This logger was only added to track the behavior after the fix we added. We agree to remove it at all. Will add a commit for this.

agadzhalov (approved):
LGTM

PR #2213, opened 2024-03-19
fix: integer overflow isuue for defaultMaxQueryPayment field

### LINKED ISSUE #2203: setDefaultMaxQueryPayment integer overflow
### Description

Currently, setDefaultMaxQueryPayment has a integer overflow problem. If setDefaultMaxQueryPayment is set to 25 hbar or more, the SDK will throw `giving Error: defaultMaxQueryPayment must be non-negative`. 

This bug is caused by this line of code. https://github.com/hashgraph/hedera-sdk-js/blob/76820600d6d7f0ea7a200b599be1c8a7b4f24d1c/src/client/Client.js#L452

### Steps to reproduce

Try to set `setDefaultMaxQueryPayment` like following and run the script

```
client.setDefaultMaxQueryPayment(new Hbar(25)); 
```

### Additional context

_No response_

### Hedera network

mainnet, testnet, previewnet, other

### Version

v2.40

### Operating system

None

svetoslav-nikol0v:
**Description**:

Handles setting larger integer numbers when are set to `defaultMaxQueryPayment` field.

Fixes #2203 

**Checklist**

### DIFF
--- a/test/integration/ClientIntegrationTest.js
+++ b/test/integration/ClientIntegrationTest.js
@@ -172,6 +172,24 @@ describe("ClientIntegration", function () {
         expect(clientTestnet.isTransportSecurity()).to.be.an("boolean");
     });
 
+    it("should return the following error message `defaultMaxQueryPayment must be non-negative` when the user tries to set a negative value to the defaultMaxQueryPayment field", async function () {
+        this.timeout(120000);
+        try {
+            env.client.setDefaultMaxQueryPayment(new Hbar(1).negated());
+        } catch (error) {
+            expect(error.message).to.be.equal(
+                "defaultMaxQueryPayment must be non-negative",
+            );
+        }
+    });
+
+    it("should set defaultMaxQueryPayment field", async function () {
+        this.timeout(120000);
+        const value = new Hbar(100);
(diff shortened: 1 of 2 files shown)


test/integration/ClientIntegrationTest.js
```diff
@@ -172,6 +172,24 @@ describe("ClientIntegration", function () {
...
+        } catch (error) {
+            expect(error.message).to.be.equal(
+                "defaultMaxQueryPayment must be non-negative",
+            );
+        }
+    });
+
+    it("should set defaultMaxQueryPayment field", async function () {
+        this.timeout(120000);
+        const value = new Hbar(100);
```
agadzhalov, line 188: for the purpose of the test to demonstrate that can handle really large numbers I would suggest
```suggestion
        const value = new Hbar(9223372036854775807);
```

agadzhalov (approved):
LGTM

test/integration/ClientIntegrationTest.js
svetoslav-nikol0v: We need pass this number in that way `Long.fromString('9223372036854775807')` because is bigger than MAX_SAFE_INTEGER number.

PR #2214, opened 2024-03-19
update: add a new getter to the TransferTransaction class

### LINKED ISSUE #2190: Transfer Transaction: isApproved flag is missing on current transfers getters
### Problem

When you create an approved transfer it is internally saved as a Transfer. When you try to use the getter for all transfers (e.g. transaction.hbarTransfers), the value is a TransferMap, which does not contain information on whether the transfer is approved.

Here is an example getter of transfers:
`
get hbarTransfers() {
        const map = new HbarTransferMap();

        for (const transfer of this._hbarTransfers) {
            map._set(transfer.accountId, transfer.amount);
        }

        return map;
    }
`

_hbarTransfers is of type Transfer[] and Transfer contains a boolean property isApproved.
This boolean property is missing when the getter is used

### Solution

A possible solution is to rework the hbarTransfers, tokenTransfers and nftTransfers or create new getters that returns a shallow representation of 'Transfer[]'

### Alternatives

_No response_

svetoslav-nikol0v:
**Description**:

A new getter was added to the TransferTransaction class.

Fixes #2190 

**Checklist**

### DIFF
--- a/src/account/TransferTransaction.js
+++ b/src/account/TransferTransaction.js
@@ -369,6 +369,13 @@ export default class TransferTransaction extends Transaction {
     }
 
     /**
+     * @returns {Transfer[]}
+     */
+    get hbarTransfersList() {
+        return this._hbarTransfers;
+    }
+
+    /**
      * @internal
      * @param {AccountId | string} accountId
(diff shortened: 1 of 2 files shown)


agadzhalov (approved):
LGTM

PR #2298, opened 2024-05-23
fix: enhanced Logger to accept a log file location

### LINKED ISSUE #2297: Enable logger constructor to accept a log file destination
### Problem

For [Solo](https://github.com/hashgraph/solo) which runs as a NodeJS CLI we want to be able to write the @hashgraph/sdk logs to a file for debug purposes.

### Solution

Update the constructor of the Logger to accept a log file location as a parameter.

### Alternatives

pass in a DestinationStream that can be created by the consumer.

jeromy-cannon:
**Description**:

- enhanced Logger to accept a log file location
- fixed some jsdoc mispellings
- added `src/grpc/GrpcServiceError.js` `toString()` method, updated message, and add caused by stack trace
- increased timeouts on some tests to prevent random failures in the pipeline when under a load
- increased Logger test coverage
- added `sinon` library for spy capabilities
- added cgroup configuration and calls to build workflow to fix issue where GitHub runner loses communication with GitHub

**Related issue(s)**:

Fixes #2297 

**Notes for reviewer**:

**Checklist**


src/grpc/GrpcServiceError.js
```diff
@@ -55,10 +55,19 @@ export default class GrpcServiceError extends Error {
         if (obj.code != null && obj.details != null) {
             const status = GrpcStatus._fromValue(obj.code);
             const err = new GrpcServiceError(status);
-            err.message = obj.details;
+            err.stack +=
+                "\nCaused by: " + (obj.stack ? obj.stack.toString() : "");
```
agadzhalov, line 59: For keeping consistency I would suggest to use template literals here 
```suggestion
            err.stack += `\nCaused by: ${obj.stack ? obj.stack.toString() : ""}`;
```

test/unit/LoggerTest.js
```diff
@@ -68,4 +70,23 @@ describe("Logger", function () {
         expect(levels).to.include("error");
         expect(levels).to.include("fatal");
     });
+
```
agadzhalov, line 73: I would suggest adding one more unit test with no logFile

agadzhalov (changes requested):

src/grpc/GrpcServiceError.js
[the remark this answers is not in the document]
svetoslav-nikol0v: I agree, it should be  

```suggestion
           err.stack += `${obj.stack ? `"\nCaused by:" ${obj.stack.toString()}` : ""}`
```

src/grpc/GrpcStatus.js
```diff
@@ -89,43 +89,43 @@ export default class GrpcStatus {
     toString() {
         switch (this) {
             case GrpcStatus.Ok:
-                return "OK";
+                return `OK (${this._code})`;
```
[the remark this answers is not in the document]
svetoslav-nikol0v: Looking the other classes like `Status` for example and to preserve code consistency, `toString()` should return only name of the gRPC error. If it's necessary to build the this format `Status: OK, Code: 0` it could be achieved like this `Status: ${error.status.toString()}, Code: ${error.status.valueOf()}`

test/unit/LoggerTest.js
jeromy-cannon: complete

src/grpc/GrpcStatus.js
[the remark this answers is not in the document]
jeromy-cannon: updated

.github/workflows/build.yml
```diff
@@ -65,6 +68,47 @@ jobs:
...
+              cgset -r memory.max=${AGENT_MEM_LIMIT} ${AGENT_GROUP_NAME}
+              cgset -r memory.swap.max=${E2E_MEM_LIMIT} ${E2E_GROUP_NAME}
+              cgset -r memory.swap.max=${AGENT_MEM_LIMIT} ${AGENT_GROUP_NAME}
+            echo "::endgroup::"
+
+            echo "::group::Move Runner Processes to Control Groups"
+              sudo cgclassify --sticky -g cpu,memory:${AGENT_GROUP_NAME} $(pgrep 'Runner.Listener' | tr '\n' ' ')
+              sudo cgclassify -g cpu,memory:${AGENT_GROUP_NAME} $(pgrep 'Runner.Worker' | tr '\n' ' ')
+            echo "::endgroup::"
+
```
agadzhalov, line 111: Can you give more details about this snippet what it does and why is needed? Also is this related to the logger functionality?

agadzhalov: Thanks for the explanation, it's a good idea and good job 🙌. I don't mind keeping it here as we are already aware of this

rbarkerSL (approved):
LGTM

agadzhalov (approved):
LGTM, thank you @jeromy-cannon

PR #2321, opened 2024-06-02
chore: pin used pnpm version in package.json

mehcode:
For documentation purposes at least, the expected `pnpm` version should be declared in `package.json`. See https://nodejs.org/api/packages.html#packagemanager

I would recommend updating to pnpm v9 when time allows. I would have done it here, but it involves an automatic reformat of the lockfile and has a high chance of conflicting any current work.


svetoslav-nikol0v (approved):

svetoslav-nikol0v:
This version of the package manager breaks the CI. The version used in the CI is `8.15.4`.

PR #2322, opened 2024-06-02
chore: update long to v5.x

mehcode:
Long v5 is very old (2+ years) at this point. There are no breaking changes from v4. The primary change is using an ESM module style compared to CommonJS. 

If you'll look at the diff of this PR, the Hedera SDK already depended on Long v5 and generated code for Long v5 via ProtobufJS but implicitly casted the Long v5 type to Long v4 when presenting it the consumers of the SDK. This is effectively a bug fix.


svetoslav-nikol0v:
Hello [mehcode](https://github.com/mehcode),
What do you mean by `This is effectively a bug fix`? Is there an actual bug caused by `Long` v4?

mehcode:
The Hashgraph SDK _**does not actually use**_ Long v4. The primary usage comes from generated code from ProtobufJS. The ProtobufJS library uses Long v5. This only works right now because there are no breaking changes from v4 to v5. It's still technically a bug because the Hashgraph SDK is misrepresenting an object created with Long v5 (inside of code generated by ProtobufJS) as Long v4.

ivaylonikolov7:
resolved via #2692

PR #2388, opened 2024-07-04
proj: break down task into build:prep and build:actual

bguiz:
**Description**:

This PR modifies `Taskfile.yml` in order to support
* a faster feedback cycle when modifying source only

**Related issue(s)**:

Nil

**Notes for reviewer**:

Nil

**Checklist**

- [ ] Tested (unit, integration, etc.) - N/A


ivaylonikolov7:
This PR stopped working after we migrated to hiero-ledger. I suspect the runners are causing the issue. I'll recreate the PR. I've already discussed this with @bguiz via direct messages and he doesn't mind it.

PR #2401, opened 2024-07-16
update: handle PLATFORM_NOT_ACTIVE error gracefully

### LINKED ISSUE #2353: Handle platform_not_active error gracefully
### Problem

The SDK offers a convenience feature where when a transaction is signed by the user it will pre-sign transactions for 1/3 of the nodes in the network. This is to provide redundancy in cases where a transaction could fail and can pick up the next signed transaction to a different node. A transaction cannot be sent to all nodes as the signed transaction requires you to specify a single node before user the signs the transaction. If there was not such a feature you would imagine that the user would have to be pinged to resign another transaction each time it tries to send the transaction to a different node. 

In some cases, this feature could provide an unwanted outcome in that the first transaction it tries that node could return an error to the user while it is trying to submit the following transaction to a different node. The user would then observe that their transaction failed (true for the first transaction), but was actually successful (when it got submitted to the next node).

This issue is to investigate the above scenario and see if there are ways to improve this experience so that the user does not get mislead. Also to give guidance to developers on how their apps should account for this scenario.

This example provided that kicked off this discussion is the following:

Receiving a PLATFORM_NOT_ACTIVE error node a node. At the time this was received we believe this from a node that was marked as degraded. The the following the transaction was successful on the explorer 

https://hashscan.io/mainnet/transaction/1718307329.984046337

image (17)

### Solution

Research potential ways to improve and provide guidance to developers for how to manage these cases. 

### Alternatives

_No response_

svetoslav-nikol0v:
**Description**:

When the SDK gets `PLATFORM_NOT_ACTIVE` error, will retry to execute the transaction until reaching the max attempts limit.

Fixes #2353 

**Checklist**

### DIFF
--- a/src/query/Query.js
+++ b/src/query/Query.js
@@ -492,6 +492,7 @@ export default class Query extends Executable {
             case Status.Busy:
             case Status.Unknown:
             case Status.PlatformTransactionNotCreated:
+            case Status.PlatformNotActive:
                 return [status, ExecutionState.Retry];
             case Status.Ok:
                 return [status, ExecutionState.Finished];
--- a/src/transaction/TransactionReceiptQuery.js
+++ b/src/transaction/TransactionReceiptQuery.js
@@ -221,6 +221,7 @@ export default class TransactionReceiptQuery extends Query {
             case Status.Busy:
             case Status.Unknown:
             case Status.ReceiptNotFound:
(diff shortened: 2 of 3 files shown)


agadzhalov (approved):
LGTM

PR #2402, opened 2024-07-17
feat: create node transaction

### LINKED ISSUE #2335: Add HIP-869: `NodeCreateTransaction`

svetoslav-nikol0v:
**Description**:

[HIP-869]

Fixes #2335 

**Notes for reviewer**:
Still in progress...

**Checklist**


src/node/NodeCreateTransaction.js
```diff
@@ -0,0 +1,441 @@
...
+     * @returns {NodeCreateTransaction}
+     */
+    setGossipEndpoints(gossipEndpoints) {
+        if (gossipEndpoints.length == 0) {
+            throw new Error("GossipEndpoints list must not be empty.");
+        }
+
+        if (gossipEndpoints.length > GOSSIP_ENDPOINTS_MAX_LENGTH) {
+            throw new Error(
+                `GossipEndpoints list must not contain more than ${GOSSIP_ENDPOINTS_MAX_LENGTH} entries.`,
```
agadzhalov, line 248: i believe this comma is not needed
```suggestion
                `GossipEndpoints list must not contain more than ${GOSSIP_ENDPOINTS_MAX_LENGTH} entries.`
```

src/node/NodeCreateTransaction.js
```diff
@@ -0,0 +1,441 @@
...
+     * @returns {NodeCreateTransaction}
+     */
+    setServiceEndpoints(serviceEndpoints) {
+        if (serviceEndpoints.length == 0) {
+            throw new Error("ServiceEndpoints list must not be empty.");
+        }
+
+        if (serviceEndpoints.length > SERVICE_ENDPOINTS_MAX_LENGTH) {
+            throw new Error(
+                `ServiceEndpoints list must not contain more than ${SERVICE_ENDPOINTS_MAX_LENGTH} entries.`,
```
agadzhalov, line 287:
```suggestion
                `ServiceEndpoints list must not contain more than ${SERVICE_ENDPOINTS_MAX_LENGTH} entries.`
```

src/node/NodeCreateTransaction.js
```diff
@@ -0,0 +1,441 @@
...
+
+        return this;
+    }
+
+    /**
+     * @description Get list of service endpoints for gRPC calls.
+     * @returns {?Array<ServiceEndpoint>}
+     */
+    get getServiceEndpoints() {
+        return this._gossipEndpoints;
```
agadzhalov, line 301: I think you meant `_serviceEndpoints` instead of `_gossipEndpoints` as the method is `getServiceEndpoints()` and the method is for gRPC calls not for gossip
```suggestion
        return this._serviceEndpoints;
```

src/node/ServiceEndpoint.js
```diff
@@ -0,0 +1,163 @@
...
+    /**
+     * @param {Uint8Array} ipAddressV4
+     * @description Set 4-byte IPv4 address of the endpoint.
+     * @returns {ServiceEndpoint}
+     *
+     */
+    setIpAddressV4(ipAddressV4) {
+        if (this._domainName != null) {
+            throw new Error(
+                "Cannot set IP address when domain name is already set.",
```
agadzhalov, line 71:
```suggestion
                "Cannot set IP address when domain name is already set."
```

src/node/ServiceEndpoint.js
```diff
@@ -0,0 +1,163 @@
...
+                serviceEndpoint.ipAddressV4 != null
+                    ? serviceEndpoint.ipAddressV4
+                    : undefined,
+            port:
+                serviceEndpoint.port != null ? serviceEndpoint.port : undefined,
+            domainName:
+                serviceEndpoint.domainName != null
+                    ? serviceEndpoint.domainName
+                    : undefined,
+        });
```
agadzhalov, line 149: I think some type of error handling can be implemented here. For example we can check whether ipAddressV4 and port are different than null or undefined AND domainName is null and vice versa. And after all the checks are passed to `return new ServiceEndpoint`. I suggest that because imagine a case where the three of `ipAddressV4`, `port` and `domainName` or null or undefined. Is this scenario even possible? @svetoslav-nikol0v

src/node/NodeCreateTransaction.js
svetoslav-nikol0v: It's coming from prettier rules.

src/node/NodeCreateTransaction.js
svetoslav-nikol0v: It's coming from prettier rules.

src/node/ServiceEndpoint.js
svetoslav-nikol0v: It's coming from prettier rules.

src/node/ServiceEndpoint.js
svetoslav-nikol0v: So you mean we need to add a check if `ipAddressV4` or `domainName` is set and another check if the `port` is also set, right?

Issue #2419, opened 2024-07-26
Mnemonics -> ECDSA key -> ECDSA key doesn't yield same public key


Volind:
### Description

I'm recovering a private key from Mnemonics and printing the public key as a string.
I then use the private key string above to instantiate a PrivateKey and print its public key as a string.
The two public keys differ, yet printing both private keys as string yield the same result, see example below

### Steps to reproduce

```
import {Mnemonic, PrivateKey} from "@hashgraph/sdk";

const seedStr = "marriage bounce fiscal express wink wire trick allow faith mandate base bone";
const mnemonic = await Mnemonic.fromString(seedStr);
const privateKey = await mnemonic.toStandardECDSAsecp256k1PrivateKey();
const publicKeyStr = privateKey.publicKey.toStringRaw();
console.log("Public from PrivateKey directly: ", publicKeyStr);

const privateKeyStr = privateKey.toStringDer();
console.log("PrivateKey directly: ", privateKeyStr);

const privateKeyFromStr = await PrivateKey.fromStringDer(privateKeyStr);
const privateKeyFromStrStr = privateKeyFromStr.toStringDer();
console.log("PrivateKey after conversion from string: ", privateKeyFromStrStr);


const brokenPublicKeyStr = privateKeyFromStr.publicKey.toStringRaw();
console.log("Public from PrivateKey string: ", brokenPublicKeyStr);

console.log("Private keys string matching: ", privateKeyStr === privateKeyFromStrStr);
console.log("Public Keys matching: ", publicKeyStr === brokenPublicKeyStr);
```

outputs:

```
Public from PrivateKey directly:  036ebe6dffaa5290f002a790a7f1db0e0984332ed0c3b2906f2134db05294871ab
PrivateKey directly:  3030020100300706052b8104000a04220420a5b9dedc7c7767c4a88a0ac3cc6b9171948bada579226bd550d3342db11dc800
PrivateKey after conversion from string:  3030020100300706052b8104000a04220420a5b9dedc7c7767c4a88a0ac3cc6b9171948bada579226bd550d3342db11dc800
Public from PrivateKey string:  03009472c2a77901577f8a60b10e13cd77748a49fb4aa03e711e615c55aaf8f3d2
Private keys string matching:  true
Public Keys matching:  false
```

[...]

PR #2451, opened 2024-08-07
fix: ECDSA and ED25519 public key mismatch when you get it from mnemonic

ivaylonikolov7:
**Description**: 

This PR modifies the crypto package and in particular the ECDSA and ED25519 private key classes. A user found out that a mnemonic and a key that share the same private key mismatch their public key. In order to fix it, this PR modifies the ECDSA and ED25519 so they don't add trailing zeroes. This addition only happens only when the private key is shorter than 32 bytes. Example: 3030020100300706052b8104000a04220420a5b9dedc7c7767c4a88a0ac3cc6b9171948bada579226bd550d3342db11dc800

The last 2 trailing zeroes should be added after the DER prefix  bytes (3030020100300706052b8104000a04220420) 
e.g 3030020100300706052b8104000a0422042000a5b9dedc7c7767c4a88a0ac3cc6b9171948bada579226bd550d3342db11dc8

**Fixes**

 #2419

**Notes for reviewer**:
Not sure if 100 tests are enough of a stress test but otherwise the test gets slow. For example 10000 iterations take around 15 seconds per test case. Considering there is 4 of these per file, thats gonna take 2 minute in total which in my opiion is a lot. 

**Checklist**

### DIFF
--- a/test/unit/Ed25519PrivateKey.js
+++ b/test/unit/Ed25519PrivateKey.js
@@ -1,6 +1,6 @@
 import { expect } from "chai";
 
-import { PrivateKey } from "../../src/index.js";
+import { Mnemonic, PrivateKey } from "../../src/index.js";
 import * as hex from "../../src/encoding/hex.js";
 
 const RAW_KEY =
@@ -14,6 +14,8 @@ const DER_PRIVATE_KEY_BYTES = new Uint8Array([
 ]);
 const DER_PUBLIC_KEY =
     "302a300506032b65700321004a6892f034d2d1c9b1a76acca8e34884055172f4210a0c02e3c7d55084f224d1";
+
+const STRESS_TEST_ITERATION_COUNT = 100;
 
 describe("Ed25519PrivateKey", function () {
     it("generate should return  object", function () {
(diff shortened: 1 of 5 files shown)


test/unit/Ed25519PrivateKey.js
```diff
@@ -15,6 +15,8 @@ const DER_PRIVATE_KEY_BYTES = new Uint8Array([
 const DER_PUBLIC_KEY =
     "302a300506032b65700321004a6892f034d2d1c9b1a76acca8e34884055172f4210a0c02e3c7d55084f224d1";
 
+const STRESS_TEST_ITERATION_COUNT = 100;
```
gsstoykov, line 18: Try a bigger value here. There are no heavy calls. In theory you could even try more than 10^7.

ivaylonikolov7: It takes a lot of time :( 10 000 tests take 50 seconds to run. Therefore 1000 test take like 5 seconds. 

Now that I think about it... considering the CI takes 30 minutes to run, then 1 minute per test case isn't that much. 

But they are 8 tests in total, so basically 8 minutes addition.

svetoslav-nikol0v: 100 is too small and doesn't guarantee that the method works correctly.
Increase tests to 1000 and timeout to 200000.

svetoslav-nikol0v (approved):

agadzhalov (approved):
LGTM

Issue #2433, opened 2024-08-01
`TCK-JS`: Implement JSON-RPC methods for `AccountCreateTransaction`


Petyo-Lukanov:
### Problem

We should implement JSON-RPC methods (endpoints) for AccountCreateTransaction for the JС TCK server

### Solution

Add the following methods:
- Account: createAccount

PR #2476, opened 2024-08-22
feat: Added TCK server json-rpc methods for account create transaction

ivaylogarnev-limechain:
**Description**:
This PR implements the TCK JSON-RPC server using TypeScript. Developed and integrated the AccountCreate method, and successfully validated its functionality through comprehensive tests within the TCK project.

**Related issue(s)**:
#2433 

**Checklist**

### DIFF
--- a/tck/mapping.ts
+++ b/tck/mapping.ts
@@ -1,0 +1,33 @@
+import { sdk } from "./sdk_data";
+import { Input } from "./models/sdk";
+
+/**
+ * Very primitive catch-all mapping prototype
+ * @returns {Promise<*>}
+ * @param {Input} input
+ */
+export default async function mapMethods({
+    callClass,
+    methods,
+}: Input): Promise<string> {
+    const cl: any = (await import("@hashgraph/sdk"))[callClass];
+
+    let currentObject: any = new cl();
+    for (let { name, param } of methods) {
+        console.log(`.${name}(${param})`);
+        if (param === "client") {
+            param = sdk.getClient();
+        }
+
+        if (typeof currentObject[name] === "function") {
+            currentObject = await currentObject[name](param);
+        } else if (typeof cl[name] === "function") {
+            currentObject = await cl[name](param);
+        } else if (typeof currentObject[name] === "object") {
+            currentObject = await currentObject[name];
+        } else {
+            throw Error(`${callClass}.${name}() isn't a function`);
+        }
+    }
+    return currentObject;
+}
--- a/tck/methods/account.ts
+++ b/tck/methods/account.ts
@@ -1,0 +1,102 @@
+import {
+    AccountCreateTransaction,
+    Hbar,
+    AccountId,
+    TransactionResponse,
+    TransactionReceipt,
+} from "@hashgraph/sdk";
+
+import { sdk } from "../sdk_data";
+import { getKeyFromString } from "./key";
+import { applyCommonTransactionParams } from "../utils/common-tx-params";
+import {
+    CreateAccountParams,
+    CreateAccountResponse,
+} from "../models/account-methods";
+
+export const createAccount = async ({
+    key,
+    initialBalance,
+    receiverSignatureRequired,
+    maxAutoTokenAssociations,
+    commonTransactionParams,
+    stakedAccountId,
+    stakedNodeId,
+    declineStakingReward,
+    memo,
+    autoRenewPeriod,
+    alias,
+}: CreateAccountParams): Promise<CreateAccountResponse> => {
+    let transaction = new AccountCreateTransaction().setGrpcDeadline(30000);
+
+    if (key != null) {
+        transaction.setKey(getKeyFromString(key));
+    }
+
+    if (initialBalance != null) {
+        transaction.setInitialBalance(Hbar.fromTinybars(initialBalance));
+    }
+
+    if (receiverSignatureRequired != null) {
+        transaction.setReceiverSignatureRequired(receiverSignatureRequired);
+    }
+
+    if (maxAutoTokenAssociations != null) {
+        transaction.setMaxAutomaticTokenAssociations(maxAutoTokenAssociations);
+    }
+
+    if (stakedAccountId != null) {
+        const accountId = AccountId.fromString(stakedAccountId);
+
+        transaction.setStakedAccountId(accountId);
+    }
+
+    if (stakedNodeId != null) {
+        transaction.setStakedNodeId(stakedNodeId);
+    }
+
+    if (declineStakingReward != null) {
+        transaction.setDeclineStakingReward(declineStakingReward);
+    }
+
+    if (memo != null) {
+        transaction.setAccountMemo(memo);
+    }
+
+    if (autoRenewPeriod != null) {
+        transaction.setAutoRenewPeriod(autoRenewPeriod);
+    }
+
+    if (alias != null) {
+        transaction.setAlias(alias);
+    }
+
+    if (commonTransactionParams != null) {
+        applyCommonTransactionParams(
+            commonTransactionParams,
+            transaction,
+            sdk.getClient(),
+        );
+    }
+
+    // Sign the transaction with the client operator private key if not already signed and submit to a Hedera network
+    let txResponse: TransactionResponse;
+    try {
+        txResponse = await transaction.execute(sdk.getClient());
+    } catch (error) {
+        throw new Error(error);
+    }
+
+    // Get the receipt of the transaction
+    let receipt: TransactionReceipt;
+    try {
+        receipt = await txResponse.getReceipt(sdk.getClient());
+    } catch (error) {
+        throw new Error(error);
+    }
+
+    return {
+        accountId: receipt.accountId.toString(),
+        status: receipt.status.toString(),
+    };
+};
--- a/tck/methods/key.ts
+++ b/tck/methods/key.ts
@@ -1,0 +1,310 @@
+import { PrivateKey, PublicKey, KeyList, Key } from "@hashgraph/sdk";
+import { proto } from "@hashgraph/proto";
+
+import { asn1DecodeStringDer } from "../utils/asn1-decoder";
+import { invalidParamError } from "../utils/invalid-param-error";
+
+import { AccountKey } from "../enums/account-key";
+
+import {
+    AccountKeyGenerationParams,
+    AccountKeyGenerationResponse,
+} from "../models/account-key";
+
+const generateKey = async (
+    params: AccountKeyGenerationParams,
+): Promise<AccountKeyGenerationResponse> => {
+    if (
+        params.fromKey &&
+        ![
+            AccountKey.ED25519_PUBLIC_KEY,
+            AccountKey.ECDSA_SECP256K1_PUBLIC_KEY,
+            AccountKey.EVM_ADDRESS,
+        ].includes(params.type as AccountKey)
+    ) {
+        return invalidParamError<AccountKeyGenerationResponse>(
+            "fromKey should only be provided for specific public key types or EVM address",
+        );
+    }
+
+    if (params.threshold && params.type !== AccountKey.THRESHOLD_KEY) {
+        return invalidParamError<AccountKeyGenerationResponse>(
+            "threshold should only be provided for thresholdKey types",
+        );
+    }
+
+    if (
+        params.keys &&
+        ![AccountKey.KEY_LIST, AccountKey.THRESHOLD_KEY].includes(
+            params.type as AccountKey,
+        )
+    ) {
+        return invalidParamError<AccountKeyGenerationResponse>(
+            "keys should only be provided for keyList or thresholdKey types",
+        );
+    }
+
+    if (
+        (params.type === AccountKey.THRESHOLD_KEY ||
+            params.type === AccountKey.KEY_LIST) &&
+        !params.keys
+    ) {
+        return invalidParamError<AccountKeyGenerationResponse>(
+            "keys list is required for generating a KeyList type",
+        );
+    }
+
+    if (params.type === AccountKey.THRESHOLD_KEY && params.threshold === 0) {
+        return invalidParamError<AccountKeyGenerationResponse>(
+            "threshold is required for generating a ThresholdKey type",
+        );
+    }
+
+    const response: AccountKeyGenerationResponse = {
+        key: "",
+        privateKeys: [],
+    };
+
+    try {
+        const { key, privateKeys } = await processKeyRecursively(
+            params,
+            response,
+            false,
+        );
+
+        response.key = key;
+        response.privateKeys = privateKeys;
+    } catch (error) {
+        return invalidParamError<AccountKeyGenerationResponse>(error.message);
+    }
+
+    return response;
+};
+
+const processKeyRecursively = async (
+    params: AccountKeyGenerationParams,
+    response: AccountKeyGenerationResponse,
+    isList: boolean,
+): Promise<AccountKeyGenerationResponse> => {
+    switch (params.type) {
+        case AccountKey.ED25519_PRIVATE_KEY:
+            const ed25519PrivateKey = PrivateKey.generateED25519();
+            if (isList)
+                response.privateKeys.push(ed25519PrivateKey.toStringDer());
+            return {
+                key: ed25519PrivateKey.toStringDer(),
+                privateKeys: response.privateKeys,
+            };
+
+        case AccountKey.ECDSA_SECP256K1_PRIVATE_KEY:
+            const ecdsaPrivateKey = PrivateKey.generateECDSA();
+            if (isList)
+                response.privateKeys.push(ecdsaPrivateKey.toStringDer());
+            return {
+                key: ecdsaPrivateKey.toStringDer(),
+                privateKeys: response.privateKeys,
+            };
+
+        case AccountKey.ED25519_PUBLIC_KEY:
+        case AccountKey.ECDSA_SECP256K1_PUBLIC_KEY:
+            let publicKey: string;
+            let privateKey: string;
+
+            if (params.fromKey) {
+                if (params.type === AccountKey.ED25519_PUBLIC_KEY) {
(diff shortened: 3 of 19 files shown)


tck/methods/key.ts
```diff
@@ -0,0 +1,310 @@
...
+            }
+
+        default:
+            return invalidParamError<AccountKeyGenerationResponse>(
+                "key type not recognized",
+            );
+    }
+};
+
+const getKeyFromString = (
```
0xivanov, line 214: I'd move `getKeyFromString` and all the methods below it in a `key-utils.ts` file in the utils folder.

tck/utils/common-tx-params.ts
```diff
@@ -0,0 +1,56 @@
...
+    Transaction,
+    Hbar,
+    TransactionId,
+    PrivateKey,
+} from "@hashgraph/sdk";
+
+import { ApplyCommonTransactionInputParams } from "../models/common-tx-params";
+import { getKeyFromString } from "../methods/key";
+
+export const applyCommonTransactionParams = (
```
0xivanov, line 12: Can this be moved inside `common-tx-params.ts` in the /params folder alongside with the interface. I think it looks cleaner when the logic is in the same file.
nit: it make sense for `ApplyCommonTransactionInputParams` to be a class, but this still works i guess.

tck/models/sdk.ts
```diff
@@ -0,0 +1,26 @@
...
+
+type Method = {
+    readonly name: string;
+    readonly param: any;
+};
+
+export interface Input {
+    readonly callClass: string;
+    readonly methods: Method[];
+}
```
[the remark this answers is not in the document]
ivaylogarnev-limechain: I needed to design the **Input** interface to use it as a parameter for the mapMethods function in _(mapping.ts)_ and server.addMethod in _(server.ts)_.

tck/models/account-key.ts
```diff
@@ -0,0 +1,11 @@
+export interface AccountKeyGenerationParams {
+    readonly type: string;
+    readonly fromKey?: string;
+    readonly threshold?: number;
+    readonly keys?: AccountKeyGenerationParams[];
+}
+
+export interface AccountKeyGenerationResponse {
```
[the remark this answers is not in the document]
ivaylogarnev-limechain: The reason I organized the models this way is primarily to avoid certain TypeScript anti-patterns, like over-segregating types, which can sometimes lead to unnecessary complexity and harder maintainability. However, I completely agree with you that for the sake of consistency and to align with the existing TCK projects structure, it's important to maintain a unified approach. I’ll go ahead and refactor the code accordingly and update the PR.

tck/methods/account.ts
```diff
@@ -0,0 +1,101 @@
...
+            sdk.getClient(),
+        );
+    }
+
+    // Sign the transaction with the client operator private key if not already signed and submit to a Hedera network
+    let txResponse: TransactionResponse;
+    try {
+        txResponse = await transaction.execute(sdk.getClient());
+    } catch (error) {
+        throw new Error(error);
```
0xivanov, line 86: if you throw the error here, add some comment like "error occurred during transaction execute..." or something like that. Same for receipt.

tck/methods/key.ts
```diff
@@ -0,0 +1,209 @@
...
+
+import { invalidParamError } from "../utils/invalid-param-error";
+import { getKeyFromString, convertToKeyListHex } from "../utils/key";
+
+import { AccountKey } from "../enums/account-key";
+
+import { KeyGenerationParams } from "../params/key";
+import { KeyGenerationResponse } from "../response/key";
+
+const generateKey = async (
```
0xivanov, line 11: remove the async

tck/server.ts
```diff
@@ -0,0 +1,132 @@
...
+const exceptionMiddleware = async (
+    next: SimpleJSONRPCMethod,
+    request: JSONRPCRequest,
+    serverParams: JSONRPCParams,
+) => {
+    try {
+        return await next(request, serverParams);
+    } catch (error) {
+        if (error.status && error.status._code) {
+            return createJSONRPCErrorResponse(
```
0xivanov, line 68: add some logging here, like in go `log.Printf("Error occurred processing JSON-RPC request: %s, Response error: %s", req, err)`.
You can also create different messages for hedera errors and internal errors.

tck/utils/key.ts
```diff
@@ -0,0 +1,107 @@
...
+        const byteArray = Buffer.from(keyString, "hex");
+
+        let decodedKeyList: proto.KeyList;
+        let decodedThresholdKey: proto.ThresholdKey;
+
+        // Attempt to decode as KeyList
+        try {
+            decodedKeyList = proto.KeyList.decode(byteArray);
+        } catch (keyListError) {
+            console.error("Failed to decode as KeyList:", keyListError);
```
0xivanov, line 48: I'd remove the error messages, since this logic relies on some of these methods to fail, so it's normal.
Stick to logging only when we have the top level error in the `exceptionMiddleware`.

tck/server.ts
```diff
@@ -0,0 +1,132 @@
...
+    } catch (err) {
+        console.warn("Port args error! Defaulting to port 80");
+    }
+}
+
+app.listen(port, () => {
+    console.log(`Server is running on port ${port}`);
+});
+
+console.log("-- JSON-RPC JS server running on localhost port " + port);
```
0xivanov, line 132: This message is the same as the one above it.

tck/server.ts
0xivanov: Also, please check how this func handles the `invalidParamError` because this error has it's own response code and should be handled separately from hedera error and internal error.

tck/methods/account.ts
0xivanov: Or just don't catch it at all, i think it would be caught in the middleware.

ivaylogarnev-limechain: Yeah, it's being caught in the middleware. Most likely I left this try/catch block in the code while debugging something earlier.

tck/server.ts
```diff
@@ -64,36 +64,59 @@ const exceptionMiddleware = async (
...
+        if (error.code === 32602) {
+            console.error(
+                `Params error occurred processing JSON-RPC request: ${requestMethod} \nError: ${errorMessage}`,
+            );
 
             return createJSONRPCErrorResponse(
                 request.id,
-                -32603,
+                -32602,
                 "Internal error",
```
0xivanov, line 94: the message should be `Invalid Params` here

tck/server.ts
```diff
@@ -64,36 +64,59 @@ const exceptionMiddleware = async (
...
+        try {
+            parsedError = JSON.parse(error.toString().substring(7));
+        } catch (parseError) {}
+
+        return createJSONRPCErrorResponse(
+            request.id,
+            -32603,
+            "Internal error",
+            {
+                ...parsedError,
```
0xivanov, line 116: This `parsedError` logic is a bit weird and i don't see the point of it. Can be removed in my opinion.

tck/mapping.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+export default async function mapMethods({
+    callClass,
+    methods,
+}: RpcMethodParams): Promise<string> {
+    const cl: any = (await import("@hashgraph/sdk"))[callClass];
+
+    let currentObject: any = new cl();
+    for (let { name, param } of methods) {
+        console.log(`.${name}(${param})`);
```
ivaylonikolov7, line 17: You can remove the additional console.logs, they are going to flood the CLi and it's going to be harder when we use console.logs for debug purposes.

tck/methods/key.ts
```diff
@@ -0,0 +1,207 @@
...
+import { KeyGenerationResponse } from "../response/key";
+
+const generateKey = (params: KeyGenerationParams): KeyGenerationResponse => {
+    if (
+        params.fromKey &&
+        ![
+            AccountKey.ED25519_PUBLIC_KEY,
+            AccountKey.ECDSA_SECP256K1_PUBLIC_KEY,
+            AccountKey.EVM_ADDRESS,
+        ].includes(params.type as AccountKey)
```
ivaylonikolov7, line 18: What do you think if we export this to a 
```javascript 
const isValidFromKeyType = [
            AccountKey.ED25519_PUBLIC_KEY,
            AccountKey.ECDSA_SECP256K1_PUBLIC_KEY,
            AccountKey.EVM_ADDRESS,
        ].includes(params.type as AccountKey)
```

tck/methods/key.ts
```diff
@@ -0,0 +1,207 @@
...
+        return invalidParamError<KeyGenerationResponse>(
+            "threshold should only be provided for thresholdKey types",
+        );
+    }
+
+    if (
+        params.keys &&
+        ![AccountKey.KEY_LIST, AccountKey.THRESHOLD_KEY].includes(
+            params.type as AccountKey,
+        )
```
ivaylonikolov7, line 35: same as previous comment about exporting this to a constant

tck/methods/key.ts
```diff
@@ -0,0 +1,207 @@
...
+
+        case AccountKey.ECDSA_SECP256K1_PRIVATE_KEY:
+            const ecdsaPrivateKey = PrivateKey.generateECDSA();
+            if (isList)
+                response.privateKeys.push(ecdsaPrivateKey.toStringDer());
+            return {
+                key: ecdsaPrivateKey.toStringDer(),
+                privateKeys: response.privateKeys,
+            };
+
```
ivaylonikolov7, line 102: Actually impressed that you used recursion. 200 IQ move here but can we use something simpler? 

Chances are at some points its going to lead to the TCK, eating someone's Mac memory and then he is going to create an issue that we have to resolve.

tck/methods/key.ts
```diff
@@ -0,0 +1,207 @@
...
+            } catch (error) {
+                return invalidParamError<KeyGenerationResponse>(error.message);
+            }
+
+        default:
+            return invalidParamError<KeyGenerationResponse>(
+                "key type not recognized",
+            );
+    }
+};
```
ivaylonikolov7, line 205: If there's one unpopular opinion that I am willing to die on the hill for - it's that switch cases are terrible! 

Lookup table are like a million times better. If you haven't used one I can help you. Best case is that you should be using polymorphism but in your case that's not possible.

tck/methods/key.ts
```diff
@@ -0,0 +1,207 @@
...
+
+export { generateKey };
```
ivaylonikolov7, line 207: Ngl: haven't checked this but `export default generateKey` ?

tck/methods/key.ts
ivaylonikolov7: If you think this is the cleanest approach, I am eager to accept it tho.

tck/methods/account.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    if (stakedAccountId != null) {
+        const accountId = AccountId.fromString(stakedAccountId);
+
+        transaction.setStakedAccountId(accountId);
+    }
+
+    if (stakedNodeId != null) {
+        transaction.setStakedNodeId(stakedNodeId);
+    }
```
ivaylonikolov7, line 49: Again Github doesn't let me quote the whole thing but the end of this is row 73. 

Anyway, this might be one of the few really rare cases where passing all these properties in the `AccountCreateTransaction` constructor is a proper use case. The SDK already checks if these are null so you wouldn't have to write all these checks. 

I think this pattern might be useful for you in the SDK so this might have given some life to this SDK feature so we can maintain it.

tck/params/common-tx-params.ts
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    if (memo) {
+        transaction.setTransactionMemo(memo);
+    }
+
+    if (regenerateTransactionId) {
+        transaction.setRegenerateTransactionId(regenerateTransactionId);
+    }
+
```
ivaylonikolov7, line 56: Same as AccountCreateTransaction. Pass the params in the transaction constructor.

ivaylonikolov7:
Btw the DCO fails. I would recommend you to sign it now so you wouldn't have to do some git gymnastics to sign non-last commits.

tck/methods/key.ts
0xivanov: The json rpc request has a recursive structure and it's easier to do it with recursion.

tck/methods/account.ts
0xivanov: I see what you mean, but some of these parameters still need to be checked for null because we are doing some type conversions in here, for example `Hbar.fromTinybars(initialBalance)` would not work if initialBalance is null. 

Anyways, if it's doable i'm ok with it.

tck/methods/key.ts
ivaylogarnev-limechain: Originally, that file contained 4 or 5 methods, all of which were being exported. I then moved those methods to a different file, leaving only one method in the original file. I forgot to update the export statement to export default when I left the file with a single method.

tck/methods/account.ts
ivaylogarnev-limechain: It is indeed possible to remove some of these checks. However, I believe it is beneficial to retain them for all parameters.

ivaylonikolov7: Resolved in private conversation.

0xivanov (dismissed):
LGMT

tck/server.ts
```diff
@@ -0,0 +1,146 @@
...
+        const errorMessage = error.message || error;
+
+        if (error.status && error.status._code) {
+            console.error(
+                `Hedera error occurred processing JSON-RPC request: ${requestMethod}, Error: ${errorMessage}`,
+            );
+
+            return createJSONRPCErrorResponse(
+                request.id,
+                -32001,
```
ivaylonikolov7, line 77: Can we move this to a constant for better readability. Otherwise looks like a [Magic number](https://en.wikipedia.org/wiki/Magic_number_(programming))

tck/server.ts
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        if (error.code === 32602) {
+            console.error(
+                `Params error occurred processing JSON-RPC request: ${requestMethod} \nError: ${errorMessage}`,
+            );
+
+            return createJSONRPCErrorResponse(
+                request.id,
+                -32602,
```
ivaylonikolov7, line 93: Same as previous comment about magic number.

tck/methods/sdk.ts
```diff
@@ -0,0 +1,35 @@
...
+        if (nodeIp && nodeAccountId && mirrorNetworkIp) {
+            const node = { [nodeIp]: AccountId.fromString(nodeAccountId) };
+            sdk.client = Client.forNetwork(node);
+            clientType = "custom";
+        } else {
+            sdk.client = Client.forTestnet();
+            clientType = "testnet";
+        }
+        sdk.getClient().setOperator(operatorAccountId, operatorPrivateKey);
+        sdk.getClient().setRequestTimeout(30000);
```
ivaylonikolov7, line 25: is there a reason for these `sdk.getClient` methods. Why not just use a `const client = ...` and then read it from the constant. This way you can have an immutable client.It's like a weird mix between the singleton pattern and a class. I might be missing something so you can correct me.

tck/server.ts
```diff
@@ -0,0 +1,146 @@
...
+                -32001,
+                "Hedera Error",
+                {
+                    status: error.status.toString(),
+                    message: errorMessage,
+                },
+            );
+        }
+
+        if (error.code === 32602) {
```
ivaylonikolov7, line 86: Same as previous comment about magic number. I don't think I can find any hardcoded integers like this in the file but if there are please change them.

tck/utils/key.ts
```diff
@@ -0,0 +1,104 @@
...
+                return PrivateKey.fromStringED25519(keyString);
+            }
+        }
+    } catch (error) {
+        console.error("Error parsing key: ", error);
+        return null;
+    }
+
+    return null;
+};
```
ivaylonikolov7, line 35: Seems like this function has more than 1 responsibility. It handles both public keys, private keys, keylists. Maybe separate the parsing of these keys in different methods like `parsePublickey` `parsePrivatekey` and handle the errors in their respective functions. It's going to make the code more readable and reusable.

tck/utils/key.ts
```diff
@@ -0,0 +1,104 @@
...
+        const byteArray = Buffer.from(keyString, "hex");
+
+        let decodedKeyList: proto.KeyList;
+        let decodedThresholdKey: proto.ThresholdKey;
+
+        // Attempt to decode as KeyList
+        try {
+            decodedKeyList = proto.KeyList.decode(byteArray);
+        } catch (keyListError) {
+            // Expected that this may fail if it's not a KeyList
```
ivaylonikolov7, line 48: I think you can do `console.error` here?

tck/utils/key.ts
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        // Attempt to decode as ThresholdKey
+        try {
+            decodedThresholdKey = proto.ThresholdKey.decode(byteArray);
+        } catch (thresholdKeyError) {
+            // Expected that this may fail if it's not a ThresholdKey
```
ivaylonikolov7, line 55: Same, I think it's missing the error message?

tck/utils/key.ts
ivaylogarnev-limechain: Since it's expected, there's no need for this. Please refer to the previous comment for the rationale.

tck/methods/sdk.ts
ivaylogarnev-limechain: Good catch—thanks for pointing this out. The current setup was designed with flexibility in mind, but it seems that using an immutable client behaves as expected. I’ll go ahead and adopt the immutable approach for simplicity. If we find that we need more flexibility in the future, we can revisit and adjust the SDK setup accordingly.

ivaylonikolov7 (approved):
LGMT!

ivaylonikolov7:
btw I just noticed that one of the commits isn't signed. The DCO passes but when we merge to main there is going to be a problem.

0xivanov (approved):

PR #2491, opened 2024-08-27
feat: Add removeSignature/clearAllSignatures methods

ivaylogarnev-limechain:
**Description**:
This PR introduces the removeSignature and clearAllSignatures methods, allowing transactions to manage signatures more flexibly. Specifically, these methods enable the removal of individual signatures or the clearing of all signatures from a transaction after they have been added and signed.

**Related issue(s)**:
#2461 

**Checklist**

### DIFF
--- a/src/transaction/Transaction.js
+++ b/src/transaction/Transaction.js
@@ -789,20 +789,33 @@ export default class Transaction extends Executable {
     /**
      * Add a signature explicitly
      *
-     * This method requires the transaction to have exactly 1 node account ID set
-     * since different node account IDs have different byte representations and
-     * hence the same signature would not work for all transactions that are the same
-     * except for node account ID being different.
+     * This method supports both single and multiple signatures. A single signature will be applied to all transactions,
+     * While an array of signatures must correspond to each transaction individually.
      *
      * @param {PublicKey} publicKey
-     * @param {Uint8Array} signature
+     * @param {Uint8Array | Uint8Array[]} signature
      * @returns {this}
      */
     addSignature(publicKey, signature) {
-        // Require that only one node is set on this transaction
-        // FIXME: This doesn't consider if we have one node account ID set, but we're
-        // also a chunked transaction. We should also check transaction IDs is of length 1
-        this._requireOneNodeAccountId();
+        const isSingleSignature = signature instanceof Uint8Array;
+        const isArraySignature = Array.isArray(signature);
+
+        // Check if it is a single signature with NOT exactly one transaction
+        if (isSingleSignature && this._signedTransactions.length !== 1) {
+            throw new Error(
+                "Signature array must match the number of transactions",
+            );
+        }
+
+        // Check if it's an array but the array length doesn't match the number of transactions
+        if (
+            isArraySignature &&
+            signature.length !== this._signedTransactions.length
+        ) {
+            throw new Error(
+                "Signature array must match the number of transactions",
+            );
+        }
 
         // If the transaction isn't frozen, freeze it.
         if (!this.isFrozen()) {
@@ -817,7 +830,7 @@ export default class Transaction extends Executable {
             return this;
         }
 
-        // Transactions will have to be regenerated
+        // If we add a new signer, then we need to re-create all transactions
         this._transactions.clear();
 
         // Locking the transaction IDs and node account IDs is necessary for consistency
@@ -826,21 +839,22 @@ export default class Transaction extends Executable {
         this._nodeAccountIds.setLocked();
         this._signedTransactions.setLocked();
 
-        // Add the signature to the signed transaction list. This is a copy paste
-        // of `.signWith()`, but it really shouldn't be if `_signedTransactions.list`
-        // must be a length of one.
-        // FIXME: Remove unnecessary for loop.
-        for (const transaction of this._signedTransactions.list) {
-            if (transaction.sigMap == null) {
-                transaction.sigMap = {};
+        const signatureArray = isSingleSignature ? [signature] : signature;
+
+        // Add the signature to the signed transaction list
+        for (let index = 0; index < this._signedTransactions.length; index++) {
+            const signedTransaction = this._signedTransactions.get(index);
+
+            if (signedTransaction.sigMap == null) {
+                signedTransaction.sigMap = {};
             }
 
-            if (transaction.sigMap.sigPair == null) {
-                transaction.sigMap.sigPair = [];
+            if (signedTransaction.sigMap.sigPair == null) {
+                signedTransaction.sigMap.sigPair = [];
             }
 
-            transaction.sigMap.sigPair.push(
-                publicKey._toProtobufSignature(signature),
+            signedTransaction.sigMap.sigPair.push(
+                publicKey._toProtobufSignature(signatureArray[index]),
             );
         }
 
--- a/src/transaction/Transaction.js
+++ b/src/transaction/Transaction.js
@@ -852,13 +852,12 @@ export default class Transaction extends Executable {
     }
 
     /**
-     * This method removes a signature from the transaction based on the public key or signature provided.
+     * This method removes all signatures from the transaction based on the public key provided.
      *
      * @param {PublicKey} publicKey - The public key associated with the signature to remove.
-     * @param {Uint8Array} [signature] - The signature to remove. If omitted, all signatures for the given public key will be removed.
      * @returns {Uint8Array[]} The removed signatures.
      */
-    removeSignature(publicKey, signature) {
+    removeSignature(publicKey) {
         // If the transaction isn't frozen, freeze it.
         if (!this.isFrozen()) {
             this.freeze();
@@ -880,7 +879,6 @@ export default class Transaction extends Executable {
                 this._removeSignaturesFromTransaction(
                     transaction,
                     publicKeyHex,
-                    signature,
                 );
 
             removedSignatures.push(...removedSignaturesFromTransaction);
@@ -906,14 +904,14 @@ export default class Transaction extends Executable {
      *
      * @returns {{ [transactionId: string]: { [nodeAccountId: string]: Uint8Array[] } }} The removed signatures in the specified format.
      */
-    clearAllSignatures() {
+    removeAllSignatures() {
         // If the transaction isn't frozen, freeze it.
         if (!this.isFrozen()) {
             this.freeze();
         }
 
         // Collect all signatures
-        const removedSignatures = this._collectSignaturesInSpecificFormat();
+        const removedSignatures = this._collectSignaturesByPublicKey();
 
         // Iterate over the signed transactions and clear all signatures
         for (const transaction of this._signedTransactions.list) {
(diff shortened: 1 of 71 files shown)


src/transaction/Transaction.js
```diff
@@ -851,6 +851,86 @@ export default class Transaction extends Executable {
...
+    /**
+     * This method removes a signature from the transaction based on the public key or signature provided.
+     *
+     * @param {PublicKey} publicKey - The public key associated with the signature to remove.
+     * @param {Uint8Array} [signature] - The signature to remove. If omitted, all signatures for the given public key will be removed.
+     * @returns {Uint8Array[]} The removed signatures.
+     */
+    removeSignature(publicKey, signature) {
+        // If the transaction isn't frozen, freeze it.
+        if (!this.isFrozen()) {
```
0xivanov, line 863: Do we need this? I mean if we signed we already froze the txn?

src/transaction/Transaction.js
```diff
@@ -851,6 +851,86 @@ export default class Transaction extends Executable {
...
     }
 
+    /**
+     * This method removes a signature from the transaction based on the public key or signature provided.
+     *
+     * @param {PublicKey} publicKey - The public key associated with the signature to remove.
+     * @param {Uint8Array} [signature] - The signature to remove. If omitted, all signatures for the given public key will be removed.
+     * @returns {Uint8Array[]} The removed signatures.
+     */
+    removeSignature(publicKey, signature) {
```
jbair06, line 861: I'm not sure if removing a singular signature would actually have a use case. Removing all signatures for a specific publicKey does. Can this function be simplified? Or is there a use case for removing a single signature in this manner? In either a multi-node or chunked transaction use case, removing a single signature without removing all signatures for a publicKey might be problematic.

jbair06: My original idea was that if a signature is used instead of a public key, the function could perhaps try and determine the public key tied to the signature and remove all signatures for that public key. But after thinking about it further, I think it is not the best approach, and it would be better to just rely on the public key.

src/transaction/Transaction.js
```diff
@@ -851,6 +851,86 @@ export default class Transaction extends Executable {
...
+
+        if (!this._signerPublicKeys.has(publicKeyHex)) {
+            throw new Error("The public key has not signed this transaction");
+        }
+
+        /** @type {Uint8Array[]} */
+        const removedSignatures = [];
+
+        // Iterate over the signed transactions and remove matching signatures
+        for (const transaction of this._signedTransactions.list) {
```
jbair06, line 878: Is the order of the _signedTransaction list locked at this point and unchangeable? If the order is trusted to remain the same throughout the entire add/remove signature calls, then the Uint8Array[] is a good structure for the return value. If the list order isn't trustable, then this structure could be problematic. 
Does the list remain locked even if the transaction is converted to bytes and back into a transaction? Or does the lock get lost in the conversion.

src/transaction/Transaction.js
```diff
@@ -851,6 +851,86 @@ export default class Transaction extends Executable {
...
+        return removedSignatures;
+    }
+
+    /**
+     * This method clears all signatures from the transaction and returns them in a specific format.
+     *
+     * It will call collectSignatures to get the removed signatures, then clear all signatures
+     * from the internal tracking.
+     *
+     * @returns {{ [transactionId: string]: { [nodeAccountId: string]: Uint8Array[] } }} The removed signatures in the specified format.
```
jbair06, line 907: The data structure for this return type should probably be `{ [publicKey: PublicKey]: Uint8Array[] }` to be consistent with the `removeSignature` function, I would think. Or is that not preferable?

src/transaction/Transaction.js
ivaylogarnev-limechain: A single public key can only be associated with one signature at a time. The addSignature() method already enforces this by checking if a signature has already been added for that public key. Given this, I agree that we can safely rely on the public key alone when removing signatures. I initially included the signature parameter because the original method proposal included both a public key and a signature, which led me to believe there might be a specific use case in mind. However, I totally agree that it would be simpler and more effective to rely only on the public key.

src/transaction/Transaction.js
ivaylogarnev-limechain: The concern about the order of the _signedTransaction list in the context of serialization and deserialization is valid. Based on Hedera's transaction model, when a transaction is serialized into bytes (using .toBytes()) and then deserialized back (using Transaction.fromBytes()), the transaction structure—including the order of signatures—should be preserved.

However, once a transaction is frozen using freezeWith() or freeze(), it becomes immutable. This immutability should include locking the order of signatures within the transaction, ensuring that the order remains consistent through the entire lifecycle of the transaction, including during serialization and deserialization.

I will make further investigations to be 100% sure on this.

src/transaction/Transaction.js
ivaylogarnev-limechain: I'll rename the function to removeAllSignatures for consistency. I also agree that it would be simpler and more consistent to return the publicKey associated with the signatures. Since each publicKey can only have one associated signature, the return type won't need to be an array. This approach aligns better with the logic in the removeSignature function.

src/transaction/Transaction.js
ivaylogarnev-limechain: In most cases, yes, but we also have this check in the addSignature() method since it interacts with states, and we need to ensure that it is frozen.

test/unit/Transaction.js
```diff
@@ -313,4 +313,227 @@ describe("Transaction", function () {
...
+            const transactionId = new TransactionId(account, validStart);
+
+            key1 = PrivateKey.generateED25519();
+            key2 = PrivateKey.generateED25519();
+            key3 = PrivateKey.generateED25519();
+
+            transaction = new AccountCreateTransaction()
+                .setInitialBalance(new Hbar(2))
+                .setTransactionId(transactionId)
+                .setNodeAccountIds([nodeAccountId])
```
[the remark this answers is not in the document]
ivaylogarnev-limechain: fixed

src/transaction/Transaction.js
ivaylonikolov7: My only concern is the transfer transaction. When you convert it to bytes it sorts the transactions inside and it might cause some issues. If you can please check if transfer transaction with different token ids works as expected. Keep in mind that the order of `.addTokenTransfer` should be not... well, in order :D So if there is a token with id `0.0.1`, `0.0.2` and `0.0.3` don't pass the token transfers:
```javascript
addTokenTransfer(0.0.1, sender, receiver, amount)
addTokenTransfer(0.0.2 sender, receiver, amount)
addTokenTransfer(0.0.3, sender, receiver, amount)
```

src/transaction/Transaction.js
```diff
@@ -1745,6 +1823,99 @@ export default class Transaction extends Executable {
...
+                    );
+
+                    if (shouldRemove) {
+                        if (sigPair.ed25519) {
+                            removedSignatures.push(sigPair.ed25519);
+                        } else if (sigPair.ECDSASecp256k1) {
+                            removedSignatures.push(sigPair.ECDSASecp256k1);
+                        }
+                    }
+
```
ivaylonikolov7, line 1853: `removedSignatures.push(sigPair.ed25519 || sigpair.ECDSASecp256k1);`

src/transaction/Transaction.js
```diff
@@ -1745,6 +1823,99 @@ export default class Transaction extends Executable {
...
+     *
+     * @param {HashgraphProto.proto.ISignedTransaction} transaction - The transaction object to process.
+     * @param {string} publicKeyHex - The hexadecimal representation of the public key.
+     * @returns {Uint8Array[]} An array of removed signatures.
+     */
+    _removeSignaturesFromTransaction(transaction, publicKeyHex) {
+        /** @type {Uint8Array[]} */
+        const removedSignatures = [];
+
+        if (transaction.sigMap && transaction.sigMap.sigPair) {
```
ivaylonikolov7, line 1838: ` if (!transaction.sigMap && transaction.sigMap.sigPair) return []; `

src/transaction/Transaction.js
```diff
@@ -851,6 +851,84 @@ export default class Transaction extends Executable {
...
+    /**
+     * This method clears all signatures from the transaction and returns them in a specific format.
+     *
+     * It will call collectSignatures to get the removed signatures, then clear all signatures
+     * from the internal tracking.
+     *
+     * @returns {{ [transactionId: string]: { [nodeAccountId: string]: Uint8Array[] } }} The removed signatures in the specified format.
+     */
+    removeAllSignatures() {
+        // If the transaction isn't frozen, freeze it.
```
ivaylonikolov7, line 908: No need for this comment. code is pretty obvious.

src/transaction/Transaction.js
```diff
@@ -851,6 +851,84 @@ export default class Transaction extends Executable {
...
+        if (!this.isFrozen()) {
+            this.freeze();
+        }
+
+        // Collect all signatures
```
ivaylonikolov7, line 913: Same as above. Unless there's some really weird logic I avoid adding comments. It's bloating the code. I also add these comments for the community for the js sdk examples.

src/transaction/Transaction.js
```diff
@@ -851,6 +851,84 @@ export default class Transaction extends Executable {
...
+        const removedSignatures = this._collectSignaturesByPublicKey();
+
+        // Iterate over the signed transactions and clear all signatures
+        for (const transaction of this._signedTransactions.list) {
+            if (transaction.sigMap && transaction.sigMap.sigPair) {
+                // Clear all signature pairs from the transaction's signature map
+                transaction.sigMap.sigPair = [];
```
ivaylonikolov7, line 920: wouldn't optional chaining work here? 
` transaction?.sigMap?.sigPair=[]`

src/transaction/Transaction.js
```diff
@@ -1745,6 +1823,99 @@ export default class Transaction extends Executable {
...
+                        // Initialize the structure for this publicKey if it doesn't exist
+                        if (!collectedSignatures[publicKeyHex]) {
+                            collectedSignatures[publicKeyHex] = [];
+                        }
+
+                        // Add the signature to the corresponding public key
+                        collectedSignatures[publicKeyHex].push(signature);
+                    }
+                }
+            }
```
ivaylonikolov7, line 1914: Same as above. Please:
- Use the logical OR ( || ) to determine the signature pair.
- `return []` if there is no sigmap or sigpair.
-  also `for(const {sigPair} ... )`
- `if (!signature && !sigPair.pubKeyPrefix) return;`

Now that I think of it maybe we should return empty arrays, `undefined`, `null` but actually throw an error it might help with future debugging.

src/transaction/Transaction.js
ivaylogarnev-limechain: The if statement cannot be eliminated because it ensures that both `transaction` and `sigMap` exist before attempting to assign to `sigPair`. At best, it could be written as `if (transaction?.sigMap?.sigPair)`, though I find that approach less readable.

src/transaction/Transaction.js
ivaylogarnev-limechain: I don't think this is actually an error. If you run the collect method on a transaction without signatures, it would be expected to return an empty array if there are no signatures present.

ivaylonikolov7 (dismissed):
LGMT, gj.

agadzhalov (dismissed):
LGTM

ivaylonikolov7 (dismissed):

agadzhalov (approved):
LGTM

agadzhalov (approved):

Issue #2458, opened 2024-08-16
Expo: Generating a private key from a mnemonic enters an infinite loop that freezes the application


ivaylonikolov7:
### Description

This issue happens in Expo environment. When you try to get a private key(it doesn't matter if ECDSA or ED25519) from a mnemonic it enters an infinite loop that causes the application to freeze(both the browser or the simulator freeze/crash). Tested this both on Android and the web version of the expo project. Same outcome. So far I have noticed that the seed phrase doesn't matter and happens on every seed. 

This issue isn't reproducible in node.js environment. Only in Expo environment! 

### Steps to reproduce

Run this code snippet in an Expo project:
```javascript
 const mnemonic = [
      "outer",
      "drastic",
      "print",
      "brave",
      "edge",
      "student",
      "enemy",
      "matrix",
      "chest",
      "just",
      "away",
      "robust",
      "inspire",
      "indicate",
      "trend",
      "beyond",
      "pony",
      "wood",
      "truly",
      "sword",
      "more",
      "major",
      "regular",
      "vicious",
    ];

    const generatedMnemonic = await Mnemonic.fromWords(mnemonic);

    console.log("Attempting to generate rootPrivateKeyNew3");
    let rootPrivateKeyNew3 =
      await generatedMnemonic.toStandardECDSAsecp256k1PrivateKey();
```

Start the application via an Android or iOS simulator. Web freezes the browser tab too. 

### Hedera network

mainnet, testnet, previewnet, other

### Version

2.48.1 but happens in others too

### Operating system

macOS

ivaylonikolov7:
After thorough investigation this issue happens due to incompatibility of Buffers in environment different than node.js.This is the code that causes this issue:

```javascript
export async function derive(parentKey, chainCode, index) {
    const isHardened = isHardenedIndex(index);
    const data = new Uint8Array(37);

    const publicKey = hex.decode(
        secp256k1.keyFromPrivate(parentKey).getPublic(true, "hex"),
    );

    // Hardened child
    if (isHardened) {
        // data = 0x00 || ser256(kpar) || ser32(index)
        data[0] = 0x00;
        data.set(parentKey, 1);

        // Normal child
    } else {
        // data = serP(point(kpar)) || ser32(index)
        //      = serP(Kpar) || ser32(index)
        data.set(publicKey, 0);
    }

    new DataView(data.buffer, data.byteOffset, data.byteLength).setUint32(
        33,
        index,
        false,
    );

[...]

PR #2502, opened 2024-09-04
fix: React Native's BIP32 implementation should use array instead of Buffer

ivaylonikolov7:
**Description**:
We should use Arrays instead of Buffers whenever possible because of React Native's inability to deal with Buffers BIP32. This PR changes `packages/cryptography/src/primitive/Bip32.native.js`. There is no need to change `packages/cryptography/src/primitive/BIP32.js` as it is used in different environment. And the web version of this file already uses the correct functions. 

**Related issue(s)**:
#2458 

Fixes #
#2458

**Checklist**


agadzhalov (approved):
LGTM

PR #2509, opened 2024-09-10
Feat/2481 multi node multi sig

ivaylogarnev-limechain:
**DRAFT  PR** 

ONLY FOR PRESENTATION PURPOSES

### DIFF
--- a/src/PublicKey.js
+++ b/src/PublicKey.js
@@ -111,7 +111,6 @@ export default class PublicKey extends Key {
      * @param {Uint8Array} message
      * @param {Uint8Array} signature
      * @returns {boolean}
-     * @throws {error}
      */
     verify(message, signature) {
         return this._key.verify(message, signature);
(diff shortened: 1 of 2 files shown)


src/PrivateKey.js
jbair06, line 1: The reason that I suggested to put this into the sign function was due to my personal use case. The specific use case I was looking to cover involves signing a multi-node transaction offline. Like the examples, the flow would be: 
1. create the transaction
2. send the bytes to all signers
3. they each sign the transaction
4. return the signature bytes to the creator of the transaction
5. add all signatures to the transaction

In this flow, adding the signature to the transaction itself was unnecessary. However, now that you've moved this functionality to signTransaction, I feel like it should be in both locations, so as to allow the developer to create signatures for both single and multi-node transactions no matter which function they need to use.

Do you think it would make sense to add this functionality in both locations? Or is that unnecessary?

PR #2532, opened 2024-09-18
fix: serialize deserialize FileAppend

### LINKED ISSUE #2167: FileAppendTransaction incorrect content after deserialization
### Description

Serializing and then deserializing a FileAppendTransaction causes incorrect content to be set.
After the deserialization, the content is sliced to the chunk size.
Ex. If you have a content length of 100_000 after the deserialization, the content will consist only of the first {chunkSize} bytes.

### Steps to reproduce

1. Run this snippet
```
const { FileAppendTransaction } = require('@hashgraph/sdk');

async function run() {
  const contentLength = 632421;
  const content = generateUInt8Array(contentLength);

  const transaction = new FileAppendTransaction()
    .setTransactionValidDuration(180)
    .setContents(content);

  console.log(transaction.contents.length); //632421 (content length)

  const transactionBytes = transaction.toBytes();

  const transactionFromBytes =
    FileAppendTransaction.fromBytes(transactionBytes);

  console.log(transactionFromBytes.contents.length); //4096 (default chunk size)
}
run();

function generateUInt8Array(length) {
  const uint8Array = new Uint8Array(length);

  for (let i = 0; i < length; i++) {
    uint8Array[i] = Math.floor(Math.random() * 256);
  }

  return uint8Array;
}
```

### Additional context

_No response_

### Hedera network

other

### Version

v2.42.0-beta.4

### Operating system

None

ivaylonikolov7:
**Description**:
Implement chunk logic for toBytes method and disallow addSignature function until the next release. 

Fixes #2167 #2536

**Notes for reviewer**:
There should be refactoring done. We need to implement `ChunkTransaction`. This fix is more like a patch before the major refactor for ChunkedTransaction like the other SDKs. 

**Checklist**

### DIFF
--- a/src/file/FileAppendTransaction.js
+++ b/src/file/FileAppendTransaction.js
@@ -385,6 +385,12 @@ export default class FileAppendTransaction extends Transaction {
      * @returns {Promise<TransactionResponse[]>}
      */
     async executeAll(client, requestTimeout) {
+        if (this.maxChunks && this.requiredChunks > this.maxChunks) {
+            throw new Error(
+                `cannot execute \`FileAppendTransaction\` with more than ${this.maxChunks} chunks`,
+            );
+        }
+
         if (!super._isFrozen()) {
             this.freezeWith(client);
         }
@@ -502,6 +508,17 @@ export default class FileAppendTransaction extends Transaction {
 
         this._transactionIds.advance();
         this._transactionIds.setLocked();
+    }
+
+    /**
+     * Build all the signed transactions
+     * @override
+     * @internal
+     */
+    _buildAllTransactions() {
+        for (let i = 0; i < this._signedTransactions.length; i++) {
+            this._buildTransaction(i);
+        }
     }
 
     /**
--- a/src/file/FileAppendTransaction.js
+++ b/src/file/FileAppendTransaction.js
@@ -658,10 +658,17 @@ export default class FileAppendTransaction extends Transaction {
             throw new Error("fileID is required");
         }
 
-        return new FileAppendTransaction({
-            fileId: fileId,
-            contents: contents,
-        });
+        return Transaction._fromProtobufTransactions(
+            new FileAppendTransaction({
+                fileId: fileId,
+                contents: contents,
+            }),
+            [],
+            signedTransactions,
+            transactionIds,
+            nodeIds,
+            bodies,
+        );
     }
 
     /**
--- a/src/file/FileAppendTransaction.js
+++ b/src/file/FileAppendTransaction.js
@@ -673,6 +673,25 @@ export default class FileAppendTransaction extends Transaction {
         );
         return `FileAppendTransaction:${timestamp.toString()}`;
     }
+
+    /**
+     * @override
+     * @protected
+     * @returns {HashgraphProto.proto.IFileAppendTransactionBody}
+     */
+    _makeTransactionData() {
+        const length = this._contents != null ? this._contents.length : 0;
+        const startIndex = this._transactionIds.index * this._chunkSize;
+        const endIndex = Math.min(startIndex + this._chunkSize, length);
+
+        return {
+            fileID: this._fileId != null ? this._fileId._toProtobuf() : null,
+            contents:
+                this._contents != null
+                    ? this._contents.slice(startIndex, endIndex)
+                    : null,
+        };
+    }
 }
 
 // eslint-disable-next-line @typescript-eslint/unbound-method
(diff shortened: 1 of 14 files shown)


0xivanov (dismissed):
LGTM

test/integration/FileAppendIntegrationTest.js
gsstoykov, line 1: Since this functionality is flaky from past experience, try creating more integration tests IMHO no need to revisit this issue.

src/file/FileAppendTransaction.js
```diff
@@ -445,6 +457,90 @@ export default class FileAppendTransaction extends Transaction {
         return "fileAppend";
     }
 
+    /**
+     * Build all the transactions when transacgtions are not signed
+     * @override
+     * @internal
+     */
+    _buildIncompletedTransactions() {
```
gsstoykov, line 465: nit: *Incomplete

test/integration/FileAppendIntegrationTest.js
```diff
@@ -211,7 +215,67 @@ describe("FileAppend", function () {
         ).getReceipt(env.client);
     });
 
+    it("should keep content after deserialization", async function () {
+        this.timeout(120000);
```
gsstoykov, line 219: I really don't like that we have to do this every single time. Consider working around this or making it configurable.

gsstoykov: Maybe create an issue for it?

src/file/FileAppendTransaction.js
```diff
@@ -445,6 +457,90 @@ export default class FileAppendTransaction extends Transaction {
...
+
+            nextTransactionId = new TransactionId(
+                /** @type {AccountId} */ (nextTransactionId.accountId),
+                new Timestamp(
+                    /** @type {Timestamp} */ (
+                        nextTransactionId.validStart
+                    ).seconds,
+                    /** @type {Timestamp} */ (
+                        nextTransactionId.validStart
+                    ).nanos.add(1),
```
gsstoykov, line 509: Is this adding one nanosecond? Why don't you just generate a transaction id here?

src/transaction/Transaction.js
```diff
@@ -800,6 +810,11 @@ export default class Transaction extends Executable {
         const isSingleSignature = signature instanceof Uint8Array;
         const isArraySignature = Array.isArray(signature);
 
+        if (this.getRequiredChunks() > 1) {
+            throw new Error(
+                "Add signature is not supported for chunked transactions",
+            );
+        }
```
gsstoykov, line 817: What if you have a ChunkedTransacrtion with one chunk?

ivaylonikolov7: Just added an integration test for this case. I think it works properly as of right now but it will probably going to get thoroughly tested in the new version when we have to add the signature functionality for file append too.

src/file/FileAppendTransaction.js
ivaylonikolov7: Hm, refactored this to generating `TransactionId` - `TransactionId.generate(dummyAccountId);`. You meant like this, right?

test/integration/FileAppendIntegrationTest.js
ivaylonikolov7: I added a bunch of inttegration tests. If you think it's missing some I am open to adding them. I agree that the functionality is flaky.

test/integration/FileAppendIntegrationTest.js
ivaylonikolov7: We can work around this by editing the Taskfile that runs the integration tests.

src/file/FileAppendTransaction.js
ivaylonikolov7: Hm, I changed it but now that I think about it shouldn't it be `Incompleted`? :D

src/file/FileAppendTransaction.js
gsstoykov: The dummyAccountId is just a temp value. Why not `TransactionId.generate(nextTransactionId.accountId);` ? Or that won't work for your use case.

src/transaction/Transaction.js
gsstoykov: Where is this integration test? Couldn't find...

gsstoykov: my bad nvm

ivaylonikolov7: `FileAppendIntegrationTest.js` - "should be able to sign transaction with 1 chunk"

gsstoykov (approved):

agadzhalov (approved):
LGTM

PR #2559, opened 2024-10-03
feat: Changed the return type of removeAllSignatures method

### LINKED ISSUE #2546: removeAllSignatures return type should be map
### Problem

The removeAllSignatures function returns a dictionary of type {{ [publicKeyString: string]: Uint8Array[] }}. This is because JavaScript only deals with strings and numbers as keys in a dictionary. In order to maintain consistency with the other remove/add signature methods, this return type should be a map.

### Solution

Use a map instead of a dictionary as the return type, thus allowing the PublicKey to be used as the key, instead of a string. 

### Alternatives

_No response_

ivaylogarnev-limechain:
**Description**:
This PR changes the return type of the **removeAllSignatures()** method in the **Transaction** class. Additionally, it addresses and fixes the related unit and integration tests to ensure compatibility with this change

Fixes 
#2546

**Checklist**

### DIFF
--- a/examples/multi-node-multi-signature-removeAll.js
+++ b/examples/multi-node-multi-signature-removeAll.js
@@ -7,7 +7,6 @@ import {
     KeyList,
     TransferTransaction,
     Transaction,
-    PublicKey,
 } from "@hashgraph/sdk";
 
 import dotenv from "dotenv";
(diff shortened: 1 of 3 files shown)


ivaylonikolov7:
LGTM but considering that the codecov fails... Can we add unit tests for L1920 and L1928 for Transaction.js. I think if we keep adding unit tests at some point it's going to stop crying that it doesn't reach the threshhold.

ivaylonikolov7 (commented):
LGTM but considering that the codecov fails... Can we add unit tests for L1920 and L1928 for Transaction.js. I think if we keep adding unit tests at some point it's going to stop crying that it doesn't reach the threshold. 

It's something I tried to do in my [PendingAirdropId PR](https://github.com/hashgraph/hedera-sdk-js/pull/2567). I tried to reach 100% codecov. It doesn't fail consistently for some reason but I think its because we are on the edge of the threshold and this is why it sometimes fails and sometimes passes.

ivaylonikolov7 (approved):
LGTM

agadzhalov (approved):
LGTM

PR #2567, opened 2024-10-07
feat: expose PendingAirdropId

### LINKED ISSUE #2560: Expose PendingAirdropId
### Problem

Expose PendingAirdropId in `exports.js`

### Solution

Look above. 

### Alternatives

_No response_

ivaylonikolov7:
**Description**:
- Exposes `PendingAirdropId` in `exports.js` so it's used for our users
- Adds unit tests for PendingAirdropId to increase codecoverage.

**Related issue(s)**:

Fixes #2560 

**Checklist**


src/token/PendingAirdropId.js
```diff
@@ -121,6 +122,7 @@ export default class PendingAirdropId {
      * @returns {this}
      */
     setNftId(nftId) {
+        this._tokenId = null;
```
gsstoykov, line 125: Isn't `_tokenId` null set in the constructor?

src/token/PendingAirdropId.js
```diff
@@ -33,10 +33,10 @@ export default class PendingAirdropId {
...
-     * @param {TokenId?} props.tokenId
-     * @param {NftId?} props.nftId
+     * @param {TokenId?} [props.tokenId]
+     * @param {NftId?} [props.nftId]
      */
-    constructor(props) {
+    constructor(props = {}) {
         this._senderId = null;
         this._receiverId = null;
         this._tokenId = null;
```
gsstoykov, line 42: Why not set nftId to null here as well?

src/token/PendingAirdropId.js
ivaylonikolov7: PendingAirdropId should have either tokenId or nftid. So in case he sets nftId we should null the nftId and vice versa.

src/token/PendingAirdropId.js
ivaylonikolov7: what do you mean? It's set to null in the same line 🤔

gsstoykov: Wasn't expanded sorry.

src/token/PendingAirdropId.js
gsstoykov: nit: Would expect the setter to be with an individual responsibility. But in this specific case it is ok not to change it since it will trigger other changes for you.

PR #2569, opened 2024-10-08
fix: lock protobufjs version

ivaylonikolov7:
**Description**:
Due to a bug in protobufjs version we need to lock `protobufjs` version so we can't groom some of our PRs. This PR updates the package.json and pnpm-lock.files. Deletes pnpm-lock file in the TCK folder due to having both npm-lock and pnpm-lock. Good practices say we should only have one lock file. 

**Checklist**


agadzhalov (approved):
LGTM

PR #2576, opened 2024-10-12
Feature/2511 create account with threshold key

b-l-u-e:
**Description**:
This PR introduces an example for creating an account with a threshold key in the Hedera JavaScript SDK, which mirrors the existing example in the Hedera Java SDK.

**Problem**
The Hedera Java SDK already includes an example (CreateAccountThresholdKeyExample.java) that demonstrates how to create an account with a threshold key. This feature is essential for ensuring security by requiring multiple signatures to authorize transactions.

However, there is currently no such example in the Hedera JavaScript SDK. Developers using the JavaScript SDK would greatly benefit from a similar example to guide them through creating accounts with threshold keys in JavaScript, as it's a crucial security feature.

**Solution**

This PR provides a new example in the JavaScript SDK that follows the structure of the existing Java example. Key steps include:

- Generating key pairs using PrivateKey.generateED25519().
- Initializing a KeyList with a threshold and adding public keys to create the threshold key.
- Creating an account using the threshold key.
- Demonstrating a transaction that requires multiple signatures from the threshold key.
- Cleaning up by deleting the account, again requiring multi-signatures.

**Related issue(s)**:

Fixes # 2511


ivaylogarnev-limechain:
Thank you so much for you work! I just have a small request regarding the PR. Could you please create a new PR by cloning from the main branch and include the new file in a single commit? This will help avoid any issues with references or CI errors due to the existing commits and also there are some conflicting files. One thing also, you can change the name of the PR as the "get file contents" to be consistent.

Thanks again for your contribution, and I appreciate your understanding!

b-l-u-e:
@ivaylogarnev-limechain  thanks for the feedback. 
I have a question. since there are 2 issues am working on right now. and based on the suggestion that i create a new PR and use get file contents" does it mean i dont have to create 2 pull requests for two separate issues. i can still maintain one PR as i work on different issues

b-l-u-e:
I created the PR what happens to this now? should i close it or delete it? Please advice

PR #2583, opened 2024-10-15
fix: file append transaction id after deserialization

SvetBorislavov:
**Description**:
This pull request:
- Fixes the lost transaction ID after deserialization, if previously set
- Adds chunk interval property to bring flexibility to the interval between chunks. This is helpful if a user tries to update a file that will take longer than the transaction's `validDuration`.

**Related issue(s)**:

Fixes Partially #2167 

**Checklist**

- [X] Documented (Code comments)
- [X] Tested (unit, integration)

### DIFF
--- a/test/integration/FileAppendIntegrationTest.js
+++ b/test/integration/FileAppendIntegrationTest.js
@@ -407,8 +407,10 @@ describe("FileAppend", function () {
         expect(receipt.status).to.be.equal(Status.Success);
     });
 
-    it("should keep chunk size and correct maxChunks after deserialization", async function () {
+    it("should keep chunk size, chunk interval and correct max chunks after deserialization", async function () {
         const operatorKey = env.operatorKey.publicKey;
+        const chunkSize = 1024;
+        const chunkInterval = 230;
 
         let response = await new FileCreateTransaction()
             .setKeys([operatorKey])
@@ -419,7 +421,8 @@ describe("FileAppend", function () {
 
         const tx = new FileAppendTransaction()
             .setFileId(fileId)
-            .setChunkSize(1024)
+            .setChunkSize(chunkSize)
+            .setChunkInterval(chunkInterval)
             .setMaxChunks(99999)
             .setContents(newContents);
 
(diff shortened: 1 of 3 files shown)


test/unit/FileAppendTransaction.js
```diff
@@ -88,4 +88,52 @@ describe("FileAppendTransaction", function () {
...
+            .setNodeAccountIds([nodeAccountId])
+            .setFileId(fileId)
+            .setChunkSize(1000)
+            .setChunkInterval(chunkInterval)
+            .setContents("1".repeat(1000) + "2".repeat(1000) + "3".repeat(1000))
+            .freeze();
+
+        expect(transaction._transactionIds.list.length).to.be.equal(3);
+
+        for (let i = 0; i < 3; i++) {
```
ivaylonikolov7, line 112: Can we have the content separated in another variable/const:

`const contents = "1".repeat(1000) + "2".repeat(1000) + "3".repeat(1000)
const requiredChunks = contents.length/chunkSize` 

This way we can have `for(let I=0; I< requiredChunks; I++` and then inside the tests I don't like the `if`s. Can't we just check the difference between current transaction and previous transaction to equal chunkInterval?

test/integration/FileAppendIntegrationTest.js
```diff
@@ -351,6 +354,59 @@ describe("FileAppend", function () {
...
+            (transactionId, index, array) => {
+                if (index > 0) {
+                    const previousTimestamp = array[index - 1].validStart;
+                    const currentTimestamp = transactionId.validStart;
+                    const difference =
+                        currentTimestamp.nanos - previousTimestamp.nanos;
+                    expect(difference).to.be.equal(chunkInterval);
+                }
+            },
+        );
```
ivaylonikolov7, line 399: nit: Big brain move would have been using `.reduce` but I think this solution is more readable.

ivaylonikolov7 (commented):
Thanks for the contribution: 🙌 The chunk interval is an awesome feature.

test/integration/FileAppendIntegrationTest.js
SvetBorislavov: You are right, it can be done using `.reduce` but I think readability is more important than test optimization.

test/unit/FileAppendTransaction.js
SvetBorislavov: Adressed

test/integration/FileAppendIntegrationTest.js
ivaylonikolov7: I agree.

ivaylonikolov7 (approved):
LGTM!

agadzhalov (approved):
LGTM

PR #2601, opened 2024-10-22
feat: init commit for signing of tx changes

ivaylonikolov7:
**Description**:
TBD

This PR modifies ... in order to support ...
TBD

**Related issue(s)**:
TBD 

Fixes #
TBD 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/src/transaction/NodeAccountIdSignatureMap.js
+++ b/src/transaction/NodeAccountIdSignatureMap.js
@@ -19,36 +19,38 @@
  */
 
 import ObjectMap from "../ObjectMap.js";
-import PublicKey from "../PublicKey.js";
+import TransactionId from "./TransactionId.js";
+import SignaturePairMap from "./SignaturePairMap.js";
+import * as HashgraphProto from "@hashgraph/proto";
 
 /**
- * @augments {ObjectMap<PublicKey, Uint8Array>}
+ * @augments {ObjectMap<TransactionId, SignaturePairMap>}
  */
 export default class NodeAccountIdSignatureMap extends ObjectMap {
     constructor() {
-        super((s) => PublicKey.fromString(s));
+        super((s) => TransactionId.fromString(s));
     }
 
     /**
-     * @param {import("@hashgraph/proto").proto.ISignatureMap} sigMap
+     * @param { import('./List.js').default<import("@hashgraph/proto").proto.ISignedTransaction>} signedTransactions
      * @returns {NodeAccountIdSignatureMap}
      */
-    static _fromTransactionSigMap(sigMap) {
+    static _fromSignedTransactions(signedTransactions) {
         const signatures = new NodeAccountIdSignatureMap();
 
-        const sigPairs = sigMap.sigPair != null ? sigMap.sigPair : [];
+        for (const { bodyBytes, sigMap } of signedTransactions.list) {
+            if (bodyBytes != null && sigMap != null) {
+                const body =
+                    HashgraphProto.proto.TransactionBody.decode(bodyBytes);
 
-        for (const sigPair of sigPairs) {
-            if (sigPair.pubKeyPrefix != null) {
-                if (sigPair.ed25519 != null) {
-                    signatures._set(
-                        PublicKey.fromBytesED25519(sigPair.pubKeyPrefix),
-                        sigPair.ed25519,
+                if (body.transactionID != null) {
+                    const transactionId = TransactionId._fromProtobuf(
+                        body.transactionID,
                     );
-                } else if (sigPair.ECDSASecp256k1 != null) {
+
                     signatures._set(
-                        PublicKey.fromBytesECDSA(sigPair.pubKeyPrefix),
-                        sigPair.ECDSASecp256k1,
+                        transactionId,
+                        SignaturePairMap._fromTransactionSigMap(sigMap),
                     );
                 }
             }
--- a/src/transaction/SignatureMap.js
+++ b/src/transaction/SignatureMap.js
@@ -21,6 +21,7 @@
 import NodeAccountIdSignatureMap from "./NodeAccountIdSignatureMap.js";
 import ObjectMap from "../ObjectMap.js";
 import AccountId from "../account/AccountId.js";
+import List from "./List.js";
 
 /**
  * @augments {ObjectMap<AccountId, NodeAccountIdSignatureMap>}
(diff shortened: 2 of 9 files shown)


src/transaction/NodeAccountIdSignatureMap.js
```diff
@@ -19,41 +19,60 @@
...
     }
+
+    /**
+     *
+     * @param {TransactionId} txId
+     * @param {import("../SignerSignature.js").PublicKey} publicKey
+     * @param {Uint8Array} signature
+     */
+    addSignature(txId, publicKey, signature) {
+        if (!this.get(txId)) {
```
ivaylogarnev-limechain, line 69: 1. You could store `this.get(txId)` in a constant and use it in both places to avoid redundant calls.

2. If you have time you can add an explanation of what the function does.

src/transaction/SignatureMap.js
```diff
@@ -37,17 +42,65 @@ export default class SignatureMap extends ObjectMap {
...
+     * @param {import("../SignerSignature.js").PublicKey} publicKey
+     * @param {Uint8Array} signature
+     * @returns {SignatureMap}
+     */
+    addSignature(nodeId, txId, publicKey, signature) {
+        if (!this.get(nodeId)) {
+            this._set(nodeId, new NodeAccountIdSignatureMap());
+        }
+
+        const nodeAccountIdSigdMap = this.get(nodeId);
```
ivaylogarnev-limechain, line 82: 1. You can put this variable in the beginning of function and use it in the if statement also.

2. If you have time you can add an explanation of what the function does.

src/transaction/SignatureMap.js
```diff
@@ -37,17 +42,65 @@ export default class SignatureMap extends ObjectMap {
     static _fromTransaction(transaction) {
```
ivaylogarnev-limechain, line 42: If you have time you can add an explanation of what the function does and maybe a little comments what the rowLength and columns represent in terms of signatures, nodes and etc.

test/unit/Transaction.js
```diff
@@ -491,48 +383,67 @@ describe("Transaction", function () {
...
         const signAndAddSignatures = (transaction, ...keys) => {
-            // Map through the keys to sign the transaction and add signatures
-            const signatures = keys.map((key) => {
-                const signature = key.signTransaction(transaction);
-                transaction.addSignature(key.publicKey, signature);
-                return signature;
-            });
+            for (const key of keys) {
+                const sigMap = createSigmap(transaction, key);
+                transaction.addSignature(key.publicKey, sigMap);
```
0xivanov, line 388: This looks like a changed API for me, I suggest to add an example. Also it seems like it's kinda hard to get the `sigmap` from the transaction.
Would it be possible to create a helper method to do that?
Something like `transaction.addSignature(key.publicKey, transaction.getSigmap());`

src/transaction/Transaction.js
```diff
@@ -868,9 +837,35 @@ export default class Transaction extends Executable {
...
+                    throw new Error(
+                        "Transaction ID or Node Account ID not found in the signed transaction",
+                    );
+                }
+
+                const transactionId =
+                    TransactionId._fromProtobuf(transactionID);
+                const nodeAccountId = AccountId._fromProtobuf(nodeAccountID);
+
+                const signature = signatureMap
```
ivaylogarnev-limechain, line 856: IMO: You can split this nesting into 3 different variables that way it will be more readable.

```
const accountSignatures / nodeSignatures = signatureMap.get(nodeAccountId);
const transactionSignatures = accountSignatures?.get(transactionId);
const signature = transactionSignatures?.get(publicKey);
```

test/unit/PrivateKey.js
```diff
@@ -1,121 +1,99 @@
...
+            sigMap.addSignature(
+                new AccountId(3),
+                txId,
+                privateKey.publicKey,
+                sig,
+            );
+        });
+        multisignatureTransaction.addSignature(privateKey.publicKey, sigMap);
+
+        const flatSigs = multisignatureTransaction
```
ivaylogarnev-limechain, line 94: Could  you please check if the signatures are the same as well.

test/unit/Transaction.js
```diff
@@ -317,167 +318,58 @@ describe("Transaction", function () {
...
-
-            // Validate the calls to _toProtobufSignature
-            sinon.assert.calledWith(
-                mockedPublicKey._toProtobufSignature,
-                mockedSignatures[0],
-            );
-            sinon.assert.calledWith(
-                mockedPublicKey._toProtobufSignature,
-                mockedSignatures[1],
+            expect(signatures.length).to.equal(
```
ivaylogarnev-limechain, line 358: Could you please check if the signatures are the same as well. 

*this is valid for all the signature related tests

src/transaction/SignatureMap.js
ivaylonikolov7: Wouldn't that confuse people? You will have to think about "oh, it's extracted in a variable but it also mutates the state... hm, would that change the variable value or would it change the class instance?"

I agree for the explanation.

src/transaction/NodeAccountIdSignatureMap.js
ivaylonikolov7: Same as comment below.

test/unit/Transaction.js
Svetoslavb04: The `transaction.getSignatures()` returns a `SignatureMap` with all the signatures.
`transaction.addSignature` accepts a public key and even if you pass the whole signature map, only the signatures for the provided public key will be added.

It is not worth creating a helper to get the signatures for a particular public key because that helper will have to iterate through the whole signature map.
The current approach saves one iteration by passing the whole signature map and adding a single public key.
The problem occurs if you want to add signatures for more than one public key.
If you do so, you must iterate n times where n is the number of public keys.

0xivanov: Ok, let's add an example to see how it's used from user perspective so I get see it clearly.

src/transaction/SignatureMap.js
ivaylogarnev-limechain: No, it wouldn’t be confusing. Having the variable at the beginning of the function would help anyone new to the code quickly understand what **this.get(nodeId)** or **this.get(txId)** represents.

0xivanov (approved):

agadzhalov (approved):
LGTM

PR #2605, opened 2024-10-24
fix: exports react native entrypoint for the sdk

KVNLS:
**Description**:

This PR fixes the SDK usage with the latest metro resolver.
For Ledger Live we had to do a patch to make it work with a Bare React Native project. https://github.com/LedgerHQ/ledger-live/blob/develop/patches/%40hashgraph__sdk%402.14.2.patch

See for reference:
- https://nodejs.org/docs/latest-v19.x/api/packages.html#community-conditions-definitions
- https://metrobundler.dev/docs/package-exports/#replacing-browser-and-react-native-fields

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


ivaylonikolov7:
we had to do some extra steps to use `lib` instead of `src`. We will make the necessary changes to fix this here: 

https://github.com/hiero-ledger/hiero-sdk-js/pull/3388/files

PR #2613, opened 2024-10-28
fix: incorrect protobuf body field used

kantorcodes:
**Description**:

This PR fixes `_fromProtobuf` inside of `TokenUpdateNftsTransaction` to use the correct `tokenUpdateNfts` field instead of `tokenUpdate`. Upon investigating https://github.com/hashgraph/hedera-wallet-connect/issues/265, we realized this property was not being referenced properly. 

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-wallet-connect/issues/265

**Notes for reviewer**:

You can confirm the error happens outside of this branch by running the following code: 

```
import Long from "long";
import { TokenUpdateNftsTransaction, Transaction } from "./exports.js";

const updateNftsTransaction = new TokenUpdateNftsTransaction()
    .setTokenId("0.0.123")
    .setSerialNumbers([Long.fromInt(1)])
    .setMetadata(new TextEncoder().encode("metadata"));

const parsedTx = Transaction.fromBytes(updateNftsTransaction.toBytes());

console.log("parsedTx: ", parsedTx._makeTransactionBody());
```

**Checklist**

### DIFF
--- a/src/token/TokenUpdateNftsTransaction.js
+++ b/src/token/TokenUpdateNftsTransaction.js
@@ -101,7 +101,7 @@ export default class TokenUpdateNftsTransaction extends Transaction {
         const body = bodies[0];
         const tokenUpdate =
             /** @type {HashgraphProto.proto.ITokenUpdateNftsTransactionBody} */ (
-                body.tokenUpdate
+                body.tokenUpdateNfts
             );
 
         return Transaction._fromProtobufTransactions(
--- a/test/unit/TokenUpdateNftsTransaction.js
+++ b/test/unit/TokenUpdateNftsTransaction.js
@@ -1,0 +1,40 @@
+import { TokenUpdateNftsTransaction } from "../../src/index.js";
+import TransactionId from "../../src/transaction/TransactionId.js";
+import AccountId from "../../src/account/AccountId.js";
+
+describe("TokenUpdateNftsTransaction", function () {
+    it("should create from a protobuf", function () {
+        const tokenId = "0.0.123";
+        const serials = [1, 2, 3];
+        const metadata = new Uint8Array([1, 2, 3]);
+
+        /**
+         * @type {HashgraphProto.proto.ITokenUpdateNftsTransactionBody}
+         */
+        const body = {
+            tokenUpdateNfts: {
+                token: { shardNum: 0, realmNum: 0, tokenNum: 123 },
+                serialNumbers: serials,
+                metadata: { value: metadata },
+            },
+        };
+
+        const transactions = [{ body }];
+        const signedTransactions = [{}];
+        const transactionIds = [TransactionId.generate(new AccountId(1))];
+        const nodeIds = [new AccountId(3)];
+        const bodies = [body];
+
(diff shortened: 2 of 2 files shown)


ivaylogarnev-limechain:
Hi @kantorcodes , thank you for submitting this PR! It looks great. 

Would you mind adding a unit test for this change in the `TokenUpdateTransaction.js` to satisfy the code coverage in the CI check?

0xivanov:
Nice catch! This should be the correct behaviour.

Looks like we were missing some unit tests for this class as a whole and we also have a `TokenNftsUpdateTransaction` class which should be deleted.
@ivaylogarnev-limechain do you think we can do this in a follow up issue?

0xivanov (dismissed):
LGTM

ivaylogarnev-limechain:
> Nice catch! This should be the correct behaviour.
> 
> Looks like we were missing some unit tests for this class as a whole and we also have a `TokenNftsUpdateTransaction` class which should be deleted. @ivaylogarnev-limechain do you think we can do this in a follow up issue?

For sure, that sounds like a better option.

ivaylogarnev-limechain (changes requested):
Could you just please sign the last commit with:
`git commit --amend -s -S`

agadzhalov:
Hi @kantorcodes thanks for your effort. Your commit is not verified, it's signed, but it's not verified which may cause us problems in the future, can you try to sign it again may be using `ammend`.  I can also see that your commits are not verified in other [repos](https://github.com/kantorcodes/hedera-improvement-proposal/commits/main/). 

Are you sure you've setup your GPG keys properly? You can look at this [page](https://docs.github.com/en/authentication/managing-commit-signature-verification/displaying-verification-statuses-for-all-of-your-commits) for more info ([Signing commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits)).

image

agadzhalov (changes requested):
commits must be verified

kantorcodes:
> Hi @kantorcodes thanks for your effort. Your commit is not verified, it's signed, but it's not verified which may cause us problems in the future, can you try to sign it again may be using `ammend`. I can also see that your commits are not verified in other [repos](https://github.com/kantorcodes/hedera-improvement-proposal/commits/main/).
> 
> Are you sure you've setup your GPG keys properly? You can look at this [page](https://docs.github.com/en/authentication/managing-commit-signature-verification/displaying-verification-statuses-for-all-of-your-commits) for more info ([Signing commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits)).
> 
> image

Thanks! I think I've got these signed properly now :)

ivaylogarnev-limechain (approved):
LGTM, thank you!

kantorcodes:
> commits must be verified

Thanks for taking a look. I think commits should be signed properly. I realized afterwards (and likely why this went unnoticed initially) that fromProtobuf had no test coverage and this seems to have blocked the merge, so I added another commit to add a basic test in as well.

ivaylonikolov7 (approved):

Issue #2632, opened 2024-11-01
Add option to automatically query for an `AddressBook` if a `Client` is initialized only with a `mirrorNetwork`


rwalworth:
To allow users to fully initialize a `Client` with only a mirror node address, there should be an option that allows for the initialization of the `Client`'s consensus node network automatically with an `AddressBookQuery` to the mirror node.

ivaylonikolov7:
Resolved via #2635

PR #2635, opened 2024-11-06
feat: Add method to `NodeClient` for Initialization with `MirrorNetwork` only

ivaylogarnev-limechain:
**Description**:

This PR adds a new method to the NodeClient class, allowing it to be initialized with only a MirrorNetwork address. After setting the MirrorNetwork, an AddressBookQuery is executed to retrieve the full consensus node network, which is then configured in the client. An example usage is also provided for clarity.

**Related issue(s)**:
#2632 

**Checklist**

### DIFF
--- a/src/client/NodeClient.js
+++ b/src/client/NodeClient.js
@@ -30,8 +30,8 @@ import * as mainnet from "./addressbooks/mainnet.js";
 import * as testnet from "./addressbooks/testnet.js";
 import * as previewnet from "./addressbooks/previewnet.js";
 import * as hex from "../encoding/hex.js";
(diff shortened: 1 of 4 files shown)


src/client/NodeClient.js
```diff
@@ -189,6 +191,28 @@ export default class NodeClient extends Client {
         return new NodeClient({ network: "testnet", ...props });
     }
 
+    /**
+     * @param {string} network
```
0xivanov, line 195: We can make this param `{string[] | string}` as it's for `setMirrorNetwork`

0xivanov (approved):
LGTM

ivaylonikolov7 (approved):
LGTM!

PR #2657, opened 2024-11-20
fix: _fromProtobuf functions where google primitive wrappers used

SvetBorislavov:
**Description**:
This pull request updates the `_fromProtobuf` function of the following classes:
- AccountUpdateTransaction
- ContractFunctionResult
- ContractUpdateTransaction
- FileUpdateTransaction
- NodeUpdateTransaction
- TokenTransfer
- TokenUpdateNftsTransaction
- TokenUpdateTransaction
- TopicUpdateTransaction

This change is required because the protobufs for the bodies (or the data) of these classes use the [primitive wrappers](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/wrappers.proto)
These wrappers are necessary when the property is nullable. Since the wrapper is a sub-message with a `value` it assumed that:
- if the sub-message is null, the value is null
- if the sub-message does not contain the value property, the value is null
- if the sub-message contains the value property, you should use that value

The problem with the current implementation is that it checks whether the wrapper's `value` is null, which is incorrect because the wrapper's `prototype` stores the property's default value.
Because the `prototype` has set the value to default one, if the protobuf, does not contain value (it is null), the instance still has the `value` property, which is not enumerable but it is accessible.

The correct check is to ensure that the instance has its own property `value` and not a derived one.

The current `_makeTransactionData` implementation hides the actual problem because it strips the value if it is not set at all, but the problem occurs when these values are explicitly set (correctly) to an object with a `value` property that can be undefined

### DIFF
--- a/src/account/AccountUpdateTransaction.js
+++ b/src/account/AccountUpdateTransaction.js
@@ -52,16 +52,16 @@ export default class AccountUpdateTransaction extends Transaction {
      * @param {object} props
      * @param {AccountId} [props.accountId]
      * @param {Key} [props.key]
-     * @param {boolean} [props.receiverSignatureRequired]
+     * @param {?boolean} [props.receiverSignatureRequired]
      * @param {AccountId} [props.proxyAccountId]
      * @param {Duration | Long | number} [props.autoRenewPeriod]
      * @param {Timestamp | Date} [props.expirationTime]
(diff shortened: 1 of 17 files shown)


0xivanov (approved):
LGTM

Issue #2329, opened 2024-06-10
Increase in returning invalid node account ID error when upgrading to 2.45.0


SimiHunjan:
### Description

This report is from the Kabila team. 

1. I believe since we updated to "@hashgraph/sdk": "2.45.0"
2. We were using 2.44.0 and before we never had these errors, this change was made 2 weeks ago
3. The payload is the same than any other transaction, we do not have logs of nodes or something else.
4. The exact flow is just generate a transaction in a DApp, sign within a wallet and execute in the DApp. It happens super randomly, you can execute 50 transaction and no one of them fails and then execute 3 in a row and these 3 fail. 
5. Browser & Browser + Backend.
    Browser Flow: DApp create the transaction --> Send to wallet --> Wallet Signs --> Return to DApp to execute.
    Browser + Backend: Transaction is being created in Backend --> returned to the FrontEnd --> FrontEnd signs --> Send back to the backend to execute.
In both flows happen.
6. The node custom list on Wallet is all the proxies except the 15 (Since 15 is always down, the DNS does not resolve) and from the DApp as any other DApp using wallet connect, you can only generate the transaction with one node, so we choose from the API whenever we setup the client one ramdom node that is available
7. Mainnet.

The issues are being reported using Kabila Wallet and Hashpack Wallet atm

### Steps to reproduce

See above

### Hedera network

mainnet

### Version

2.45.0

### Operating system

None

moshenskyDV:
<img width="548" alt="Screenshot 2024-06-11 at 13 23 52" src="https://github.com/hashgraph/hedera-sdk-js/assets/15670713/f92b6015-0585-452c-af23-fb9d232aaf40">
Could be related. Happens super rare for me (noticed this 2 times for last week).

teacoat:
+1 We have also been seeing elevated reports of this error, though from what I've seen it may not be SDK related

SimiHunjan:
Noting that SaucerSwap is using v2.30.0 and Kabila/HashPack are on v2.45.0 and both experienced an uptick of seeing this error over the last week or so.

AdrianKBL:
Do we have any idea what's causing this error? We keep seeing more and more often... getting daily reports from users.

agadzhalov:
@AdrianKBL we are still trying to reproduce it and communicating with our colleagues. Did you noticed any specific transaction fail more often? You mentioned that it happens randomly but I am wondering if there is any specific transaction that is executed very often and started to fail more?

[...]

PR #2661, opened 2024-11-25
feat: Added `InvalidNode` check in the Executable

ivaylogarnev-limechain:
**Description**:
This PR introduces a validation in the Executable class to check if the client executing the transaction and the transaction's node IDs differ.

**Related issue(s)**:
#2668
#2329 

**Notes for reviewer**:

**Checklist**


src/Executable.js
```diff
@@ -153,6 +162,19 @@ export default class Executable {
...
     }
 
+    /**
+     * Get the list of node account IDs set in the transaction. If no nodes are set, then null is returned.
+     *
+     * @returns {?string[]}
+     */
+    get transactionNodeAccountIds() {
+        if (this._transactionNodeIds.isEmpty) {
+            return null;
```
ivaylonikolov7, line 172: Might be better to have it return [] so you wouldn't have to worry about types like `undefined`, `null`, etc.

gsstoykov (commented):
I didn't see a check for the payment transactions created from the queries. Please verify it will work correctly there and let me know.

test/unit/AccountInfoMocking.js
```diff
@@ -565,7 +565,7 @@ describe("AccountInfoMocking", function () {
         expect(err).to.be.false;
     });
 
-    it("should re-create a transaction if sign on demand is enabled and a random node is chosen which is not in the current list", async function () {
+    xit("should re-create a transaction if sign on demand is enabled and a random node is chosen which is not in the current list", async function () {
```
ivaylonikolov7, line 568: I think we might need to add an eslint ignore here or otherwise it will say "unexpected skipped mocha test" on some actions.

src/Executable.js
```diff
@@ -595,6 +629,19 @@ export default class Executable {
                 );
             }
 
+            if (this.transactionNodeAccountIds?.length) {
```
ivaylonikolov7, line 632: if we have the empty array return this check wouldn't be needed and the code would be more "flat".

ivaylogarnev-limechain: Even if the array is empty, its length must still be checked because `isNodeAccountIdValid` will always evaluate to false. This is because the chosen node for the transaction cannot be found in an empty array.

test/unit/AccountInfoMocking.js
ivaylogarnev-limechain: The test needed to be updated first, which is why it wasn't ready for review earlier. It's done now, so you can check it again.

src/Executable.js
ivaylogarnev-limechain: All the other getters return `null` when empty, which is preferable because it eliminates the need to check the length when adding conditions or verifying existence.

src/Executable.js
```diff
@@ -166,6 +188,18 @@ export default class Executable {
         return this;
     }
 
+    /**
+     * Set the node account IDs of the transaction
+     *
+     * @param {AccountId[]} nodeIds
+     * @returns {this}
+     */
+    setTransactionNodeIds(nodeIds) {
```
0xivanov, line 197: I think this should not be exposed as an API

ivaylogarnev-limechain: You are correct; it’s better to limit this property to only this specific use case. Removed the getter/setters for this property.

ivaylogarnev-limechain:
> I didn't see a check for the payment transactions created from the queries. Please verify it will work correctly there and let me know.

Added the check for CostQuery as well.

ivaylogarnev-limechain:
> I didn't see a check for the payment transactions created from the queries. Please verify it will work correctly there and let me know.

Added the check for cost query as well.

src/Executable.js
```diff
@@ -595,6 +604,18 @@ export default class Executable {
                 );
             }
 
+            if (this.transactionNodeIds.length) {
+                const isNodeAccountIdValid = this.transactionNodeIds.includes(
+                    nodeAccountId.toString(),
+                );
+
+                if (!isNodeAccountIdValid) {
```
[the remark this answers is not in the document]
0xivanov: I am not sure if it's applicable, just suggesting.

0xivanov (dismissed):
LGTM

src/Executable.js
```diff
@@ -595,6 +604,18 @@ export default class Executable {
...
+                    throw new Error(
+                        `Attempting to execute a transaction against node ${nodeAccountId.toString()}, which is not present in the Client's node list. Please review and adjust your setup.`,
```
rwalworth, line 614: Would this still throw in the case of 1 invalid and 2 valid node account IDs being supplied? For example, if `{1, 2, 3}` node account IDs were supplied manually by the user and node account ID `1` was tried first, would this error then be thrown and execution stopped? I think we'd still want to try `2` and `3` in this case as they are still valid. Or is there processing somewhere that I missed that would make sure `1` never got selected?

ivaylogarnev-limechain: I initially considered a more restrictive approach, where an error is thrown if the client is not configured correctly, to prevent the user from making whatever he choices. However, your suggestion makes perfect sense, and I’ve updated the logic accordingly. 

Please note that with this update, we won't immediately throw a full error to the user. Instead, an error message will be logged to the console, and the process will continue with the next available node account ID. After 10 attempts, the user will be notified via a console log indicating an issue, and the error from the original flow will be thrown with the message: `"Max attempts of 10 were reached for the request, with the last error being:"``

An example of what happens when only invalid accounts are provided:
image


An example of what happens when a mix of invalid and one valid account is provided:
image


`* You can execute the newly added integration tests to cover all the different scenarios.`

rwalworth: I think this flow looks good!

rwalworth (approved):
LGTM

agadzhalov (dismissed):

agadzhalov (approved):
LGTM

PR #2697, opened 2024-12-05
fix: add long types declaration locally so we dont have to rely on remote declaration

ivaylonikolov7:
**Description**:
Adds `types/long.d.ts` due to the transpiler not being able to find the default `Long` type. We already use `allowSyntheticDefaultImports` property but if I try to import only the type of `long` using `typedef {import("long")}` it will transpile the code to `typedef {import("long")}.default` inside the `.d.ts` files in the SDK. This will show an error. I try to circumvent this error by importing the `Long` class declaration locally. I think it might have something to do with the way they export the Long class. Similar issues has been raised [here](https://github.com/dcodeIO/long.js/issues/118) and [here](https://github.com/dcodeIO/long.js/issues/109).

**Checklist**

### DIFF
--- a/packages/proto/src/proto/services/response_code.proto
+++ b/packages/proto/src/proto/services/response_code.proto
@@ -1596,33 +1596,19 @@ enum ResponseCodeEnum {
 
     /**
      * A scheduled transaction configured to wait for expiry to execute was given
-     * an expiry time not strictly after the time at which its creation reached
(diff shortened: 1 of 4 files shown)


ivaylogarnev-limechain (dismissed):
LGTM

agadzhalov (approved):

PR #2705, opened 2024-12-09
Update util.js

BugAnnihilator:
The `convertToNumber`function in the [utils](https://github.com/hashgraph/hedera-sdk-js/blob/main/src/util.js) file in the Hedera JavaScript SDK (`hedera-sdk-js`) improperly allows strings as valid input. According to JavaScript documentation, converting strings directly to numbers using `parseInt` is potentially unsafe and can lead to unexpected or improper behavior.

### Vulnerable Function :

```javascript
export function convertToNumber(variable) {
    requireNonNull(variable);
    if (
        isBigNumber(variable) ||
        isString(variable) ||
        isNumber(variable) ||
        isLong(variable)
    ) {
        const num = parseInt(variable);
        if (isNaN(num)) {
            throw new Error(FUNCTION_CONVERT_TO_NUMBER_PARSE_ERROR);
        } else {
            return num;
        }
    } else {
        throw new Error(FUNCTION_CONVERT_TO_NUMBER_ERROR);
    }
}
```

The `convertToNumber` function accepts strings as input due to this check:

```javascript
function isString(variable) {
    return isNonNull(variable) && typeof variable == "string";
}
```

This allows any non-null string to be passed to the convertToNumber function. Once passed, the string is processed as follows:

```javascript
const num = parseInt(variable);
```

image

## Example Impact

For example, if the input "0xFF" is passed:

```javascript
const variable = "0xFF";
console.log(convertToNumber(variable)); // Outputs: 255
```

While "0xFF" is a valid hexadecimal representation, its acceptance as input may result in logical errors when non-decimal strings are unintentionally interpreted as numbers.

## Recommendation for Fix

To mitigate the issue, it's recommended to remove the check for `isString(variable)` in the `convertToNumber` function, as it allows non-decimal strings (such as hexadecimal strings) to be passed through, leading to potential misinterpretations and errors.

Here’s the modified code:

```javascript
export function convertToNumber(variable) {
    requireNonNull(variable);
    if (
        isBigNumber(variable) ||
        isNumber(variable) ||
        isLong(variable)
    ) {
        const num = parseInt(variable);
        if (isNaN(num)) {
            throw new Error(FUNCTION_CONVERT_TO_NUMBER_PARSE_ERROR);
        } else {
            return num;
        }
    } else {
        throw new Error(FUNCTION_CONVERT_TO_NUMBER_ERROR);
    }
}

```

## Impact

As the SDK evolves and more functions are added in the future, the continued use of the `convertToNumber` function without proper validation could introduce compounding issues. Functions that rely on this conversion may inherit the same vulnerabilities, leading to inconsistent data processing and calculations. This can result in incorrect transaction amounts, unexpected behavior, and potential failures in complex workflows. As the SDK is integrated into more critical systems, such as financial platforms and smart contract interactions, the risk of severe errors increases. Therefore, relying on this unchecked function in future developments could create a cascading effect of unreliable behavior and security vulnerabilities, impacting the overall stability and trustworthiness of the SDK.

The use of the `convertToNumber` function with strings can cause significant issues in the future, especially when the SDK is used in critical operations such as transactions. Allowing strings like `"0xFF"` or any other arbitrary string to be passed to the function can lead to:

1. **Improper Amount Handling**: Since strings can be interpreted as numbers (e.g., `"0xFF"` becomes `255`), this can result in incorrect numeric conversions, leading to improper amounts being used in transactions or contract calls.
   
2. **Transaction Failures**: In cases where precise amounts are required, allowing strings to be passed unchecked may cause invalid or unexpected transaction amounts, leading to failed transactions, incorrect balances, or even loss of funds.

3 **Logic Errors in Smart Contracts**: If incorrect numeric values are passed, the logic of smart contracts or other business rules may be violated, causing unintended behavior or contract malfunctions.

4. **Loss of Trust and Reliability**: Allowing such invalidated inputs could undermine the trust and reliability of the SDK, especially in high-stakes applications where financial transactions are involved. The potential for critical errors is increased as the SDK is used in real-world environments.

Given the potential for significant disruptions, this issue should be addressed immediately to ensure that only valid numeric inputs are accepted and to prevent future operational issues.

Thanks,
Fahad Faisal

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -299,7 +299,6 @@ export function convertToBigNumber(variable) {
     requireNonNull(variable);
     if (
         isBigNumber(variable) ||
-        isString(variable) ||
         isNumber(variable) ||
         isLong(variable)
     ) {


BugAnnihilator:
Hello team, 

Any update on this request?

Thanks,
Fahad

ivaylonikolov7:
Closing due to suspicious activity

BugAnnihilator:
Man why?

BugAnnihilator:
I am a security researcher currently working on hedera BBp on hackerone.

Why is this flagged as a suspicious activity?

Issue #2717, opened 2024-12-12
`Bug`: Improper Verification of gRPC Communication via Nodes Without Certificates in NODE_CERT File


ivaylogarnev-limechain:
### Description

The Hedera SDK fails to properly verify **gRPC** communication for nodes whose certificates are not included in the `NODE_CERT.js` file. This issue prevents secure connections with such nodes, creating a challenge to fully supporting TLS communication..

### Steps to reproduce

1. Configure a transaction to target a node without its certificate listed in the NODE_CERT file _(e.g., Node 8 on testnet)_.
2. Execute any transaction or query against this node.
3. Observe the connection failure caused by the missing certificate.

### Expected behavior 

1. Dynamically retrieve the certificate during the TLS handshake as per the standard TLS protocol.
2. Validate the certificate against the hash provided in the address book.
3. Establish a secure connection without relying on hardcoded certificates in NODE_CERT.

### Hedera network

mainnet, testnet, previewnet

### Version

latest

### Operating system

macOS

PR #2718, opened 2024-12-13
chore: Enhance gRPC TLS by Dynamic Certificate Retrieval

ivaylogarnev-limechain:
**Description**:
This PR fixes an issue where the Hedera SDK fails to establish secure connections with nodes whose certificates are not included in the **NODE_CERT.js** file. It enhances the SDK to dynamically retrieve and validate certificates during the TLS handshake, ensuring secure communication with all nodes on mainnet, testnet, and previewnet.

**Related issue(s)**:
#2717 

**Checklist**


src/constants/ClientConstants.js
```diff
@@ -77,108 +78,112 @@ export const ALL_WEB_NETWORK_NODES = {
...
+    "3.77.94.254": "0.0.31",
+    "34.107.78.179": "0.0.31",
+    "34.86.186.151": "0.0.32",
+    "3.20.81.230": "0.0.32",
+    "18.136.65.22": "0.0.33",
+    "34.142.172.228": "0.0.33",
+    "34.16.139.248": "0.0.34",
+    "35.155.212.90": "0.0.34",
     // TESTNET
-    "34.94.106.61:50211": "0.0.3",
```
0xivanov, line 153: why remove the ports

src/channel/NodeChannel.js
```diff
@@ -18,73 +18,151 @@
...
+            socket.on("error", reject);
+        });
+    }
+
+    /**
+     * Initialize the gRPC client
+     * @returns {Promise<void>}
+     */
+    async _initializeClient() {
+        if (clientCache[this.address]) {
```
0xivanov, line 125: if i have 1 client executing 2 transactions, the flow will get through here 2 times, right?

src/channel/NodeChannel.js
```diff
@@ -18,73 +18,151 @@
...
+    }
+
+    /**
+     * Retrieve the server's certificate dynamically.
+     * @param {string} address
+     * @returns {Promise<string>}
+     */
+    async _retrieveCertificate(address) {
+        if (certificateCache[address]) {
+            return certificateCache[address];
```
0xivanov, line 86: if we have the cert in the cache, it will not return a Promise, but the String?

src/channel/NodeChannel.js
```diff
@@ -18,73 +18,151 @@
...
 
     /**
      * @override
      * @returns {void}
      */
     close() {
-        this._client.close();
+        if (this._client) {
+            this._client.close();
+            delete clientCache[this.address];
```
0xivanov, line 164: Why not clean the cert cache as well?

src/channel/NodeChannel.js
```diff
@@ -18,73 +18,151 @@
...
- * @property {?HashgraphProto.proto.SmartContractService} _smartContract
- * @property {?HashgraphProto.proto.FileService} _file
- * @property {?HashgraphProto.proto.FreezeService} _freeze
- * @property {?HashgraphProto.proto.ConsensusService} _consensus
- * @property {?HashgraphProto.proto.NetworkService} _network
- */
+/** @type {{ [key: string]: string }} */
+const certificateCache = {};
+/** @type {{ [key: string]: Client }} */
+const clientCache = {};
```
0xivanov, line 31: Do we need the `clientCache`? If we have the cert in cache, the only step we skip with the addition of the `clientCache` is `security = credentials.createSsl(certificate);` OR `security = credentials.createInsecure();`

0xivanov (commented):
GJ, just some clarification questions.

src/constants/ClientConstants.js
ivaylogarnev-limechain: The issue arises from this part of the code:

```
new GrpcServicesError(
   GrpcStatus.Timeout,
   ALL_NETWORK_IPS[this.nodeIp],
),
```

A client requested that we display which node the failure occurs on. Previously, this wasn't working for both TLS and non-TLS scenarios due to differing ports. Now, we are only comparing the IP addresses to address this.

src/channel/NodeChannel.js
ivaylogarnev-limechain: Yep and since we have caching in place, it will check if the client is already in the cache and return the cached client instead of going through the entire flow again.

src/channel/NodeChannel.js
ivaylogarnev-limechain: Yep, that's correct.

src/channel/NodeChannel.js
ivaylogarnev-limechain: That's a good question, I might have forgotten about it. I'll make sure to clean the cert as well.

src/channel/NodeChannel.js
0xivanov: so we skip the `credentials.createSsl(certificate)` OR `credentials.createInsecure()` which are not doing any call through the network and we are keeping X-number of `Client` objects in memory in return. 
 The client is quite a big object, in my opinion it's not worth to keep it in memory.

ivaylogarnev-limechain: This `client` is actually a channel client from `@grpc/grpc-js` rather than the SDK client. I will remove the certificateCache as its pointless in the situation

ivaylonikolov7 (approved):

PR #2797, opened 2025-01-17
chore(deps): bumped Typescript version

ivaylogarnev-limechain:


ivaylonikolov7:
Not a huge fan of these patch fixes with the `ts-ignore`. Can't we do it any other way?

ivaylogarnev-limechain:
> Not a huge fan of these patch fixes with the `ts-ignore`. Can't we do it any other way?


I tried using `JSDOC` for type assertions, but it didn't work as expected. Since these files are written in JavaScript and not TypeScript, using TypeScript type assertions isn't an option also. Ultimately, whether we use `ts-ignore` or specify the type with `JSDOC` explicitly, neither approach are correct. 

Another potential solution would be to refactor the code to meet the new type constraints. However, given that this is legacy code within the cryptography package, a refactor doesn’t seem like a practical solution at this stage. Furthermore, I’ve noticed that `ts-ignore` is already used in various parts within the project so I think its not a big deal.

ivaylonikolov7:
Okay, can you just update the package.json?

ivaylonikolov7 (approved):

Issue #2754, opened 2025-01-08
`TopicMessageQuery` fails with `Did not receive message from query`


0xivanov:
### Description

After enabling mirror node network for integration tests `TopicMessageQuery-should be executable` test starts failing.
Investigate and implement a solution.

### Steps to reproduce

-

### Version

latest

### Operating system

None

ivaylonikolov7:
resolved via #2799

PR #2799, opened 2025-01-20
test: fix TopicMessage related tests 

ivaylonikolov7:
**Description**:
- refactored topic message query for better readability
- fix topic tests 

**Related issue(s)**:
#2754

**Checklist**


0xivanov (commented):
`src/topic/TopicMessageQuery.js` still extends `Query`, right?
I think it should not, since it's not a HAPI query.

test/integration/TopicMessageQueryTest.js
```diff
@@ -12,57 +14,54 @@ describe("TopicMessageQuery", function () {
...
             .setSubmitKey(operatorKey)
             .setAutoRenewAccountId(operatorId)
             .execute(env.client);
 
-        const topic = (await response.getReceipt(env.client)).topicId;
+        const { topicId } = await response.getReceipt(env.client);
         const contents = "Hello from Hedera SDK JS";
-
-        await (
-            await new TopicMessageSubmitTransaction()
```
0xivanov, line 40: was the problem that we were doing the `TopicMessageQuery` after `TopicMessageSubmitTransaction`?

ivaylonikolov7: There were multiple issues with these tests. 
-The structure of them was broken. When we created the new topic then we needed to wait for mirror node to catch up and find the new topic. This wasn't implemented that way. 

-After that we needed to execute TopicMessageQuery before submitting transactions and wait a little more(around 2500ms). Otherwise the first topic message wasn't received by the SDK.

These bugs were breaking the `setLimit` functionality. Therefore breaking the tests in a lot of places too.

0xivanov: Ok, looks good overall. Can you see what would happen if we don't extend Query?

src/topic/TopicMessageQuery.js
```diff
@@ -120,19 +117,32 @@ export default class TopicMessageQuery extends Query {
...
          */
         this._completionHandler = () => {
-            if (this._logger) {
-                this._logger.info(
-                    `Subscription to topic ${
-                        this._topicId != null ? this._topicId.toString() : ""
-                    } complete`,
-                );
-            }
+            console.log(
```
0xivanov, line 120: why remove using the logger?

src/topic/TopicMessageQuery.js
```diff
@@ -497,7 +397,144 @@ export default class TopicMessageQuery extends Query {
                 throw new Error("(BUG) listener is unexpectedly not set");
             }
         } catch (error) {
+            console.log(Buffer.from(topicMessage.contents).toString("utf-8"));
```
0xivanov, line 400: probably leftover from debugging

examples/consensus-pub-sub.js
```diff
@@ -39,6 +40,9 @@ async function main() {
 
         console.log(`topicId = ${topicId.toString()}`);
 
+        // need to wait before the mirror node can see new topic
+        await setTimeout(2500);
```
0xivanov, line 44: I'd make it 5 secs, just in case, since users can run this on testnet and it's even slower there

examples/consensus-pub-sub.js
```diff
@@ -52,6 +56,10 @@ async function main() {
                     console.log(Buffer.from(message.contents).toString("utf8")),
             );
         const MESSAGES_LIMIT = 20;
+
+        // need to wait some time before we can receive messages
+        await setTimeout(2500);
```
0xivanov, line 61:
```suggestion
        await setTimeout(5000);
```

test/integration/TopicMessageIntegrationTest.js
```diff
@@ -7,39 +7,32 @@ import {
...
-        const handle = new TopicMessageQuery()
-            .setTopicId(topic)
+        const { topicId } = await (
+            await new TopicCreateTransaction()
+                .setAdminKey(operatorKey)
+                .execute(env.client)
+        ).getReceipt(env.client);
+
+        // wait for mirror node to see new topic id
+        await setTimeout(2500);
```
ivaylogarnev-limechain, line 32: Are we confident that all the `setTimeout` calls within the tests are provided with sufficient time? Wouldn't it be better to implement a retry mechanism similar to the one used in the class?

test/integration/TopicMessageQueryTest.js
```diff
@@ -2,8 +2,10 @@ import {
     TopicMessageQuery,
     TopicCreateTransaction,
     TopicMessageSubmitTransaction,
+    TopicDeleteTransaction,
 } from "../../src/exports.js";
 import IntegrationTestEnv from "./client/NodeIntegrationTestEnv.js";
+import { setTimeout } from "timers/promises";
```
ivaylogarnev-limechain, line 8: Could you please place this at the beginning of the file, alongside the other imports from libraries?

src/topic/TopicMessageQuery.js
ivaylonikolov7: The setter and getter were part of the Query class and decided that there's not much need for this logger in this class.

0xivanov (approved):
LGTM
cannot resolve the convos though

PR #2819, opened 2025-01-23
fix: Refactor the unhealthy logic inside the Executable class and int…

ivaylogarnev-limechain:
…roduces excludeCurrent method in the List class

### DIFF
--- a/src/Executable.js
+++ b/src/Executable.js
@@ -425,20 +425,6 @@ export default class Executable {
     }
 
     /**
-     * Advance the request to the next node
-     *
-     * FIXME: This method used to perform different code depending on if we're
-     * executing a query or transaction, but that is no longer the case
-     * and hence could be removed.
-     *
-     * @protected
-     * @returns {void}
-     */
-    _advanceRequest() {
-        this._nodeAccountIds.advance();
-    }
-
-    /**
      * Determine if we should continue the execution process, error, or finish.
      *
      * FIXME: This method should really be called something else. Initially it returned
@@ -654,27 +640,25 @@ export default class Executable {
             const channel = node.getChannel();
             const request = await this._makeRequestAsync();
 
-            // advance the internal index
-            // non-free queries and transactions map to more than 1 actual transaction and this will cause
-            // the next invocation of makeRequest to return the _next_ transaction
-            // FIXME: This is likely no longer relavent after we've transitioned to using our `List` type
-            // can be replaced with `this._nodeAccountIds.advance();`
-            this._advanceRequest();
-
             let response;
 
-            // If the node is unhealthy, wait for it to be healthy
-            // FIXME: This is wrong, we should skip to the next node, and only perform
-            // a request backoff after we've tried all nodes in the current list.
-            if (!node.isHealthy() && this._nodeAccountIds.length > 1) {
-                if (this._logger) {
-                    this._logger.debug(
-                        `[${logId}] node is not healthy, skipping waiting ${node.getRemainingTime()}`,
+            if (!node.isHealthy()) {
+                if (this._nodeAccountIds.length > 1) {
+                    this._nodeAccountIds.excludeCurrent();
+
+                    if (this._logger) {
+                        this._logger.debug(
+                            `[${logId}] node is not healthy, trying with the next node.`,
+                        );
+                    }
+                    continue;
+                } else {
+                    throw new Error(
+                        `Network connectivity issue: All nodes are unhealthy. Original node list: ${this._nodeAccountIds.list.join(
+                            ", ",
+                        )}`,
                     );
                 }
-
-                // We don't need to wait, we can proceed to the next attempt.
-                continue;
             }
 
             try {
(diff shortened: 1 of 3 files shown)


src/Executable.js
```diff
@@ -654,29 +640,31 @@ export default class Executable {
...
+                        `[${logId}] Node is not healthy, trying the next node.`,
                     );
                 }
 
-                // We don't need to wait, we can proceed to the next attempt.
+                this._nodeAccountIds.advance();
                 continue;
             }
 
+            this._nodeAccountIds.advance();
```
0xivanov, line 666: some notes:
1. Before always `advance`'d (before the `if (!node.isHealthy())`), now we have 2 `advance` calls, is this necessary.
2. Around line 740 in executable.js, we have `client._network.increaseBackoff(node);` which removes the node from the healthy nodes list and sets new value for `_readmitTime`. This logic works as expected, right?

test/unit/AccountInfoMocking.js
```diff
@@ -591,4 +591,83 @@ describe("AccountInfoMocking", function () {
             );
         }
     });
+
+    describe("Node health checks", function () {
+        beforeEach(async function () {
+            const responses1 = [
+                { response: ACCOUNT_INFO_QUERY_COST_RESPONSE },
+                { response: ACCOUNT_INFO_QUERY_RESPONSE },
+            ];
```
0xivanov, line 600: We have `it("should retry on UNAVAILABLE", async function () {` , what is the behaviour there with more than 1 node?

src/Executable.js
ivaylogarnev-limechain: 1. Both advances handle different scenarios in the retry/rotation logic.

    1.1 The first `advance()` inside the health check is specifically for when a node is unhealthy.
    
    1.2 The second `advance()` is part of the normal node rotation logic when trying different nodes for retries. 
    (_tested in `AccountInfoMocking.js` - should retry on `INTERNAL` and retry multiple nodes_)

2. Yes, it still reaches that point because this code accounts for the scenario where the node is initially healthy but throws an error after making a request. If the error is a `GrpcService` or `HttpError`, the `increaseBackOff()` method is triggered, as expected.

0xivanov: About the 1st one - this means we advance in both cases - can we have only 1 advance above `if (!node.isHealthy())` on line 644?

test/unit/AccountInfoMocking.js
ivaylogarnev-limechain: Test added that covers this case.

src/Executable.js
ivaylogarnev-limechain: The issue is that with a single advance before the health check, we're effectively skipping the health check of the first node and this would break the node health checking functionality as demonstrated by the failing test `"should skip unhealthy node and execute with healthy node"`.

0xivanov: I see.

test/unit/AccountInfoMocking.js
```diff
@@ -591,4 +591,107 @@ describe("AccountInfoMocking", function () {
...
+            const info = await new AccountInfoQuery()
+                .setNodeAccountIds([
+                    new AccountId(3),
+                    new AccountId(4),
+                    new AccountId(5),
+                ])
+                .setAccountId("0.0.3")
+                .execute(client);
+
+            expect(info.accountId.toString()).to.be.equal("0.0.10");
```
ivaylonikolov7, line 694: - Can we set node account ids to 3, 4 and 10. Because `0.0.10` looks like a magic number to me. I get it's because of the mocker but we set node ids to 3,4 and 5. We set 3 and 5 unhealthy and then we check if the accountId is 10.

ivaylonikolov7 (commented):
Did you manage to test the new functionality on testnet where we actually have unhealthy nodes? I know it's hard to make integration tests for this functionality because you can't make localnode unhealthy easily and then you are going to need a second account that will execute the transaction on?

ivaylogarnev-limechain:
> Did you manage to test the new functionality on testnet where we actually have unhealthy nodes? I know it's hard to make integration tests for this functionality because you can't make localnode unhealthy easily and then you are going to need a second account that will execute the transaction on?

This functionality is being tested in `AccountBalanceIntegrationTest.js` under the test case "_can connect to testnet with TLS._" 

Currently, node 5 on the testnet is down, yet the tests still pass. If you explicitly hardcode the nodeAccountId to only use node 5 with:

`.setNodeAccountIds([new AccountId(5)])`

it will throw the error:
"Network connectivity issue: All nodes are unhealthy. Original node list: 0.0.5

test/unit/AccountInfoMocking.js
0xivanov: Now looking at this, where does this account id 10 even come from ? :D

ivaylogarnev-limechain: I had the same thought but initially didn't want to refactor the old code for this test. However, I agree that it's quite confusing, so I went ahead and refactored the entire test file - `AccountInfoMocking.js` to use appropriate accountId values and updated the mocker response accordingly.

ivaylonikolov7 (dismissed):

ivaylonikolov7 (approved):

0xivanov (approved):
LGTM

PR #2823, opened 2025-01-24
feat: `TokenRejectFlow` refactor

0xivanov:
### Description
Currently, the `TokenRejectFlow` class and similar transaction flows are overly reliant on class hierarchies that treat flows as if they are transactions themselves. This design complicates the implementation and maintenance of transaction flows.  

This PR does not change the API, but only adds getters to allow modification to the individual transactions.

### DIFF
--- a/test/integration/TokenRejectFlowIntegrationTest.js
+++ b/test/integration/TokenRejectFlowIntegrationTest.js
@@ -202,6 +202,157 @@ describe("TokenRejectIntegrationTest", function () {
         }
     });
 
+    it.only("can execute via modifying individual transactions", async function () {
+        env = await IntegrationTestEnv.new();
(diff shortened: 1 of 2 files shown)


agadzhalov (dismissed):

test/integration/TokenRejectFlowIntegrationTest.js
```diff
@@ -202,6 +202,157 @@ describe("TokenRejectIntegrationTest", function () {
         }
     });
 
+    it.only("can execute via modifying individual transactions", async function () {
```
0xivanov, line 205: remove .only

ivaylogarnev-limechain (approved):
LGTM, just remove the `.only` from tests

ivaylogarnev-limechain (dismissed):

PR #2834, opened 2025-01-31
feat: account create transaction with key derived alias

venilinvasilev:
Description:
This PR extends the AccountCreateTransaction with the following APIs:

- **setECDSAKeyWithAlias(ECDSAKey)** - Sets ECDSA private key, derives and sets it's EVM address in the background.
- **setKeyWithAlias(Key, ECDSAKey)** - Allows for setting the account key and a separate ECDSA private key that the EVM address should be derived from. A user must sign the transaction with both keys for this flow to be successful.
- **setKeyWithoutAlias(Key)** - Explicitly calls out that the alias is not set.

### 🚨 Deprecation Notice:
- **`setKey(Key)`** is now **deprecated** in favor of `setKeyWithoutAlias`.

Related issue(s):
https://github.com/hiero-ledger/hiero-sdk-js/issues/2795

## ✅ Checklist:
- [x] **Extended the APIs** of `AccountCreateTransaction`
- [x] **Deprecated** `setKey(Key)` API of `AccountCreateTransaction`
- [x] **Tested** (unit, integration, etc.)
- [x] **Replaced deprecated API (`setKey(Key)`)** occurences with the new relevant one (`setKeyWithoutAlias(Key)`)
- [x]  **Add examples**

### DIFF
--- a/examples/create-account-with-alias.js
+++ b/examples/create-account-with-alias.js
@@ -4,7 +4,6 @@ import {
     Client,
     Hbar,
     AccountInfoQuery,
-    TransactionReceiptQuery,
     AccountCreateTransaction,
 } from "@hashgraph/sdk";
 
@@ -74,39 +73,26 @@ async function main() {
          *
          * Use the `AccountCreateTransaction`
          *   - Populate `setAlias(evmAddress)` field with the Ethereum public address
-         */
-        const accountCreateTx = new AccountCreateTransaction()
-            .setInitialBalance(Hbar.fromTinybars(100))
-            .setKeyWithoutAlias(operatorKey)
-            .setAlias(evmAddress)
-            .freezeWith(client);
-
-        /**
-         *
-         * Step 5
-         *
          * Sign the `AccountCreateTransaction` transaction with both the new private key and key paying for the transaction fee
-         */
-        const accountCreateTxSign = await accountCreateTx.sign(privateKey);
-        const accountCreateTxResponse =
-            await accountCreateTxSign.execute(client);
-
-        /**
-         *
-         * Step 6
-         *
          * Get the account ID of the newly created account
          */
-        const receipt = await new TransactionReceiptQuery()
-            .setTransactionId(accountCreateTxResponse.transactionId)
-            .execute(client);
+        const receipt = await (
+            await (
+                await new AccountCreateTransaction()
+                    .setInitialBalance(Hbar.fromTinybars(100))
+                    .setKeyWithoutAlias(operatorKey)
+                    .setAlias(evmAddress)
+                    .freezeWith(client)
+                    .sign(privateKey)
+            ).execute(client)
+        ).getReceipt(client);
 
         const newAccountId = receipt.accountId.toString();
         console.log(`Account ID of the newly created account: ${newAccountId}`);
 
         /**
          *
-         * Step 7
+         * Step 5
          *
          * Get the `AccountInfo` and show that the account has contractAccountId
          */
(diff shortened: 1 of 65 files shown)


src/account/AccountCreateTransaction.js
```diff
@@ -282,6 +283,7 @@ export default class AccountCreateTransaction extends Transaction {
      * If `receiverSignatureRequired` is true, then the key must also sign
      * any transfer into the account.
      *
+     * @deprecated Use `setKeyWithAlias` or `setKeyWithoutAlias` instead.
```
0xivanov, line 286:
```suggestion
     * @deprecated Use `setKeyWithoutAlias` instead.
```

0xivanov: Since `setKeyWithoutAlias` is the direct replacement.

src/account/AccountCreateTransaction.js
```diff
@@ -292,6 +294,73 @@ export default class AccountCreateTransaction extends Transaction {
...
+     * @param {PrivateKey} aliasKey - The ECDSA private key used to derive the EVM address.
+     * @returns {this}
+     */
+
+    /**
+     * @param {PrivateKey | Key} key - The primary signing key (ECDSAKey or another Key type).
+     * @param {PrivateKey} [aliasKey] - Optional. The ECDSA private key used to derive the EVM address.
+     * @returns {this}
+     */
+    setKeyWithAlias(key, aliasKey) {
```
0xivanov, line 318: This single multi-purpose function is confusing. If JS does not support method overloading I suggest creating 2 methods : `setECDSAKeyWithAlias(PrivateKey)` and `setKeyWithAlias(Key, PrivateKey)`
This is what I've done in Golang.

@ivaylonikolov7 What do you think?

examples/create-account-without-alias.js
```diff
@@ -0,0 +1,115 @@
...
+
+import dotenv from "dotenv";
+
+dotenv.config();
+
+/*
+Reference: [Streamline key and alias specifications for AccountCreateTransaction #2795](https://github.com/hiero-ledger/hiero-sdk-js/issues/2795)
+*/
+
+async function main() {
```
0xivanov, line 19: You can remove the newly added examples and modify `create-account-with-alias.js` to use the 2 new methods. You would need to create one more account in order to showcase both.

src/account/AccountCreateTransaction.js
venilinvasilev: @0xivanov [JS Doc supports method overloading](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0-beta/#@overload-support-in-jsdoc) and the method overloads are defined above.
 The method overloading definition requires it to have this multi purpose function statement at the end, however whenever I pass the required params for one of the overloads the type signature and description of the function is infered correctly in the IDE as I see. WDYT?

0xivanov: Yeah, this is up to discussion. I am looking from a user's perspective might be confusing. 
Anyways, do we need to have 3 overloads? I think we need only `(PrivateKey)` and `(Key, PrivateKey)`

examples/create-account-with-alias.js
```diff
@@ -123,6 +123,187 @@ async function main() {
...
+        const receipt = await new TransactionReceiptQuery()
+            .setTransactionId(accountCreateTxResponse.transactionId)
+            .execute(client);
+
+        const newAccountId = receipt.accountId.toString();
+        console.log(`Account ID of the newly created account: ${newAccountId}`);
+
+        /**
+         *
+         * Step 4
```
ivaylogarnev-limechain, line 177: Just a small detail: This is step 5.

examples/create-account-with-alias.js
```diff
@@ -123,6 +123,187 @@ async function main() {
...
+            console.log(
+                `The newly created account has an EVM alias ${accountInfo.contractAccountId} and it ${matchesOrNot} the alias from the ECDSA private account key`,
+            );
+        } else {
+            console.log(`The new account doesn't have an EVM alias`);
+        }
+    } catch (error) {
+        console.error(error);
+    }
+
```
ivaylogarnev-limechain, line 213: You can insert a terminal log separator between the two examples to prevent them from mixing together in the terminal.

examples/create-account-with-alias.js
ivaylogarnev-limechain, line 1: Overall, the example looks good, but I don't think we should check if the public key matches or if we have a contractId and the other conditions. The purpose here is to demonstrate the functionality, and these checks should already be covered in the tests. WDYT? @ivaylonikolov7 @0xivanov

src/account/AccountCreateTransaction.js
venilinvasilev: Agree with concerns, also as we have it this way in Golang, lets keep it consistent.

examples/create-account-with-alias.js
venilinvasilev: Good point.
I saw these kind of checks in the other examples, but I agree that examples should be declarative - not containing if checks or so.
If something is wrong it should be caught by the tests.
I would prefer removing these.

0xivanov: I think it's good to have these, but we would simplify it. For example in go :
```go
	// Verify account is created with the provided EVM address
	fmt.Println(info.ContractAccountID == ecdsaPrivateKey.PublicKey().ToEvmAddress())
	// Verify the account Id is the same from the create account transaction
	fmt.Println(info.AccountID.String() == newAccountId.String())
	// Verify the account key is the same as the operator key
	fmt.Println(info.Key.String() == operatorKey.PublicKey().String())
```

examples/create-account-with-alias.js
```diff
@@ -123,6 +123,194 @@ async function main() {
...
+         * Step 4
+         *
+         * Get the account ID of the newly created account
+         */
+        const receipt = await new TransactionReceiptQuery()
+            .setTransactionId(accountCreateTxResponse.transactionId)
+            .execute(client);
+
+        const newAccountId = receipt.accountId.toString();
+        console.log(`Account ID of the newly created account: ${newAccountId}`);
```
0xivanov, line 176: you can chain steps 2, 3 and 4. Making it something like this: 
```js
const newAccountId = new AccountCreateTransaction().set(...).freezeWith(...).execute(...).getReceipt(...).accountId.
```

0xivanov: same it other places

ivaylogarnev-limechain (approved):
LGTM

0xivanov (dismissed):

ivaylonikolov7 (approved):

Issue #2728, opened 2024-12-18
Add validation for creating PublicKey from bytes


0xivanov:
### Problem

A user may create invalid ECDSA or ED25519 Public Key from bytes.

### Solution

Add validation for the above mentioned flows.

ivaylonikolov7:
Closing this as it might not be feasible because of the way the algorithm works.

PR #2841, opened 2025-02-03
feat: add validation for public key

ivaylonikolov7:
**Description**:
Add validation to `PublicKey`s by installing a new NPM package that checks if the public key is on the curve.

**Related issue(s)**:

#2728

**Notes for reviewer**:

**Checklist**

- [x] Tested (unit, integration, etc.) - N/A

### DIFF
--- a/packages/cryptography/src/EcdsaPublicKey.js
+++ b/packages/cryptography/src/EcdsaPublicKey.js
@@ -4,6 +4,7 @@ import { arrayEqual } from "./util/array.js";
 import * as hex from "./encoding/hex.js";
 import * as ecdsa from "./primitive/ecdsa.js";
 import { keccak256 } from "./primitive/keccak.js";
+import { secp256k1 } from "@noble/curves/secp256k1";
 import elliptic from "elliptic";
 const ec = new elliptic.ec("secp256k1");
 
@@ -92,7 +93,8 @@ export default class EcdsaPublicKey extends Key {
                 `cannot decode ECDSA private key data from DER format`,
             );
         }
-        return new EcdsaPublicKey(ecdsaPublicKeyBytes);
+
+        return EcdsaPublicKey.fromBytesRaw(ecdsaPublicKeyBytes);
     }
 
     /**
@@ -106,7 +108,36 @@ export default class EcdsaPublicKey extends Key {
             );
         }
 
-        return new EcdsaPublicKey(data);
+        if (this.isValid(data)) {
+            return new EcdsaPublicKey(data);
+        } else {
+            throw new BadKeyError("invalid public key");
+        }
+    }
+
+    /**
+     * @param {Uint8Array} data
+     * @returns {boolean}
+     */
+    static isValid(data) {
+        const EMPTY_BUFFER_33_BYTES = Buffer.from(new Uint8Array(33).fill(0));
+        const EMPTY_BUFFER_65_BYTES = Buffer.from(new Uint8Array(65).fill(0));
+        const bufferData = Buffer.from(data);
+
+        // SEE HIP-540
+        if (
+            EMPTY_BUFFER_33_BYTES.equals(bufferData) ||
+            EMPTY_BUFFER_65_BYTES.equals(bufferData)
+        ) {
+            return true;
+        }
+
+        try {
+            secp256k1.ProjectivePoint.fromHex(data);
+            return true;
+        } catch (error) {
+            return false;
+        }
     }
 
     /**
--- a/packages/cryptography/src/EcdsaPublicKey.js
+++ b/packages/cryptography/src/EcdsaPublicKey.js
@@ -98,8 +98,10 @@ export default class EcdsaPublicKey extends Key {
     }
 
     /**
-     * @param {Uint8Array} data
+     * Creates an ECDSA public key from raw bytes after validating the input
+     * @param {Uint8Array} data - Raw byte data for the public key
      * @returns {EcdsaPublicKey}
+     * @throws {BadKeyError} If the key is invalid or has an incorrect length
      */
     static fromBytesRaw(data) {
         if (data.length != 33) {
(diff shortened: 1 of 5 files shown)


packages/cryptography/src/EcdsaPublicKey.js
```diff
@@ -106,7 +108,36 @@ export default class EcdsaPublicKey extends Key {
...
+        }
+    }
+
+    /**
+     * @param {Uint8Array} data
+     * @returns {boolean}
+     */
+    static isValid(data) {
+        const EMPTY_BUFFER_33_BYTES = Buffer.from(new Uint8Array(33).fill(0));
+        const EMPTY_BUFFER_65_BYTES = Buffer.from(new Uint8Array(65).fill(0));
```
gsstoykov, line 124: Do you need those buffers in memory to do the following checks?

packages/cryptography/src/Ed25519PublicKey.js
```diff
@@ -106,6 +111,25 @@ export default class Ed25519PublicKey extends Key {
...
+    /**
+     * @param {Uint8Array} data
+     * @returns {boolean}
+     */
+    static isValid(data) {
+        if (!data) return false;
+
+        const EMPTY_BUFFER = Buffer.from(new Uint8Array(33).fill(0));
+        const bufferData = Buffer.from(data);
+        if (bufferData.equals(EMPTY_BUFFER)) return true;
```
gsstoykov, line 123: Seem like some redundant operations to me.

packages/cryptography/src/Ed25519PublicKey.js
```diff
@@ -106,6 +111,25 @@ export default class Ed25519PublicKey extends Key {
...
+
+        try {
+            ed25519.ExtendedPoint.fromHex(data);
+            return true;
+        } catch {
+            return false;
```
gsstoykov, line 129: If an error is caught wouldn't it make sense to propagate it.

packages/cryptography/src/EcdsaPublicKey.js
```diff
@@ -4,6 +4,7 @@ import { arrayEqual } from "./util/array.js";
 import * as hex from "./encoding/hex.js";
 import * as ecdsa from "./primitive/ecdsa.js";
 import { keccak256 } from "./primitive/keccak.js";
+import { secp256k1 } from "@noble/curves/secp256k1";
```
gsstoykov, line 7: Don't libs we currently have imported support the point checks :( ?

ivaylonikolov7: nope :(

test/unit/PublicKey.js
```diff
@@ -106,4 +106,44 @@ describe("PublicKey", function () {
...
+
+        expect(() => {
+            PublicKey.fromBytesED25519(invalidPrivateKeyBytes);
+        }).to.throw;
+    });
+
+    it("should throw error when creating ECDSA key from invalid bytes", function () {
+        // Test cases with invalid byte
+        const invalidPrivateKeyBytes = Uint8Array.from([
+            0x00, 0xca, 0x35, 0x4b, 0x7c, 0xf4, 0x87, 0xd1, 0xbc, 0x43, 0x5a,
```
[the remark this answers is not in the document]
ivaylonikolov7: I copied the invalid bytes from the Java SDK. I think the first byte checks if the key is compressed or not. Therefore 0x00 making it invalid.

packages/cryptography/src/EcdsaPublicKey.js
ivaylonikolov7: Could you elaborate a little more?

packages/cryptography/src/Ed25519PublicKey.js
ivaylonikolov7: I think considering the method's name is `isValidKey` you would expect it to return `true` or `false`.

packages/cryptography/src/Ed25519PublicKey.js
ivaylonikolov7: How so?

test/unit/PublicKey.js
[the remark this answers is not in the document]
gsstoykov: It is really a question of convention:
`048ae371ee33535a3fa40e31c20485cd8bf869c151184c56cae2775b0156eb3a4b29421847d57caf98b271ab954a42bfb1bed565a9c238b2f7b30274cbd793a163` - also a valid key.
The bad thing about conventions around this topic is that many would trim the 0x02, 0x03 prefix or the 0x04 prefix. So to do a fully invalid key check i would go with bytes which I know for sure a wrong like `1+` the maximum value of the curve which would be N.

packages/cryptography/src/EcdsaPublicKey.js
```diff
@@ -106,7 +108,36 @@ export default class EcdsaPublicKey extends Key {
...
+        const EMPTY_BUFFER_33_BYTES = Buffer.from(new Uint8Array(33).fill(0));
+        const EMPTY_BUFFER_65_BYTES = Buffer.from(new Uint8Array(65).fill(0));
+        const bufferData = Buffer.from(data);
+
+        // SEE HIP-540
+        if (
+            EMPTY_BUFFER_33_BYTES.equals(bufferData) ||
+            EMPTY_BUFFER_65_BYTES.equals(bufferData)
+        ) {
+            return true;
```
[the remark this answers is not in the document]
ivaylonikolov7: Yeah because of HIP-540 the key 0x0000..00 is supposed to be valid.

packages/cryptography/src/Ed25519PublicKey.js
gsstoykov: if (Buffer.from(data).equals(Buffer.alloc(33))) return true;

packages/cryptography/src/EcdsaPublicKey.js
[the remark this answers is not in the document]
gsstoykov: Oh, forgot about that one :D

packages/cryptography/src/EcdsaPublicKey.js
gsstoykov: I mean do you need to allocate those buffers and save them in variables to do the checks? Isn't there some sort of faster check for this without having EMPTY_BUFFER_33_BYTES, EMPTY_BUFFER_65_BYTES

packages/cryptography/src/Ed25519PublicKey.js
gsstoykov: throw new BadKeyError("invalid public key"); is thrown in another layer. The message of the error for the key parse would be useful there imho. I know you've defined it so but that error could get up there somehow

packages/cryptography/src/EcdsaPublicKey.js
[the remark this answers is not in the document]
gsstoykov: could resolve this

packages/cryptography/src/EcdsaPublicKey.js
gsstoykov: Could resolve this as well

packages/cryptography/src/Ed25519PublicKey.js
```diff
@@ -106,6 +111,25 @@ export default class Ed25519PublicKey extends Key {
         return Ed25519PublicKey.fromBytes(hex.decode(text));
     }
 
+    /**
+     * @param {Uint8Array} data
+     * @returns {boolean}
+     */
+    static isValid(data) {
+        if (!data) return false;
```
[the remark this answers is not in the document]
gsstoykov: Also if you have this here why not have it in the other public key class as well?

test/unit/PublicKey.js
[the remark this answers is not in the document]
gsstoykov: This way you can validate the point verification function and will be really useful.

packages/cryptography/src/Ed25519PublicKey.js
0xivanov: Since this method is created in this PR, we can ditch it and do the validation in the `fromBytesRaw` func. There we can `throw new BadKeyError("invalid public key: ", e);`, giving more info to the user as Georgi said.

packages/cryptography/src/EcdsaPublicKey.js
ivaylonikolov7: I can use the `Buffer.alloc(33)` without allocating it to a `const` I am not sure if it's descriptive enough for the next dev who's going to read it tho.

gsstoykov: I will leave it up to you it seemed better to me.

PR #2849, opened 2025-02-09
feat(HIP-991): Permissionless revenue generating topics

venilinvasilev:
Related Issues:
https://github.com/hiero-ledger/hiero-sdk-js/issues/2551


src/topic/TopicMessageSubmitTransaction.js
```diff
@@ -93,6 +98,8 @@ export default class TopicMessageSubmitTransaction extends Transaction {
          */
         this._chunkSize = CHUNK_SIZE;
 
+        this._customFees = null;
```
0xivanov, line 101: I wonder if we can remove this and use customFeeLimit from `Transaction.js`.

src/topic/TopicMessageSubmitTransaction.js
```diff
@@ -126,6 +133,9 @@ export default class TopicMessageSubmitTransaction extends Transaction {
             /** @type {HashgraphProto.proto.IConsensusSubmitMessageTransactionBody} */ (
                 body.consensusSubmitMessage
             );
+        if (body.maxCustomFees != null) {
+            body.maxCustomFees[0].accountId;
+        }
```
0xivanov, line 138: I wonder if we can remove this.

examples/create-topic-with-revenue.js
```diff
@@ -0,0 +1,317 @@
...
+        );
+    }
+    const operatorId = AccountId.fromString(process.env.OPERATOR_ID);
+    const operatorKey = PrivateKey.fromStringDer(process.env.OPERATOR_KEY);
+
+    const nodes = {
+        "127.0.0.1:50211": new AccountId(3),
+    };
+
+    const client = Client.forNetwork(nodes).setOperator(
```
0xivanov, line 35: close the client in `finally` block

examples/create-topic-with-revenue.js
```diff
@@ -0,0 +1,319 @@
...
+
+        const customFee = new CustomFixedFee()
+            .setAmount(new Hbar(1).toTinybars())
+            .setFeeCollectorAccountId(operatorId);
+
+        const { topicId } = await (
+            await new TopicCreateTransaction()
+                .setAdminKey(operatorKey)
+                .setFeeScheduleKey(operatorKey)
+                .setMaxTransactionFee(100)
```
0xivanov, line 74: Remove this setter in example and integration tests.
Add this value in the `TopicCreateTransaction` itself. See how it's done in `TransferTransaction`. 25 Hbars should be sufficient:
```js
this._defaultMaxTransactionFee = new Hbar(25);
```

0xivanov: Also, bring back the balance option for the test env with 1000 Hbars, should be enough.

test/integration/TopicCreateIntegrationTest.js
```diff
@@ -67,6 +86,744 @@ describe("TopicCreate", function () {
...
+            let customFeeLimit = new CustomFeeLimit()
+                .setAccountId(payerAccount.accountId)
+                .setFees([
+                    new CustomFixedFee()
+                        .setAmount(1)
+                        .setDenominatingTokenId(new TokenId(0)),
+                ]);
+
+            // Submit a message to the revenue generating topic with custom fee limit
+            env.client.setOperator(
```
0xivanov, line 784: You have shared env for all tests in this suite, changing the operator in one test will break other. I suggest chaning the `before` statement to `beforeEach`

test/unit/TopicCreateTransaction.js
```diff
@@ -0,0 +1,173 @@
...
+            topicCreateTransaction.addFeeExemptKey(feeExemptKeyToBeAdded);
+
+            [feeExemptKeyToBeAdded].forEach((feeExemptKey, index) => {
+                expect(
+                    topicCreateTransaction
+                        .getFeeExemptKeys()
+                        // eslint-disable-next-line no-unexpected-multiline
+                        [index].toString(),
+                ).to.eql(feeExemptKey.toString());
+            });
```
ivaylogarnev-limechain, line 54: Is there a specific reason to iterate over a loop when there's only a single element, or would it be sufficient to simply do something like this?

```
    expect(topicCreateTransaction.getFeeExemptKeys()[0].toString())
        .to.eql(feeExemptKey.toString());
```

_*the same is valid for `TopicUpdateTransaction.js`_

src/topic/TopicCreateTransaction.js
```diff
@@ -238,6 +288,71 @@ export default class TopicCreateTransaction extends Transaction {
...
+    }
+
+    /**
+     * Adds a key that will be exempt from paying fees.
+     * @param {Key} key
+     * @returns {this}
+     */
+    addFeeExemptKey(key) {
+        this._requireNotFrozen();
+
```
ivaylogarnev-limechain, line 338: Would you mind removing the empty line, as well as in the other functions where it appears? Since there isn’t one in the previous methods, just to be consistent.

examples/create-topic-with-revenue.js
```diff
@@ -0,0 +1,319 @@
...
+        );
+    }
+    const operatorId = AccountId.fromString(process.env.OPERATOR_ID);
+    const operatorKey = PrivateKey.fromStringDer(process.env.OPERATOR_KEY);
+
+    const nodes = {
+        "127.0.0.1:50211": new AccountId(3),
+    };
+
+    const client = Client.forNetwork(nodes).setOperator(
```
ivaylonikolov7, line 35: `const client = Client.forName(process.env.HEDERA_NETWORK);` so the users can use it easier for other networks too. This way you won't need setting `const nodes = {...` too.

examples/create-topic-with-revenue.js
```diff
@@ -0,0 +1,319 @@
...
+        /*
+         * Step 11:
+         * Submit another message to that topic, paid with bob, without specifying max custom fee amount.
+         */
+
+        const bobBalanceBefore = await new AccountBalanceQuery()
+            .setAccountId(bobAccountId)
+            .execute(client);
+
+        client.setOperator(bobAccountId, bobKey);
```
ivaylonikolov7, line 286: Is this pattern used somewhere else? I think the common usage to change the sender of the transaction is to freeze the transaction and then sign it with new opertor.

src/topic/TopicInfo.js
```diff
@@ -41,8 +42,11 @@ export default class TopicInfo {
      * @param {?Timestamp} props.expirationTime
      * @param {?Key} props.adminKey
      * @param {?Key} props.submitKey
+     * @param {?Key} props.feeScheduleKey
+     * @param {?Key[]} props.feeExemptKeys
```
ivaylonikolov7, line 46: i think `{Key[]}` should be enough here?

src/transaction/CustomFeeLimit.js
```diff
@@ -0,0 +1,121 @@
...
+ * @typedef {import("@hashgraph/proto").proto.IFixedFee} HashgraphProto.proto.IFixedFee
+ * @typedef {import("@hashgraph/proto").proto.ICustomFeeLimit} HashgraphProto.proto.ICustomFeeLimit
+ */
+
+export default class CustomFeeLimit {
+    /**
+     *
+     * @param {object} props
+     * @param {?AccountId | string} [props.accountId]
+     * @param {?CustomFixedFee[]} [props.fees]
```
ivaylonikolov7, line 15: same as the previous comment. I think `CustomFixrdFee[]`without the `?` should be enough

test/integration/utils/Fixtures.js
```diff
@@ -0,0 +1,67 @@
...
+ */
+export const createAccount = async (client, propOverwrites = {}) => {
+    const privateKey = propOverwrites.key
+        ? propOverwrites.key
+        : PrivateKey.generateECDSA();
+
+    const accountCreateTransaction = new AccountCreateTransaction({
+        initialBalance: new Hbar(1),
+        key: privateKey,
+        ...propOverwrites,
```
ivaylonikolov7, line 59: Might need some kind of `setMaxAutomaticTokenAssociations` too.

ivaylonikolov7 (commented):
We might need to incorporate that file `fixtures.js` in all our tests. it doesn't look good if we use it only one integration test.

test/integration/utils/Fixtures.js
ivaylonikolov7: Nevermind I saw the ...`propOverwrites`

test/unit/TopicCreateTransaction.js
```diff
@@ -0,0 +1,173 @@
...
+        });
+
+        it("should add fee exempt key to empty list", function () {
+            const topicCreateTransaction = new TopicCreateTransaction();
+
+            const feeExemptKeyToBeAdded = PrivateKey.generateECDSA();
+
+            topicCreateTransaction.addFeeExemptKey(feeExemptKeyToBeAdded);
+
```
ivaylonikolov7, line 46: Is there a reason we create the topicCreateTransaction and then add `setter` functionality. We could just do 

```
const topicCreateTransaction = new TopicCreateTransaction().addFeeExemptKey(...)```

File: examples/create-topic-with-revenue.js (hunk 9 above)
Comment (0xivanov) [reply examples/create-topic-with-revenue.js:286]: We need to make it a payer, this is the easiest way.

Comment (0xivanov) [reply examples/create-topic-with-revenue.js:286]: Only signing will not make the account a payer

Comment (ivaylonikolov7) [reply examples/create-topic-with-revenue.js:286]: I think I had a simillar issue. We set the transaction id and freeze/sign.

**Comment by 0xivanov** [comment]
> We might need to incorporate that file `fixtures.js` in all our tests. it doesn't look good if we use it only one integration test.

Let's save that for other PR.

File: examples/create-topic-with-revenue.js (hunk 9 above)
Comment (0xivanov) [reply examples/create-topic-with-revenue.js:286]: Yeah, setting the transaction id would work, but we'd need to sign/freeze. In my opinion setting the operator is easier and cleaner.

File: test/unit/TopicCreateTransaction.js (hunk 6 above)
Comment (venilinvasilev) [reply test/unit/TopicCreateTransaction.js:54]: No specific reason, makes sense, hence removing the iteration and going with index based access.

File: src/topic/TopicInfo.js (hunk 10 above)
Comment (venilinvasilev) [reply src/topic/TopicInfo.js:46]: Well I think this would make the prop required and you would always need to pass an empty array which can be annoying if you don't want to. That is if the library is used in TS enabled environment.
Also the protobuf's type mentions that this can be null.

**Review by ivaylogarnev-limechain — DISMISSED** [review]
LGTM

File: examples/create-topic-with-revenue.js [hunk 14]
```diff
@@ -0,0 +1,320 @@
...
+
+        const customFee = new CustomFixedFee()
+            .setAmount(new Hbar(1).toTinybars())
+            .setFeeCollectorAccountId(operatorId);
+
+        const { topicId } = await (
+            await new TopicCreateTransaction()
+                .setAdminKey(operatorKey)
+                .setFeeScheduleKey(operatorKey)
+                .setMaxTransactionFee(100)
```
0xivanov, line 75: last thing - remove all calls to `.setMaxTransactionFee` in this PR.
It's unnecessary and might cause some users to call it as well.

0xivanov (approved):
LGTM, not sure what this global.d.ts file was about :D

ivaylogarnev-limechain (approved):
LGTM

ivaylonikolov7:
> LGTM, not sure what this global.d.ts file was about :D

https://github.com/dcodeIO/long.js/commit/be983d36b47a51bc150a1dc04305e800fa48ab0e - I think after these changes we don't need the `global.d.ts` type definitions. They just released a new version like 1 hour ago and it might have included these changes.

ivaylonikolov7 (approved):

PavelSBorisov (approved):

PR #2888, opened 2025-02-22
fix(ecsda): return ecdsa signature with a normalized s value

mehcode:
An ECDSA _message_ signature generated by the JS SDK does not verify in the Rust SDK.

---

When exploring signing a message with `PrivateKey` in the JS SDK and verifying it with the Rust SDK, I noticed that the JS SDK does not return a normalized `s` value in the ECDSA signature. 

This is **not** a breaking change as the underlying `elliptic` library validates both normalized and non-normalized signatures but other SDKs only validate normalized signatures.

It is documented here - https://github.com/indutny/elliptic/wiki - that we need to use `canonical: true` to be Crypto-compatible.

### DIFF
--- a/packages/cryptography/src/primitive/ecdsa.js
+++ b/packages/cryptography/src/primitive/ecdsa.js
@@ -63,7 +63,14 @@ export function sign(keydata, message) {
     const msg = hex.encode(message);
     const data = hex.decode(keccak256(`0x${msg}`));
     const keypair = secp256k1.keyFromPrivate(keydata);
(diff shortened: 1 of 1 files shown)


0xivanov:
Some of the unit tests are failing, can you check why?

ivaylonikolov7:
We are currently in the process of deprecating `eliptic` package and replacing it with `@noble/curves`. It uses normalized values. This feature will be released in `v.2.62` as we want to test it thoroughly.

PR #2914, opened 2025-03-04
refactor: used fixtures in tests

venilinvasilev:
### Description:
This refactors parts of the integration tests to use fixtures.
Adds new fixtures to be used in other tests.
Changes the way transaction state is handled in fixtures, so it is consistent as in usage.
### Related issues:
https://github.com/hiero-ledger/hiero-sdk-js/issues/2872

### DIFF
--- a/test/integration/AccountBalanceIntegrationTest.js
+++ b/test/integration/AccountBalanceIntegrationTest.js
@@ -1,12 +1,9 @@
-import {
-    AccountBalanceQuery,
-    Status,
-    TokenCreateTransaction,
-} from "../../src/exports.js";
+import { AccountBalanceQuery, Status } from "../../src/exports.js";
 import IntegrationTestEnv, {
     Client,
     skipTestDueToNodeJsVersion,
 } from "./client/NodeIntegrationTestEnv.js";
+import { createFungibleToken } from "./utils/Fixtures.js";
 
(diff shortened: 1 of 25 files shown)


test/integration/AccountBalanceIntegrationTest.js
```diff
@@ -77,23 +74,23 @@ describe("AccountBalanceQuery", function () {
...
+        const tokenId = await createFungibleToken(env.client, (transaction) => {
+            transaction
+                .setInitialSupply(0)
+                .setAdminKey(null)
+                .setFreezeKey(null)
+                .setPauseKey(null)
+                .setWipeKey(null)
+                .setFeeScheduleKey(null)
+                .setMetadataKey(null)
+                .setSupplyKey(null);
```
ivaylogarnev-limechain, line 86: Should we explicitly set the keys to null to exclude them from the token, or isn’t this already the default behavior?

venilinvasilev: In most test cases in the codebase, operator keys need to be set. That’s why they are set as the default behavior in the fixture, reducing code repetition. However, this is a special case where we need to override them as unset.

ivaylogarnev-limechain (approved):
LGTM

PR #2928, opened 2025-03-05
feat: manual update client address book

venilinvasilev:
### Description:
This gives the possibility of the sdk client to update the address book manually and adds an example of usage.

### **Note:**
Current flow of the client network update:
**Client initialization ->>> (Optional) Manual network update → Wait → Scheduled network update → Wait → Repeat...**

We chose this approach to allow users to update the network manually, because initiating the address book update during client initialization would introduce a breaking change to users. Since the update process is asynchronous, it would require awaiting the operation.

### Checklist:
- [x] Implemented method for updating the network manually.
- [x] Added example for usage of this method and update network period.

Related issues:
https://github.com/hiero-ledger/hiero-sdk-js/issues/2918

### DIFF
--- a/examples/schedule-network-update.js
+++ b/examples/schedule-network-update.js
@@ -1,0 +1,74 @@
+import { AccountId, Client, PrivateKey } from "@hashgraph/sdk";
+
+import dotenv from "dotenv";
+
+dotenv.config();
+
+async function main() {
+    if (
+        process.env.OPERATOR_ID == null ||
+        process.env.OPERATOR_KEY == null ||
+        process.env.HEDERA_NETWORK == null
+    ) {
+        throw new Error(
+            "Environment variables OPERATOR_ID, HEDERA_NETWORK, and OPERATOR_KEY are required.",
+        );
+    }
+    const operatorId = AccountId.fromString(process.env.OPERATOR_ID);
+    const operatorKey = PrivateKey.fromStringECDSA(process.env.OPERATOR_KEY);
+
+    // Step 1: Initialize the client
+    const client = Client.forName(process.env.HEDERA_NETWORK).setOperator(
+        operatorId,
+        operatorKey,
+    );
+
+    console.log(
+        `Client initialized for network: ${client.ledgerId.toString()}`,
+    );
+
+    // Step 2: Examine the default network update period
+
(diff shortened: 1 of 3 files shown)


examples/schedule-network-update.js
```diff
@@ -0,0 +1,74 @@
...
+        process.env.HEDERA_NETWORK == null
+    ) {
+        throw new Error(
+            "Environment variables OPERATOR_ID, HEDERA_NETWORK, and OPERATOR_KEY are required.",
+        );
+    }
+    const operatorId = AccountId.fromString(process.env.OPERATOR_ID);
+    const operatorKey = PrivateKey.fromStringECDSA(process.env.OPERATOR_KEY);
+
+    // Step 1: Initialize the client
```
0xivanov, line 20: add here that by default this will trigger the scheduler with 24 hour timer and it will update the addressbook from the mirror node after this time

venilinvasilev: Thanks. I guess It should be more clear now.

examples/schedule-network-update.js
```diff
@@ -0,0 +1,79 @@
...
+    let networkUpdateInHours = client.networkUpdatePeriod / (1000 * 60 * 60);
+
+    console.log(
+        `The current default network update period is: ${networkUpdateInMinutes} minutes or ${networkUpdateInHours.toFixed(2)} hours.`,
+    );
+
+    /*
+     * Step 3: Update the address book of the client
+     * Note: This is optional, but it is recommended to keep the client with updated address book.
+     * This works only in node.js environment. For the browser client, the address book is handled by GRPC Web proxies.
```
0xivanov, line 48: We could start working on https://github.com/hiero-ledger/hiero-sdk-js/issues/2922 and if we implement the `nodeIP -> webProxy` mapping we can point PR into this branch and update this comment - since it would work with `WebClient`

0xivanov, line 48: We could start working on https://github.com/hiero-ledger/hiero-sdk-js/issues/2922 and if we implement the `nodeIP -> webProxy` mapping we can point PR into this branch and update this comment - since it would work with `WebClient`

ivaylogarnev-limechain (approved):
LGTM

0xivanov (approved):
LGTM

ivaylogarnev-limechain (approved):
LGTM

PR #2929, opened 2025-03-06
refactor: use fixtures in tests

venilinvasilev:
### Description:
This refactors the integration tests to use fixtures.

### Related issues:
https://github.com/hiero-ledger/hiero-sdk-js/issues/2872

### DIFF
--- a/test/integration/AccountBalanceIntegrationTest.js
+++ b/test/integration/AccountBalanceIntegrationTest.js
@@ -75,15 +75,7 @@ describe("AccountBalanceQuery", function () {
 
     it("should reflect token with no keys", async function () {
         const tokenId = await createFungibleToken(env.client, (transaction) => {
-            transaction
-                .setInitialSupply(0)
-                .setAdminKey(null)
-                .setFreezeKey(null)
-                .setPauseKey(null)
-                .setWipeKey(null)
-                .setFeeScheduleKey(null)
-                .setMetadataKey(null)
-                .setSupplyKey(null);
+            transaction.setInitialSupply(0);
         });
 
         const balances = await new AccountBalanceQuery()
--- a/test/integration/TokenNftIntegrationTest.js
+++ b/test/integration/TokenNftIntegrationTest.js
@@ -1,5 +1,4 @@
 import {
-    Hbar,
     NftId,
     Status,
     TokenAssociateTransaction,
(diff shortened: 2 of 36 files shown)


test/integration/TokenNftIntegrationTest.js
```diff
@@ -189,51 +168,42 @@ describe("TokenNft", function () {
...
-
-        const account = (
-            await (
-                await new AccountCreateTransaction()
-                    .setKeyWithoutAlias(key)
-                    .setInitialBalance(new Hbar(2))
-                    .execute(env.client)
-            ).getReceipt(env.client)
-        ).accountId;
+        const { accountId, newKey } = await createAccount(
```
ivaylogarnev-limechain, line 171: Why do you rename `key` to `newKey` and in the other files too?

test/integration/TransactionReceiptIntegrationTest.js
```diff
@@ -9,6 +8,8 @@ import {
     TransactionId,
 } from "../../src/exports.js";
 import IntegrationTestEnv from "./client/NodeIntegrationTestEnv.js";
+import { createAccount } from "./utils/Fixtures.js";
+import { expect } from "chai";
```
ivaylogarnev-limechain, line 12: Could you please move this import to the beginning since it's from an external library rather than the current repository?

0xivanov (commented):
In Go the ratio between **using the default fixture** to **using the fixture + override**s is 2:1.
Seems like here i's more like 1:1 or even worse. We should check how to improve it.

test/integration/AccountBalanceIntegrationTest.js
```diff
@@ -75,15 +75,7 @@ describe("AccountBalanceQuery", function () {
...
-            transaction
-                .setInitialSupply(0)
-                .setAdminKey(null)
-                .setFreezeKey(null)
-                .setPauseKey(null)
-                .setWipeKey(null)
-                .setFeeScheduleKey(null)
-                .setMetadataKey(null)
-                .setSupplyKey(null);
+            transaction.setInitialSupply(0);
```
ivaylonikolov7, line 78: do we need to set this initial supply to 0 if the test case check if token has keys?

test/integration/TokenNftIntegrationTest.js
venilinvasilev: It's copied from the go-sdk integration-test fixture

ivaylonikolov7 (approved):
Okay, my opinion is changed these are actually nice improvements. 👍

ivaylogarnev-limechain (approved):
LGTM

test/integration/AccountBalanceIntegrationTest.js
venilinvasilev: You can see that we assert for this supply afterwards. Maybe the test name is not that appropriate.

PR #2937, opened 2025-03-10
fix: replaced elliptic with @noble/curves

venilinvasilev:
### Description:
This PR resolves a security issue by replacing the elliptic library (which has known vulnerabilities and could leak private keys) with [@noble/curves](https://github.com/paulmillr/noble-curves), a modern, audited, and high-performance alternative for elliptic curve cryptography.

Unlike elliptic, @noble/curves automatically enforces a normalized s value in ECDSA signatures, aligning with best practices in cryptographic security. This ensures better interoperability while preventing malleability-related issues.
This is NOT a Breaking Change. The SDK continues to function exactly as before, as @noble/curves produces valid signatures compatible with existing verification logic. The normalized s value behavior is already an industry standard (Ethereum's EIP-2, Bitcoin's BIP-66).

**Related issue(s)**:

Fixes https://github.com/hiero-ledger/hiero-sdk-js/issues/2911

Closes:  https://github.com/hiero-ledger/hiero-sdk-js/pull/2888


packages/cryptography/src/primitive/ecdsa.js
```diff
@@ -12,11 +10,12 @@ const secp256k1 = new elliptic.ec("secp256k1");
  * @returns {KeyPair}
  */
 export function generate() {
-    const keypair = secp256k1.genKeyPair();
+    const privateKey = secp256k1.utils.randomPrivateKey();
```
ivaylonikolov7, line 13: possible expo issue here. did we check if the `randomPrivateKey` works on native?

packages/cryptography/src/EcdsaPublicKey.js
```diff
@@ -58,38 +57,37 @@ export default class EcdsaPublicKey extends Key {
...
-                ecdsaPublicKeyBytes = hex.decode(compressedPublicKeyBytes);
+            default: // Uncompressed DER public keys
+                try {
+                    const keyPair = secp256k1.ProjectivePoint.fromHex(
+                        data.subarray(derPrefixBytes.length),
+                    );
+                    ecdsaPublicKeyBytes = keyPair.toRawBytes(true); // Compressed format
+                } catch (error) {
+                    throw new BadKeyError(
+                        `cannot decode ECDSA public key from DER format`,
```
ivaylonikolov7, line 82: Isn't `from THIS DER format` a better error, as the current error insinuates that we can decode public key from DER format

packages/cryptography/src/primitive/ecdsa.js
```diff
@@ -34,11 +33,12 @@ export async function generateAsync() {
...
 export function fromBytes(data) {
-    const keypair = secp256k1.keyFromPrivate(data);
+    const privateKey = new Uint8Array(data);
+    const publicKey = secp256k1.getPublicKey(privateKey, true);
 
     return {
-        privateKey: hex.decode(keypair.getPrivate("hex")),
-        publicKey: hex.decode(keypair.getPublic(true, "hex")),
+        privateKey: privateKey,
+        publicKey: publicKey,
```
ivaylonikolov7, line 41: nit: 

return {
privateKey,
publicKey
}

ivaylonikolov7 (commented):
Isn't there a situation where a wallet like Kabilla's or Hashpack's signs a transaction using their own implementation, they pass the signature to the SDK using `addSignature` and the SDK finds that it uses non-normalized signature and then the SDK will throw an error. This might be an issue for the wallets. Correct me I am wrong.

packages/cryptography/src/EcdsaPublicKey.js
venilinvasilev: Good point.

packages/cryptography/src/primitive/ecdsa.js
venilinvasilev: Should work as it uses the underlying crypto module which is pollyfilled by get-random-values.
I'll manually check this to confirm for sure.
I think we should consider adding the tests for native environment as well.

packages/cryptography/package.json
```diff
@@ -59,12 +57,12 @@
         "last 1 Firefox versions"
     ],
     "dependencies": {
+        "@noble/curves": "^1.8.1",
```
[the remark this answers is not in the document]
venilinvasilev: We use some util functions from this library which are not present in the above mentioned libs. I'll take my time to see if we could omit these somehow.
PS: We have used `elliptic` library for generating secp256k1 keys, so I guess we can use the [noble-secp256k1](https://github.com/paulmillr/noble-secp256k1) lib for that. On the other hand - for Ed25519 keys we are using `tweetnacl` which is another library and I have not touched any code there. Maybe it is a good idea to create an issue and migrate the functionallity for ED25519 keys to noble as well.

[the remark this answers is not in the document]
venilinvasilev: Done

packages/cryptography/src/EcdsaPublicKey.js
venilinvasilev: Done

packages/cryptography/src/primitive/ecdsa.js
venilinvasilev: Done

ivaylonikolov7 (approved):

ivaylogarnev-limechain (approved):

ivaylonikolov7:
> Isn't there a situation where a wallet like Kabilla's or Hashpack's signs a transaction using their own implementation, they pass the signature to the SDK using `addSignature` and the SDK finds that it uses non-normalized signature and then the SDK will throw an error. This might be an issue for the wallets. Correct me I am wrong.

I have been testing the new signatures using for the past day using Kabilla's wallet and then using Hashpack's wallet. I couldn't reproduce a issue where this might cause a regression. The scenario I was testing on is the following: 

- React dapp that uses the latest version of the SDK
- Kabilla/Hashgraph wallet that uses an old version of the SDK that uses the old signatures
- I sign an `AccountUpdateTransaction` using the wallet's signer and execute it on the wallet. The transactions doesn't fail and the signature's `verify` method return that the transaction's signature is legit.

PR #2959, opened 2025-03-17
chore: update dependencies

venilinvasilev:
**Description**:
This update dependencies of the sdk, the subpackage dependencies and the examples dependencies.


examples/react-native-example/yarn.lock
```diff
@@ -1374,8 +1374,8 @@
     "@ethersproject/bytes" "^5.7.0"
     "@ethersproject/rlp" "^5.7.0"
     "@grpc/grpc-js" "^1.12.6"
-    "@hashgraph/cryptography" "file:.yalc/@hashgraph/sdk/.yalc/@hashgraph/cryptography"
-    "@hashgraph/proto" "file:.yalc/@hashgraph/sdk/.yalc/@hashgraph/proto"
+    "@hashgraph/cryptography" "file:../../../../../Library/Caches/Yarn/v6/npm-@hashgraph-sdk-2.60.0-beta.1-f0cdc7c4-37c3-4ad5-b2dd-f27e7fd8c29f-1742144139226/node_modules/@hashgraph/sdk/.yalc/@hashgraph/cryptography"
```
ivaylonikolov7, line 1377: not sure if this should point to `/library/caches/`

ivaylonikolov7 (changes requested):

examples/react-native-example/yarn.lock
venilinvasilev: Weirdly enough I have no idea how this got in there. I've reverted this change. It should be okay now.

0xivanov:
What do you guys think to release these in 2-3 batches?

venilinvasilev:
I agree. Lets do this in bathes as it is risky to merge all bumped dependencies in one go.

PR #2961, opened 2025-03-17
feat: add transaction size for chunked and nonchunked transaction

### LINKED ISSUE #2489: Convenience function to determine current size of transaction
### Problem

In order to determine the size of the transaction to ensure it is within the size limits, a few steps needs to be taken. Currently, the most accurate way to do this looks similar to this: 
`const request = await transaction._makeRequestAsync();
proto.Transaction.encode(request).finish().length`

Especially for new developers, a method on the transaction that returns this value would be very helpful for client side apps.

### Solution

Add a convenience function that will return the protobuf size of the transaction that will actually be received by the network. 

### Alternatives

_No response_

ivaylonikolov7:
**Description**:
Adds size for normal transaction and a getter that returns an array with the size of all chunked transactions for transactions like `FileAppendTransaction`.

Fixes #2489

**Checklist**


0xivanov:
Looks like a cool new feature 🚀

src/transaction/Transaction.js
```diff
@@ -542,6 +542,34 @@ export default class Transaction extends Executable {
...
+    /**
+     * Get the transaction body size
+     * Protobuf encoding has specific rules about how data is serialized
+     * Different fields take different amounts of space depending on their values
+     * The actual wire format size can only be determined after encoding
+     *
+     * @returns {number}
+     */
+    get bodySize() {
+        const body = this._makeTransactionBody(AccountId.fromString("0.0.0"));
```
ivaylogarnev-limechain, line 568: Why here its hardcoded `AccountId.fromString("0.0.0")`?

test/unit/Transaction.js
```diff
@@ -640,4 +639,181 @@ describe("Transaction", function () {
             expect(result.size).to.equal(0);
         });
     });
+
+    describe.only("size", function () {
```
ivaylogarnev-limechain: Reminder to `.only`

0xivanov (commented):
Maybe it's a good idea to have some kind of process for adding new APIs like this.
Might need to involve the TSC, @SimiHunjan @hendrikebbers what do you think?

src/file/FileAppendTransaction.js
```diff
@@ -245,6 +245,31 @@ export default class FileAppendTransaction extends Transaction {
...
+     * Get the body sizes for all chunks in a FileAppendTransaction.
+     * For transactions with multiple chunks (like large file appends),
+     * this returns an array containing the size of each chunk's transaction body.
+     * The size is calculated by encoding the transaction body to protobuf format.
+     *
+     * @returns {number[]} An array of body sizes, where each element represents
+     * the size in bytes of a chunk's transaction body
+     *
+     */
+    get bodySizeAllChunks() {
```
0xivanov, line 258: shouldn't this function be in Transactino.js?

venilinvasilev (approved):
LGTM

ivaylogarnev-limechain (approved):
LGTM

0xivanov (approved):

Issue #2923, opened 2025-03-04
`TCK-JS`: Implement JSON-RPC method endpoint for `TransferTransaction`


rwalworth:
Implement the `transferCrypto` endpoint and test the JS SDK against the tests in the TCK for `TransferTransaction`.

rwalworth:
Closed by #2975

PR #2975, opened 2025-03-24
chore(TCK): Added `transferCrypto` endpoint

ivaylogarnev-limechain:
**Description**:
This PR adds `transferCrypto` endpoint listed in the following [issue](https://github.com/hashgraph/hedera-sdk-js/issues/2923).

**Related issue(s)**:
#2923

**Checklist**


src/EvmAddress.js
```diff
@@ -28,10 +28,23 @@ export default class EvmAddress extends Key {
...
 
     /**
-     * @param {string} text
+     * Creates an EvmAddress from a hex string representation.
+     * @param {string} text - The hex string representing the EVM address
      * @returns {EvmAddress}
+     * @throws {Error} If the input string is not the correct size
      */
     static fromString(text) {
+        const EVM_ADDRESS_BYTES = 20; // Standard EVM address is 20 bytes
```
0xivanov, line 37:
```js
  /**
     * @param {string} evmAddress
     * @returns {EvmAddress}
     */
    static fromString(evmAddress) {
        evmAddress = evmAddress.startsWith("0x")
            ? evmAddress.slice(2)
            : evmAddress;

        // Standard EVM address is 20 bytes
        if (evmAddress.length !== 20) {
            throw new Error("Input EVM address string is not the correct size");
        }

        return new EvmAddress(hex.decode(evmAddress));
    }
```

tck/params/common-tx-params.ts
```diff
@@ -32,8 +33,15 @@ export const applyCommonTransactionParams = (
...
-        transaction.setTransactionId(txId);
+        try {
+            const txId = TransactionId.fromString(transactionId);
+            transaction.setTransactionId(txId);
+        } catch (error) {
+            const txId = TransactionId.generate(
+                AccountId.fromString(transactionId),
+            );
+            transaction.setTransactionId(txId);
+        }
```
0xivanov, line 44:
```suggestion
transaction.setTransactionId(TransactionId.generate(AccountId.fromString(payerId)))
```

ivaylogarnev-limechain: We don't have to generate AccountId from string every time, so we can first try if it's a direct transactionId and then fall back to the other approach.

0xivanov: Do we actually have tck test that passes the `transactionID` as a whole?
I think we don't, but if we did - i think it's best to leave the TCK SDK to generate the txID, and only specify the payerID. While generating txID, the SDK creates timestamp so it's best to leave it to the TCK to do it.

0xivanov (approved):
LGTM

PR #2979, opened 2025-03-24
build: disable treeshaking for queries

ivaylonikolov7:
**Description**:

Disable treeshaking for queries. Queries are treeshaken on webpack build which breaks the CostQuery. 

**Checklist**

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,9 +39,8 @@
         "transactions"
     ],
     "sideEffects": [
-        "./src/query/Query.js",
-        "./src/query/CostQuery.js",
-        "**/*Query.js"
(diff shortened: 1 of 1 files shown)


venilinvasilev:
Do these changes have any negative impact on the bundle size?
Other than that, I think it would be good to add some tests for apps built with Rollup and Webpack.

ivaylonikolov7:
Will have some impact on the bundle size but won't increase the size significantly. I will test this tomorrow and give you feedback.

ivaylonikolov7:
superseded by #2985

PR #2980, opened 2025-03-24
chore: add cdn

ivaylonikolov7:
**Description**:
TBD

**Related issue(s)**:
TBD

Fixes #

**Checklist**


package.json
```diff
@@ -41,7 +41,8 @@
     "sideEffects": false,
     "files": [
         "lib/",
-        "src/"
+        "src/",
+        "umd/"
```
bguiz, line 45: By convention, this directory would be better named `dist/`
```suggestion
        "dist/"
```
That allows us to reuse the dir in future for other bundle formats, e.g. ESM, CJS, etc. Or even include variants for different client target ECMAscript versions.

rollup.config.js
```diff
@@ -111,4 +114,47 @@ export default [
...
+                browser: true,
+                preferBuiltins: false,
+            }),
+            commonjs({
+                transformMixedEsModules: true,
+            }),
+            json(),
+        ],
+        output: {
+            file: "umd/umd.js",
```
bguiz, line 131:
```suggestion
            file: dist/umd.js",
```

rollup.config.js
```diff
@@ -111,4 +114,47 @@ export default [
...
+                transformMixedEsModules: true,
+            }),
+            json(),
+            terser(),
+        ],
+        output: {
+            format: "umd",
+            name: "sdk",
+            sourceMap: true,
+            file: "umd/umd.min.js",
```
bguiz, line 156:
```suggestion
            file: "dist/umd.min.js",
```

bguiz (approved):
LGTM. (some minor comments added, nice to have, but non-essential)

rollup.config.js
ivaylonikolov7: Fair point, I was looking at https://www.unpkg.com/ to use as CDN. I saw that they recommended using `umd/` but `dist` makes more sense as we might like to use `esm` modules too at some point.

PavelSBorisov (approved):

rollup.config.js
ivaylonikolov7: Should be resolved by now.

PR #2994, opened 2025-03-28
fix: accountId instantiated from EVM address incorrectly sets evmAddress

### LINKED ISSUE #2982: AccountId instantiated from EVM address incorrectly sets evmAddress causing transaction failures
### Description

When using AccountId.fromEvmAddress().populateAccountNum() in the Hiero SDK (JavaScript), the returned AccountId object retains the original evmAddress. Although populateAccountNum() successfully fetches and assigns the correct Hedera account number, the presence of evmAddress in the object causes subsequent transactions (such as payments or token associations) to fail with a PAYER_ACCOUNT_NOT_FOUND error. Manually setting evmAddress to null resolves the issue, indicating that the SDK improperly prioritizes the alias (evmAddress) over the correctly populated account number.

### Steps to reproduce

Instantiate a Hedera Client for Testnet.

Use AccountId.fromEvmAddress(shard, realm, evmAddress).populateAccountNum(client) to get an AccountId.

Use this AccountId directly in transactions, such as TokenAssociateTransaction or a simple transfer.

Execute the transaction.

Expected behavior:
The transaction succeeds using the correctly populated account number.

Actual behavior:
Transaction fails with PAYER_ACCOUNT_NOT_FOUND error:

StatusError: transaction 0.0.evmAddress@timestamp failed precheck with status PAYER_ACCOUNT_NOT_FOUND against node account id 0.0.5

Workaround:
Explicitly set evmAddress to null:

const accountId = await AccountId.fromEvmAddress(0, 0, evmAddress).populateAccountNum(client);
accountId.evmAddress = null;

### Additional context

This issue occurs consistently when the AccountId object includes both an evmAddress and a populated Hedera account number. The SDK should handle this scenario gracefully, prioritizing the numeric account ID when available, without requiring manual nullification of the evmAddress property.

### Hedera network

testnet

### Version

v2.61.0

### Operating system

macOS

ivaylonikolov7:
**Description**:
This PR improves the handling of account aliases and hollow accounts in the AccountId class by making the protobuf conversion more explicit and adding comprehensive test coverage.

- Add support for hollow account detection (num=0 with alias)
- More explicing protobuff conversion when:
   1.  both `AccountId` `num` and `alias` are set
   2. only `alias` is set
   3. only `AccountId` `num` is set

Fixes #2982

**Notes for reviewer**: I am aware that there is some shorter syntax with the tenary operator but I don't think it's worth to safe a bunch of bytes at the cost of readability of the code. 

**Checklist**

### DIFF
--- a/src/account/AccountId.js
+++ b/src/account/AccountId.js
@@ -320,12 +320,35 @@ export default class AccountId {
             evmAddress = this.evmAddress._bytes;
         } */
 
+        const isHollowAccount = this.num.eq(Long.fromBigInt(0n)) && alias;
+
+        if (isHollowAccount) {
+            return {
+                alias,
+                accountNum: null,
+                shardNum: this.shard,
+                realmNum: this.realm,
+            };
+        } else if (this.num && alias) {
+            return {
+                alias: null,
+                accountNum: this.num,
+                shardNum: this.shard,
+                realmNum: this.realm,
+            };
+        } else if (alias) {
+            return {
+                alias,
+                accountNum: null,
+                shardNum: this.shard,
+                realmNum: this.realm,
+            };
+        }
         return {
-            alias,
-            accountNum: this.aliasKey != null ? null : this.num,
+            alias: null,
+            accountNum: this.num,
             shardNum: this.shard,
             realmNum: this.realm,
-            //evmAddress,
         };
     }
 
--- a/src/account/AccountId.js
+++ b/src/account/AccountId.js
@@ -323,6 +323,7 @@ export default class AccountId {
         const isHollowAccount = this.num.eq(Long.fromBigInt(0n)) && alias;
 
         if (isHollowAccount) {
+            // if is hollow accont pass accountNum null
             return {
                 alias,
                 accountNum: null,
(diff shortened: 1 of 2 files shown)


src/account/AccountId.js
```diff
@@ -320,12 +320,39 @@ export default class AccountId {
...
+        } else if (alias) {
+            // then if alias is set ignore the account num
+            return {
+                alias,
+                accountNum: null,
+                shardNum: this.shard,
+                realmNum: this.realm,
+            };
+        }
+        // all other cases ignore the alias
```
venilinvasilev, line 350: Can we simplify these checks somehow. Maybe it will be better if we construct the returned AccountId object then return it? WDYT?

src/account/AccountId.js
```diff
@@ -320,12 +320,39 @@ export default class AccountId {
...
+            };
+        } else if (alias) {
+            // then if alias is set ignore the account num
+            return {
+                alias,
+                accountNum: null,
+                shardNum: this.shard,
+                realmNum: this.realm,
+            };
+        }
```
ivaylogarnev-limechain, line 349:
```suggestion
if (alias) {
    return {
        alias: isHollowAccount ? alias : null,
        accountNum: isHollowAccount ? null : this.num,
        shardNum: this.shard,
        realmNum: this.realm,
    };
}

return {
    alias: null,
    accountNum: this.num,
    shardNum: this.shard,
    realmNum: this.realm,
};
```

venilinvasilev (approved):
LGTM

PR #3021, opened 2025-04-03
feat: add signature recovery id calculation

venilinvasilev:
## Description:
This PR introduces a new method, `getRecoveryId(r, s, message)`, to the `PrivateKey` class. It enables calculation of the  recovery ID from a ECDSA signature (v) directly from the raw r and s signature components and the original message.

## Motivation:
Currently, there is no way to calculate the recovery ID (v) when signing Ethereum transactions using the JS SDK. The recovery ID is a required component of the v value in Ethereum signatures, and is essential for constructing valid EIP-155-compliant transactions.

This method provides the missing functionality, allowing developers to:

Recover the v value after signing.

Fully sign and serialize Ethereum transactions within the SDK.

Interoperate with external Ethereum clients, networks, and signing tools.

## Added
Added PrivateKey.getRecoveryId(r, s, message) method to enable recovery ID (v) calculation from raw ECDSA signature components.
This addition allows developers to compute the recovery ID directly, enabling full support for signing and serializing Ethereum transactions (including EIP-155) within the SDK. It enhances compatibility with external Ethereum clients and tooling.

**Checklist:**
- [x]  Documented (Code comments, README, etc.)
- [x]  Tested (unit, integration, etc.)

**Related issues:** https://github.com/hiero-ledger/hiero-sdk-js/issues/2269


test/integration/EthereumTransactionIntegrationTest.js
```diff
@@ -131,11 +132,15 @@ describe.skip("EthereumTransactionIntegrationTest", function () {
...
+        const message = hex.decode(type + encoded);
+        const signedBytes = privateKey.sign(message);
         const middleOfSignedBytes = signedBytes.length / 2;
         const r = signedBytes.slice(0, middleOfSignedBytes);
         const s = signedBytes.slice(middleOfSignedBytes, signedBytes.length);
-        const v = hex.decode("01"); // recovery id
+
+        const recoveryId = privateKey.getRecoveryId(r, s, message);
+
+        const v = new Uint8Array(recoveryId === 0 ? [] : [recoveryId]);
```
ivaylonikolov7, line 143: Can we explain to future devs why we do this so the developers after us knows that if recoveryId is 0 it should be empty array and not an array with 0 in it.

packages/cryptography/src/PrivateKey.js
```diff
@@ -515,6 +515,32 @@ export default class PrivateKey extends Key {
...
+     *
+     * **Note:** This method only works for ECDSA secp256k1 keys.
+     * @param {Uint8Array} r - 32-byte `r` component of the signature
+     * @param {Uint8Array} s - 32-byte `s` component of the signature
+     * @param {Uint8Array} message - The original (unhashed) message
+     * @returns {number} Recovery ID (0–3), or -1 if not found or not applicable
+     */
+    getRecoveryId(r, s, message) {
+        if (!(this._key instanceof EcdsaPrivateKey)) {
+            return -1;
```
ivaylonikolov7, line 530: Why -1? Doesn't it make sense to throw an error instead of -1?

venilinvasilev: I am reflecting the behavior from other sdks (go) for consistency.

packages/cryptography/src/primitive/ecdsa.js
```diff
@@ -81,3 +81,47 @@ export function verify(keydata, message, signature) {
...
+
+            if (equalBytes(recovered, expectedPubKey)) {
+                return recovery;
+            }
+        } catch {
+            // Ignore invalid recoveries
+        }
+    }
+
+    return -1;
```
ivaylonikolov7, line 112: Same as previous comment. Doesn't it make sense to throw an error. This -1 might cause some issues which might be harder to debug.

test/integration/EthereumTransactionIntegrationTest.js
venilinvasilev: Good idea! We sure can. Done.

packages/cryptography/src/primitive/ecdsa.js
```diff
@@ -81,3 +81,47 @@ export function verify(keydata, message, signature) {
...
+ * @param {Uint8Array} b
+ * @returns {boolean}
+ */
+function equalBytes(a, b) {
+    if (a.length !== b.length) return false;
+    for (let i = 0; i < a.length; i++) {
+        if (a[i] !== b[i]) return false;
+    }
+    return true;
+}
```
ivaylonikolov7, line 127: wdyt about moving this to the `utils.js` file. We might use this in other cases too.

packages/cryptography/src/PrivateKey.js
0xivanov: In go we cannot throw - in js it's valid way to handle it though.

packages/cryptography/src/primitive/ecdsa.js
venilinvasilev: Same answer as above -> https://github.com/hiero-ledger/hiero-sdk-go/pull/1324/files#diff-2a661bc447bce86cd902a475e5653fe2c4e9e4ae82731a6a42aabe4db87cc856

packages/cryptography/src/primitive/ecdsa.js
```diff
@@ -81,3 +81,47 @@ export function verify(keydata, message, signature) {
...
+
+/**
+ * Byte comparison utility
+ * @param {Uint8Array} a
+ * @param {Uint8Array} b
+ * @returns {boolean}
+ */
+function equalBytes(a, b) {
+    if (a.length !== b.length) return false;
```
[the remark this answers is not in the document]
venilinvasilev: Done

packages/cryptography/src/primitive/ecdsa.js
venilinvasilev: Done

packages/cryptography/src/PrivateKey.js
venilinvasilev: We agreed that we should be more explicit with the errors for different cases. I've thrown them on lower level.

packages/cryptography/src/primitive/ecdsa.js
venilinvasilev: Replied above. Done

ivaylonikolov7 (approved):

ivaylogarnev-limechain (approved):

Issue #3030, opened 2025-04-07
TopicUpdateTransaction - Maintain null state for feeExemptKeys and customFees


venilinvasilev:
## Description

In the implementation of the `TopicUpdateTransaction` class for HIP-991 (Permissionless revenue generating topics), when `feeExemptKeys` or `customFees` properties are not explicitly set, they should remain `null` in both the object state and the generated transaction data. This ensures that the transaction doesn't include empty structures for these fields when they haven't been defined.

From a product perspective, the issue is that previously, when these fields were left unset, the system would still include empty arrays in the transaction. This led to unintended updates, such as clearing out the existing topic fees and `feeExemptKeys`, which was not the user's intent.

## Current Behavior

- The `_makeTransactionData()` method in the `TopicUpdateTransaction` class may improperly handle the `null` state of `feeExemptKeys` and `customFees` properties.
- This could lead to empty arrays/objects being included in the transaction when these properties haven't been explicitly set by the user.
- As a result, existing values on the topic (fees or exempt keys) may be overwritten with empty values.

## Expected Behavior

- When `feeExemptKeys` and `customFees` haven't been set, they should remain `null` in the object state.
- The `_makeTransactionData()` method should generate transaction data with `null` values for `feeExemptKeyList` and `customFees` fields when these properties are `null`.
- After serialization and deserialization, these properties should still be `null` if they weren't set.
- Transactions should not unintentionally remove existing values for fees or exempt keys when not explicitly modified.

## Technical Details

When generating the protobuf transaction data in `_makeTransactionData()`, we need to ensure that:
- Both `feeExemptKeyList` and `customFees` correctly evaluate to `null` when their respective properties aren't set, rather than empty objects or arrays.

## Testing

Added tests to verify:
- `getFeeExemptKeys()` and `getCustomFees()` return `null` when not set.
- `_makeTransactionData()` correctly generates transaction data with `null` values for these fields.
- Serialization/deserialization preserves the `null` state of these properties.

## Related

- [[HIP-991: Permissionless revenue generating topics implementation](https://hips.hedera.com/hip/hip-991)](https://hips.hedera.com/hip/hip-991)

### Steps to reproduce

1. Create a new instance of `TopicUpdateTransaction`.
2. Do **not** set the `feeExemptKeys` or `customFees` properties.
3. Call `_makeTransactionData()` on the instance.
4. Observe the generated transaction data to check if `feeExemptKeyList` and `customFees` are `null` or empty.
[...]

PR #3031, opened 2025-04-07
fix(hip-991): null fees handling

venilinvasilev:
## Summary

Fixes a bug in `TopicUpdateTransaction` related to the handling of `feeExemptKeys` and `customFees` properties when they are unset. Previously, the transaction would include empty arrays for these fields, unintentionally clearing existing values on the topic. This change ensures that unset properties are treated as `null`, preserving existing topic configurations.

## Changes

- Ensures `feeExemptKeys` and `customFees` remain `null` in the object state when not explicitly set.
- `_makeTransactionData()` now excludes `feeExemptKeyList` and `customFees` from the transaction if they are `null`.
- Serialization and deserialization preserve `null` state for these fields.
- Added unit tests to verify correct behavior and prevent regressions.

## Motivation

Unintentionally resetting topic configurations is a critical issue, especially in revenue-generating scenarios. This fix ensures users don't accidentally remove existing fees or exemptions by omitting those properties during an update.

## Related

- Closes: #3030
- [HIP-991: Permissionless revenue generating topics implementation](https://hips.hedera.com/hip/hip-991)

## Checklist

- [x] Bug fix implemented
- [x] Tests added and passing
- [x] Verified serialization and deserialization behavior

### DIFF
--- a/src/topic/TopicUpdateTransaction.js
+++ b/src/topic/TopicUpdateTransaction.js
@@ -104,9 +104,9 @@ export default class TopicUpdateTransaction extends Transaction {
 
         /**
          * @private
-         * @type {Key[]}
+         * @type {?Key[]}
          */
-        this._feeExemptKeys = [];
+        this._feeExemptKeys = null;
 
         if (props.feeExemptKeys != null) {
             this.setFeeExemptKeys(props.feeExemptKeys);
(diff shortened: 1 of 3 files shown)


0xivanov (commented):
Looks good!
Can you also change the `clearFeeScheduleKey` to set the key to empty `KeyList` and add a test?

0xivanov (approved):
LGTM

ivaylonikolov7 (approved):

Issue #2926, opened 2025-03-05
SDK returned node unhealthy when the transaction was successful


pathornteng:
### Description

Hi team, I got a report from our partner that they got an error returned from the SDK as follows 

`
Connectivity Issue: All nodes are unhealthy 
`

but the actual transaction was successful. I don't have a way to reliably reproduce this but it keeps happening randomly. 

### Steps to reproduce

Keep sending transactions, this error will happen randomly. There is no current way to reliably reproduce this

### Hedera network

mainnet

### Version

v2.59

### Operating system

Linux

tomachianura:
this bug was reported by us days ago.
thanks @pathornteng for creating a github issue, truly appreciated.

this issue is huge, cause it's returning false negatives, causing high risks of duplicated payouts/rewards and high risks of messing up any engine relying on the SDK returned value as source of truth.

it is also unsustainable to ask us to query the mirror node to "double check" the result, the SDK must be 100% reliable on every single response you give back to the developer.

SimiHunjan:
@tomachianura you are receiving the same error `Connectivity Issue: All nodes are unhealthy` or a different one?

@pathornteng do you have more information on the following:
- example code snippet for one of the time this error is seen? Is there a specific transaction type your are executing?
- geographic location

HoudiniNFT:
Have seen this happen in both reward airdrops and nft minting.  actions will either actually fail or be successful but say they fail with that unhealthy node error or similar for no apparent reason.  its a nightmare to have to sort through 1000s of txns to find and check for errors when it occurs.  The network used to be extremely reliable but has become less so in the last 4-6 motnhs.  Have also noticed that NFT airdrops create NO txn data at all using the SDK.... no trace of the txn but NFT airdrops as expected.

ivaylonikolov7:
Last night the network was going through some maintenance. I wonder if this has something to do with it. 

If you can send us some kind of code snippet to reproduce this that would be great.

HoudiniNFT:
this has been ongoing for a few months sadly. not related to last night.
  
​

ivaylogarnev-limechain:
Thank you all for the detailed info and feedback on this issue! I’m actively investigating and want to ensure we resolve this quickly, given its impact.

From what I understand, the issue appears to be that transactions are being marked as successful, but the response contains the specific error `Connectivity Issue: All nodes are unhealthy` causing **false negatives**.

[...]

PR #3038, opened 2025-04-10
fix: Transaction receipt unhealthy error

ivaylogarnev-limechain:
**Description**:
This PR fixes execution of `getReceiptQuery` and `getRecordQuery` that are single-node requests, meaning they rely on one specific node to succeed. If that node is unhealthy, these queries can fail with errors such as "Network connectivity issue: All nodes are unhealthy." To handle this, we retry the request up to 10 times before failing with a delay.

**Related issue(s)**:
#2926 

**Checklist**

### DIFF
--- a/src/Executable.js
+++ b/src/Executable.js
@@ -634,6 +634,14 @@ export default class Executable {
                     this._nodeAccountIds.index ===
                     this._nodeAccountIds.list.length - 1;
 
+                // Check if the request is a transaction receipt or record
+                // request to retry 10 times, because its getReceiptQuery/getRecordQuery
+                // are single node requests
+                if (isTransactionReceiptOrRecordRequest(request)) {
+                    await delayForAttempt(false, 1, 500, 500);
+                    continue;
+                }
+
                 if (isLastNode || this._nodeAccountIds.length <= 1) {
                     throw new Error(
                         `Network connectivity issue: All nodes are unhealthy. Original node list: ${this._nodeAccountIds.list.join(", ")}`,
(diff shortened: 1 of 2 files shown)


src/Executable.js
```diff
@@ -634,6 +634,14 @@ export default class Executable {
                     this._nodeAccountIds.index ===
                     this._nodeAccountIds.list.length - 1;
 
+                // Check if the request is a transaction receipt or record
+                // request to retry 10 times, because its getReceiptQuery/getRecordQuery
```
0xivanov, line 638:
```suggestion
                // request to retry 10 times, because getReceiptQuery/getRecordQuery
```

src/Executable.js
```diff
@@ -813,6 +821,25 @@ export default class Executable {
...
+ * @template T
+ * @param {T} request - The request to check
+ * @returns {boolean} - True if the request is a transaction receipt or record
+ */
+function isTransactionReceiptOrRecordRequest(request) {
+    if (typeof request !== "object" || request === null) {
+        return false;
+    }
+
+    const requestType = Object.keys(request)[0];
```
venilinvasilev, line 836: Are we guaranteed that the first key of the request object will always be transactionGetReceipt or transactionGetRecord. Can those keys be on some other index. Can we make this check more deterministic and safe like:

```js
return ('transactionGetReceipt' in request || 'transactionGetRecord' in request)
```

src/Executable.js
```diff
@@ -634,6 +634,14 @@ export default class Executable {
                     this._nodeAccountIds.index ===
                     this._nodeAccountIds.list.length - 1;
 
+                // Check if the request is a transaction receipt or record
+                // request to retry 10 times, because its getReceiptQuery/getRecordQuery
+                // are single node requests
+                if (isTransactionReceiptOrRecordRequest(request)) {
+                    await delayForAttempt(false, attempt, 500, 1000);
+                    continue;
```
ivaylonikolov7, line 642: await delayForAttempt(
                        isLocalNode,
                        attempt,
                        this._minBackoff,
                        this._maxBackoff,
                    );

venilinvasilev (approved):
LGTM

ivaylonikolov7 (approved):

Issue #2878, opened 2025-02-18
HIP-551: Batch Transactions


0xivanov:
## Description

HIP: https://hips.hedera.com/hip/hip-551

This HIP introduces new transactions and updates to existing transactions in the consensus node software. SDKs will need to support these new transactions and updates to existing ones so that developers can take advantage of the new features. Propose the API design all SDKs should adhere to.

ivaylonikolov7:
resolved via #3039

PR #3039, opened 2025-04-10
feat: HIP 551

ivaylonikolov7:
**Description**:
This HIP defines a mechanism to execute batch transactions such that a series of transactions (HAPI calls) depending on each other can be rolled into one transaction that passes the ACID test (atomicity, consistency, isolation, and durability).
Example usage can be found in 
`examples/batch-tx.js`

### BatchTransaction

- `List<Transaction> innerTransactions`: The list of `Transaction`s to execute as part of the batch.
    - `BatchTransaction setInnerTransactions(List<Transaction>)`
    - `BatchTransaction addInnerTransaction(Transaction)`
    - `List<Transaction> getInnerTransactions()`
- `List<TransactionId> getInnerTransactionIds`: List of innerTransactionIds that have been added to the batch. Later used to query them and verify their status.

### Transaction

- `Key batchKey`: The key that must sign the `BatchTransaction` of which this `Transaction` is a part.
    - `Transaction batchify(Client client, Key batchKey)` : this method sets the batch key and marks a transaction as part of a batch transaction (inner transaction). The transaction is signed by the client of the operator and frozen.
    - `Transaction setBatchKey(Key)`: sets the batch key
    - `Key getBatchKey()` : gets the batch key
- `freezeWith`: Additional check if batchKey is `null`.
    - If not present → indication that this transaction is not meant to be inside a batch
    - If preset → indication that this transaction is part of a batch and setting `nodeAccountId` to `0.0.0`

Fixes #
#2878 

**Checklist**


packages/proto/src/proto/services/contract_call_local.proto
```diff
@@ -14,30 +14,15 @@ syntax = "proto3";
...
+// SPDX-License-Identifier: Apache-2.0
 option java_package = "com.hederahashgraph.api.proto.java";
 // <<<pbj.java_package = "com.hedera.hapi.node.contract">>> This comment is special code for setting PBJ Compiler java package
 option java_multiple_files = true;
 
-import "basic_types.proto";
-import "contract_types.proto";
-import "query_header.proto";
-import "response_header.proto";
+import "services/basic_types.proto";
```
0xivanov, line 22: it would be easier to remove the `services/` prefix from 10-15 proto files, than adding it to every file.

ivaylonikolov7: We would have to remove it on every protobuf update.

examples/batch-tx.js
```diff
@@ -0,0 +1,379 @@
...
+     * Step 4:
+     * Execute the batch transaction
+     */
+
+    console.log("Executing batch transaction...");
+    const batch = await (
+        await (
+            await new BatchTransaction()
+                .addInnerTransaction(aliceBatchedTransfer)
+                .addInnerTransaction(bobBatchedTransfer)
```
SimiHunjan, line 166: Don't we need to set `setBatchKey(Key)` here?

SimiHunjan: and its not on the batch transaction it is on the inner transaction so nvm thanks!

src/transaction/TransactionResponse.js
```diff
@@ -170,10 +214,34 @@ export default class TransactionResponse {
...
+     *
+     * @param {Client} client
+     * @returns {Promise<TransactionReceipt>}
+     */
+    _retryTransaction(client) {
+        if (!client.operatorAccountId) {
+            throw new Error("Operator account is not set");
+        }
+
+        this.transactionId = TransactionId.generate(client.operatorAccountId);
```
ivaylonikolov7, line 227: what if the transaction fails. but the transaction id was not generated from the client?

src/transaction/BatchTransaction.js
```diff
@@ -0,0 +1,182 @@
...
+ */
+
+/**
+ * @typedef {import("../channel/Channel.js").default} Channel
+ * @typedef {import("../transaction/TransactionId.js").default} TransactionId
+ * @typedef {import("../account/AccountId.js").default} AccountId
+ * @typedef {import("../Key.js").default} Key
+ */
+
+export default class BatchTransaction extends Transaction {
```
0xivanov, line 20: Add docs, can copy from the PR in java

src/transaction/BatchTransaction.js
```diff
@@ -0,0 +1,182 @@
...
+
+            const fromProtobuf = TRANSACTION_REGISTRY.get(txType);
+            if (!fromProtobuf) {
+                throw new Error("fromProtobuf not found");
+            }
+
+            return fromProtobuf([], [tx], [], [], [txBody]);
+        });
+
+        return Transaction._fromProtobufTransactions(
```
0xivanov, line 119: Try to use the `fromBytes` method, since it should be more robust. Can see how it's done in go and java.

src/transaction/BatchTransaction.js
```diff
@@ -0,0 +1,182 @@
...
+    _execute(channel, request) {
+        return channel.util.atomicBatch(request);
+    }
+
+    /**
+     * @param {Transaction} tx
+     * @throws {Error} If the transaction is a batch or freeze transaction
+     */
+    _validateTransaction(tx) {
+        if (tx instanceof BatchTransaction || tx instanceof FreezeTransaction) {
```
0xivanov, line 170: Should also check if the inner txn is frozen and has batch key set.

src/transaction/Transaction.js
```diff
@@ -409,6 +425,19 @@ export default class Transaction extends Executable {
         return SCHEDULE_CREATE_TRANSACTION[0]()._setScheduledTransaction(this);
     }
 
+    /**
+     * Batchify the transaction
```
0xivanov, line 429: Add more docs for this method

src/transaction/Transaction.js
```diff
@@ -1181,6 +1213,24 @@ export default class Transaction extends Executable {
         );
     }
 
+    /**
+     *
```
0xivanov, line 1217: Add more docs for this method

src/transaction/TransactionResponse.js
```diff
@@ -170,10 +214,34 @@ export default class TransactionResponse {
         };
     }
 
+    /**
+     *
+     * @param {Client} client
+     * @returns {Promise<TransactionReceipt>}
+     */
+    _retryTransaction(client) {
```
0xivanov, line 222: This method should do `transaction.execute()`. You are trying to get the receipt for a `transactionId` that has n't been submitted to the network - it would result in `RECEIPT_NOT_FOUND`.
You can keep an instance of the `transaction` object in the `TransactionResponse` as a field. This `transaction` should be set in the executable flow for every single execute, just in case the txn throttles. You shoould be able to  re-execute it here if so.

test/unit/BatchTransaction.js
```diff
@@ -0,0 +1,135 @@
...
+    AccountId,
+    PrivateKey,
+    Timestamp,
+    TransactionId,
+    BatchTransaction,
+    Transaction,
+} from "../../src/index.js";
+import { expect } from "chai";
+
+describe("BatchTransaction", function () {
```
0xivanov, line 12: Add tests for the validation of inner transactions.

examples/batch-tx.js
```diff
@@ -0,0 +1,379 @@
...
+    const operatorAccId = AccountId.fromString(process.env.OPERATOR_ID);
+    const operatorPrivKey = PrivateKey.fromStringED25519(
+        process.env.OPERATOR_KEY,
+    );
+
+    const client = Client.forName(process.env.HEDERA_NETWORK)
+        .setOperator(operatorAccId, operatorPrivKey)
+        .setLogger(new Logger(LogLevel.Silent));
+
+    //await executeBatchWithBatchify(client);
```
0xivanov, line 43: note: commented out
also it would be great to add 2 comments or logs explaining what the 2 methods showcase.

examples/batch-tx.js
```diff
@@ -0,0 +1,379 @@
...
+            client.getOperator().accountId,
+            Hbar.from(1, HbarUnit.Hbar),
+        )
+        .batchify(carolClient, batchKey3);
+
+    console.log("Created third account (Carol): " + carol.toString());
+
+    /**
+     * Step 3:
+     * Get the balanace in order to compete after this batch
```
ivaylogarnev-limechain, line 141:
```suggestion
     * Get the balance in order to compare after this batch
```

src/transaction/BatchTransaction.js
```diff
@@ -0,0 +1,182 @@
...
+    get innerTransactionIds() {
+        return this._batchTransactions.map((tx) => tx.transactionId);
+    }
+
+    /**
+     *
+     * @returns {proto.AtomicBatchTransactionBody}
+     */
+    _makeTransactionData() {
+        const signedTransactionBytes = this._batchTransactions.map((tx) =>
```
ivaylogarnev-limechain, line 70: Shouldn't we make sure that `this._batchTransactions` is not empty.

src/transaction/Transaction.js
```diff
@@ -1843,6 +1897,17 @@ export default class Transaction extends Executable {
         };
     }
 
+    /**
+     * @override
+     * @returns {boolean}
+     */
+    isBatchedAndNotBatchTransaction() {
```
ivaylogarnev-limechain, line 1904: Wouldn't it be clearer if we simply kept the title as `isBatchedTransaction`?

src/transaction/TransactionResponse.js
```diff
@@ -170,10 +214,34 @@ export default class TransactionResponse {
...
+
+    /**
+     * Cross-platform delay function that works in both Node.js and browsers
+     * @param {number} ms - Milliseconds to delay
+     * @returns {Promise<void>}
+     * @private
+     */
+    _delay(ms) {
+        return new Promise((resolve) => setTimeout(resolve, ms));
+    }
```
ivaylogarnev-limechain, line 246: Can’t we use the `wait` function from the `util` directory here?

src/transaction/Transaction.js
ivaylonikolov7: it checks both if the tx is batch transaction and if the key is set, I am open to names that could shorten the function name. Couldn't think of a better name too.

src/query/Query.js
```diff
@@ -524,4 +524,13 @@ export default class Query extends QueryBase {
...
         return HieroProto.proto.Response.encode(response).finish();
     }
+
+    /**
+     * @protected
+     * @override
+     * @returns {boolean}
+     */
+    isBatchedAndNotBatchTransaction() {
+        return false;
```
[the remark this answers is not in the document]
ivaylonikolov7: This will always be false for the `Queries`.

examples/batch-tx.js
```diff
@@ -0,0 +1,378 @@
...
+
+    await executeBatchWithManualInnerTransactionFreeze(client);
+    client.close();
+}
+
+/**
+ *
+ * @param {Client} client
+ */
+async function executeBatchWithBatchify(client) {
```
0xivanov, line 51: Looks unused.

.github/workflows/build.yml
```diff
@@ -118,7 +118,7 @@ jobs:
         id: start-local-node
         if: ${{ steps.build-sdk.conclusion == 'success' && !cancelled() && always() }}
         run: |
-          ${{ env.CG_EXEC }} npx @hashgraph/hedera-local start -d -—network local --balance=100000 --network-tag=0.60.0-alpha.0
+          ${{ env.CG_EXEC }} npx @hashgraph/hedera-local start -d -—network local --balance=100000 --network-tag=0.61.1
```
0xivanov, line 121:
```suggestion
          ${{ env.CG_EXEC }} npx @hashgraph/hedera-local start -d -—network local --balance=100000 --network-tag=0.61.4
```

.github/workflows/build.yml
```diff
@@ -210,7 +210,7 @@ jobs:
         id: start-local-node
         if: ${{ !cancelled() && always() }}
         run: |
-          ${{ env.CG_EXEC }} npx @hashgraph/hedera-local start -d -—network local --balance=100000 --network-tag=0.60.0-alpha.0
+          ${{ env.CG_EXEC }} npx @hashgraph/hedera-local start -d -—network local --balance=100000 --network-tag=0.61.1
```
0xivanov, line 213:
```suggestion
          ${{ env.CG_EXEC }} npx @hashgraph/hedera-local start -d -—network local --balance=100000 --network-tag=0.61.4
```

0xivanov (commented):
Looks good

src/transaction/BatchTransaction.js
```diff
@@ -0,0 +1,267 @@
...
+            if (!txType) {
+                throw new Error("Transaction type not found");
+            }
+
+            const fromProtobuf = TRANSACTION_REGISTRY.get(txType);
+            if (!fromProtobuf) {
+                throw new Error("fromProtobuf not found");
+            }
+
+            return fromProtobuf([], [tx], [], [], [txBody]);
```
ivaylogarnev-limechain, line 191: Could you please add description for this return?

examples/batch-tx.js
```diff
@@ -0,0 +1,382 @@
...
+        ).getReceipt(client)
+    ).accountId;
+
+    const aliceBatchedTransfer = await new TransferTransaction()
+        .addHbarTransfer(alice, Hbar.from(-1, HbarUnit.Hbar))
+        .addHbarTransfer(
+            client.getOperator().accountId,
+            Hbar.from(1, HbarUnit.Hbar),
+        )
+        .setBatchKey(batchKey2)
```
venilinvasilev, line 252: I think you should use batchKey1 here ...?

examples/batch-tx.js
```diff
@@ -0,0 +1,382 @@
...
+}
+
+/**
+ *
+ * @param {Client} client
+ */
+async function executeBatchWithBatchify(client) {
+    /**
+     * Step 2:
+     * Create three account create transactions with batch keys, but do not execute them.
```
venilinvasilev, line 55: Wouldn't it be better if the example is based on the user story cases defined in the actual HIP instead of having these crypto transfers defined in a batch?

ivaylonikolov7: I think the point was to create an example based on how the users will use the HIP. Considering this is already merged in the other SDKs and will have to be changed in the other SDKs not sure if its worth the effort.

0xivanov: The code comments are confusing.

0xivanov (commented):
The steps in the example are confusing, please check out https://github.com/hiero-ledger/hiero-sdk-go/blob/main/examples/batch_transaction/main.go

venilinvasilev (dismissed):
LGTM

ivaylogarnev-limechain (approved):
LGTM

ivaylogarnev-limechain (approved):

venilinvasilev (approved):
LGTM

PavelSBorisov (approved):

PR #3051, opened 2025-04-15
feat: extend set key with alias to work with public ecdsa keys

venilinvasilev:
This PR builds upon the previous enhancements to `AccountCreateTransaction`, expanding support and improving developer experience when setting keys and deriving EVM aliases.

---

###  What's New

- `setECDSAKeyWithAlias` and `setKeyWithAlias` now accept both `ECDSA PrivateKey` **and** `ECDSA PublicKey` inputs.
  - If a `PrivateKey` is provided, the SDK will derive the alias from its corresponding public key.
  - If a `PublicKey` is provided, the alias will be derived directly from that key.

This enables secure frontend-backend workflows (e.g., wallet creation flows), where only the public key is available on the backend — without requiring casting or deprecated APIs.

---

#### AccountCreateTransaction
- **`setECDSAKeyWithAlias(ECDSAKey)`**  
  Sets an ECDSA key (either private or public), and automatically derives and sets the corresponding EVM alias.

- **`setKeyWithAlias(Key, ECDSAKey)`**  
  Sets the account key and derives the alias from a separate ECDSA key (private or public).  
  Both keys must sign the transaction for it to succeed.

- **`setKeyWithoutAlias(Key)`**  
  Explicitly sets the key without deriving or attaching an alias.

#### PublicKey
- **`type`**
This getter exposes a string literal describing the key type (e.g., "secp256k1") for the PublicKey class.
Useful for runtime validation and for operations that depend on the specific key type.

---

### Related Issue(s)

- https://github.com/hiero-ledger/hiero-sdk-js/issues/3048

---

### Original Feature Issue:
https://github.com/hiero-ledger/hiero-sdk-js/issues/2795
### Original Feature PR:
https://github.com/hiero-ledger/hiero-sdk-js/pull/2834

### Checklist

- [x] Extended support for **both public and private ECDSA keys**
- [x] Updated `AccountCreateTransaction` key-setting APIs
- [x] Updated all relevant usage to the new APIs
- [x] Added unit and integration tests
- [x] Added usage examples in the documentation

---

This update improves security alignment and simplifies integration across frontend/backend setups and wallet providers.

### DIFF
--- a/src/PublicKey.js
+++ b/src/PublicKey.js
@@ -261,6 +261,13 @@ export default class PublicKey extends Key {
     }
 
     /**
+     * @returns {string}
+     */
+    get type() {
+        return this._key._type;
+    }
+
+    /**
      * @param {Long | number} shard
      * @param {Long | number} realm
      * @returns {AccountId}
--- a/src/account/AccountCreateTransaction.js
+++ b/src/account/AccountCreateTransaction.js
@@ -589,11 +589,11 @@ export default class AccountCreateTransaction extends Transaction {
             return key.publicKey.toEvmAddress();
         } else if (isPublicECDSAKey) {
             return key.toEvmAddress();
-        } else {
-            throw new Error(
-                "Invalid key for alias derivation provided: expected an ECDSA (secp256k1) PrivateKey or PublicKey.",
-            );
         }
+
+        throw new Error(
+            "Invalid key for alias derivation provided: expected an ECDSA (secp256k1) PrivateKey or PublicKey.",
+        );
     }
 
     /**
--- a/src/account/AccountCreateTransaction.js
+++ b/src/account/AccountCreateTransaction.js
@@ -13,7 +13,7 @@ import Long from "long";
 import Key from "../Key.js";
 import PrivateKey from "../PrivateKey.js";
 import EvmAddress from "../EvmAddress.js";
-
+import PublicKey from "../PublicKey.js";
 /**
  * @typedef {import("bignumber.js").default} BigNumber
  * @typedef {import("../channel/Channel.js").default} Channel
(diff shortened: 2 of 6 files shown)


src/account/AccountCreateTransaction.js
```diff
@@ -580,6 +571,31 @@ export default class AccountCreateTransaction extends Transaction {
...
+        const isPrivateECDSAKey =
+            key instanceof PrivateKey && key.type === "secp256k1";
+        const isPublicECDSAKey =
+            key instanceof PublicKey && key.type === "secp256k1";
+
+        if (isPrivateECDSAKey) {
+            return key.publicKey.toEvmAddress();
+        } else if (isPublicECDSAKey) {
+            return key.toEvmAddress();
+        } else {
```
ivaylogarnev-limechain, line 592: no need of else here since we have early returns

venilinvasilev: Where do we have early returns. I do not get it?

0xivanov: It would throw the error even if the last else is not present

src/PublicKey.js
```diff
@@ -260,6 +260,13 @@ export default class PublicKey extends Key {
         }
     }
 
+    /**
+     * @returns {string}
+     */
+    get type() {
```
0xivanov, line 266: Do we need this getter? Is it a new API?

0xivanov (commented):
Looks good

src/PublicKey.js
venilinvasilev: It seems it is a new API for the public key.
I exposed this so I can check if a key is of type public ECDSA.
I've added it to the PR description.

ivaylonikolov7 (approved):

0xivanov (approved):

PR #3053, opened 2025-04-15
chore: remove client constants

ivaylonikolov7:
**Description**:
Removed myhbarwallet gRPC web proxies

**Notes for reviewer**:

**Checklist**

Tested the react native changes using a testnet client on an expo app.

### DIFF
--- a/src/constants/ClientConstants.js
+++ b/src/constants/ClientConstants.js
@@ -3,7 +3,6 @@ import AccountId from "../account/AccountId.js";
 
 // MAINNET node proxies are the same for both 'WebClient' and 'NativeClient'
 export const MAINNET = {
-    "https://grpc-web.myhbarwallet.com:443": new AccountId(3),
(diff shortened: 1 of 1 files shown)


ivaylogarnev-limechain (approved):
LGTM

src/constants/ClientConstants.js
```diff
@@ -56,11 +55,11 @@ export const WEB_PREVIEWNET = {
 };
 
 export const NATIVE_TESTNET = {
```
0xivanov, line 57: We don't have `NATIVE_TESTNET`?
It's weird to me that we have separate const for react and not using the web consts

ivaylonikolov7: `// MAINNET node proxies are the same for both 'WebClient' and 'NativeClient'`

There's a comment explaining this.

0xivanov: Can't we use the web proxies for react and remove the NATIVE_ consts?

PR #3068, opened 2025-04-24
feat: key algorithm identifier

### LINKED ISSUE #2170: Implement method to determine which encryption algorithm is used for key generation - `ECDSA` or `ED25519`
Having the private/public key It would be beneficial for users to be able to determine which type of encryption algorithm (`ECDSA`,`ED25519`) was used for the key generation. 

How is this possible in cryptography:
1. Having the private key we can extract the type of algorithm that is used for generation.
2. By having the private key in DER  encoded format.
3. Then decoding it using a [ASN1 decoder](https://the-x.cn/en-US/encodings/Asn1.aspx) (just an example tool) most likely a third party library for JS project
4. Then after decoding, it will return whether the key is `secp256k1` or `curveEd25519`

For the purpose
1. Good research on what third library should be used
    1. One possible solution is `asn1js` library would be for the decoding 
    2. In Java SDK `bouncycastle` is used. A check can be done if there's alternative for JavaScript. 
        1. Example in **Java SDK**: [fromBytesDER](https://github.com/hashgraph/hedera-sdk-java/blob/546b49b8f95387666a99800e344153c3c3dddf8f/sdk/src/main/java/com/hedera/hashgraph/sdk/PrivateKey.java#L227) 
        2. Example in **C++ SDK**: [decode](https://github.com/hashgraph/hedera-sdk-cpp/blob/5030f4768d88854f277f4904c172ceb6c91370cc/src/sdk/main/include/impl/ASN1Object.h#L61) 
2. Tests with a lot of different keys (at least hundred).
    1. Same/different byte lengths
    2. Same/different types  (ECDSA, ED25519, RSA, Blowfish, Twofish, DES, TripleDES)

ivaylonikolov7:
**Description**:
- adds a library that determines the keys 
- updates the files to use the correct implementation when they have to determine the key type 

Fixes #2170 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/src/PrivateKey.js
+++ b/src/PrivateKey.js
@@ -1,15 +1,16 @@
 // SPDX-License-Identifier: Apache-2.0
-
+import { hexlify } from "@ethersproject/bytes";
 import { PrivateKey as PrivateKeyCrypto } from "@hashgraph/cryptography";
+import { proto } from "@hashgraph/proto";
 import Mnemonic from "./Mnemonic.js";
 import PublicKey from "./PublicKey.js";
 import Key from "./Key.js";
 import CACHE from "./Cache.js";
 import SignatureMap from "./transaction/SignatureMap.js";
-
 import AccountId from "./account/AccountId.js";
 import TransactionId from "./transaction/TransactionId.js";
-import { proto } from "@hashgraph/proto";
+
+import { asn1DecodeStringDer } from "./util/ASN1-Decoder.js";
 
 /**
  * @typedef {import("./transaction/Transaction.js").default} Transaction
@@ -99,6 +100,18 @@ export default class PrivateKey extends Key {
      * @returns {PrivateKey}
      */
     static fromBytes(data) {
+        const keyString = hexlify(data);
+        if (PrivateKey.isDerKey(keyString)) {
+            if (PrivateKey.getAlgorithm(keyString) === "ecdsa") {
+                return new PrivateKey(PrivateKeyCrypto.fromBytesECDSA(data));
+            } else {
+                return new PrivateKey(PrivateKeyCrypto.fromBytesED25519(data));
+            }
+        }
+
+        // If the key is not DER, we assume it's a raw private key
+        // this will default to ED25519 as we dont have a way
+        // to determine the type of the key based on the raw bytes
         return new PrivateKey(PrivateKeyCrypto.fromBytes(data));
     }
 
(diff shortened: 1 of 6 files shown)


test/integration/LocalWalletTest.js
```diff
@@ -1,12 +1,17 @@
...
+    PrivateKey,
+    AccountId,
+} from "../../src/index.js";
 
 describe("LocalWallet", function () {
     it("can fetch wallet's info", async function () {
-        const wallet = new Wallet(
-            process.env.OPERATOR_ID,
+        const operatorId = AccountId.fromString(process.env.OPERATOR_ID);
+        const operatorKey = PrivateKey.fromStringED25519(
```
ivaylogarnev-limechain, line 11: What if the user key is in ECDSA format? Wouldn't it be better to use the reworked `fromStringDer` function here?

ivaylogarnev-limechain:
I think it would be beneficial to add such a checks for the `PrivateKey.fromStringED25519/fromStringECDSA` methods as well.

test/integration/LocalWalletTest.js
ivaylonikolov7: Our integration tests don't work with ECDSA unless we change the BaseIntegrationTestEnv anyway. 

image

venilinvasilev:
Looks good, but how about adding tests to cover the new functionallity?

src/PrivateKey.js
```diff
@@ -480,6 +509,43 @@ export default class PrivateKey extends Key {
...
+    static getAlgorithm(privateKey) {
+        const decoder = new ASN1Decoder(Uint8Array.from(decode(privateKey)));
+        decoder.read();
+        const supportedKeyTypes = ["ecdsa", "ed25519"];
+        const decodedKeyType = decoder.getOidKeyTypes()[0];
+
+        if (!PrivateKey.isDerKey(privateKey)) {
+            throw new Error("Only der keys are supported");
+        }
+        if (!supportedKeyTypes.includes(decodedKeyType)) {
```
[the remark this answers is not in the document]
ivaylonikolov7: Yeah but if someone uses it incorrectly it might be a breaking change for him. :(

ivaylonikolov7:
> Looks good, but how about adding tests to cover the new functionallity?

we already have test vectors in `test/unit/ecdsaprivatekey.js` and `test/unit/ed25519privatekey`.

Did you mean for the `Asn1Decoder.js`?

venilinvasilev:
> > Looks good, but how about adding tests to cover the new functionallity?
> 
> we already have test vectors in `test/unit/ecdsaprivatekey.js` and `test/unit/ed25519privatekey`.
> 
> Did you mean for the `Asn1Decoder.js`?

I mean for the new methods in the PrivateKey class (and Asn1Decoder if applicable).

venilinvasilev (approved):
LGTM

PR #3092, opened 2025-05-08
feat: HIP-1046 

ivaylonikolov7:
### Added
- `WebClient.forMirrorNetwork()` factory method now works and correctly initializes and updates the address book during initialization.
- `WebClient` now supports scheduled network updates via the `scheduleNetworkUpdate: true` option, and manual updates using the `updateNetwork()` method.
- `AddressBookQuery` is now supported in browser environments.
- `MirrorNodeContractCallQuery` is now supported in browser environments.
- `MirrorNodeContractЕstimateQuery` is now supported in browser environments.
**Related issue(s)**:
[3102](https://github.com/hiero-ledger/hiero-sdk-js/issues/3102)
[3148
](https://github.com/hiero-ledger/hiero-sdk-js/issues/3148)

Fixes #
[3102](https://github.com/hiero-ledger/hiero-sdk-js/issues/3102)
[3148
](https://github.com/hiero-ledger/hiero-sdk-js/issues/3148)

**Checklist**

- [] Tested (unit, integration, etc.)


src/network/AddressBookQueryWeb.js
```diff
@@ -0,0 +1,334 @@
...
+ *   address: string,
+ *   service_endpoints: {
+ *     domain_name: string,
+ *     ip_address_v4: string,
+ *     port: number
+ *   }[],
+ *   description: string,
+ *   stake: number
+ * }>} nodes
+ */
```
ivaylogarnev-limechain, line 46: Don't we have interface for that?

src/channel/WebChannel.js
```diff
@@ -39,11 +39,14 @@ export default class WebChannel extends Channel {
         // eslint-disable-next-line @typescript-eslint/no-misused-promises
         return async (method, requestData, callback) => {
             try {
+                const address = this._address.includes("http")
+                    ? this._address
+                    : `http://${this._address}`;
```
venilinvasilev: What do you think about this should we instead add a check if this is in local environment to add http and use https if not?

src/client/WebClient.js
```diff
@@ -156,10 +160,43 @@ export default class WebClient extends Client {
...
+        for (const nodeAddress of addressBook.nodeAddresses) {
+            for (const endpoint of nodeAddress.addresses) {
+                if (nodeAddress.accountId != null) {
+                    console.log({ endpoint });
+                    network[endpoint.toString()] = nodeAddress.accountId;
+                }
+            }
+        }
+
+        client.setNetwork(network);
```
venilinvasilev: I cannot use client.network.setNetworkFromAddressBook because of underlying logic which strips node addresses with ports different than 50211 and 50212.
The approach proposed from @ivaylonikolov7 with branching the Network for browser env as WebNetwork is not applicable and won't work in this case so that's why I've made the network mapping here. WDYT?

src/client/WebClient.js
```diff
@@ -196,6 +233,46 @@ export default class WebClient extends Client {
...
+
+            for (const nodeAddress of addressBook.nodeAddresses) {
+                for (const endpoint of nodeAddress.addresses) {
+                    if (nodeAddress.accountId != null) {
+                        network[endpoint.toString()] = nodeAddress.accountId;
+                    }
+                }
+            }
+
+            this.setNetwork(network);
```
venilinvasilev: Same concern as my above comment.

src/client/WebClient.js
```diff
@@ -185,17 +205,75 @@ export default class WebClient extends Client {
...
+    async updateNetwork() {
+        if (this._isUpdatingNetwork) {
+            return;
+        }
+
+        this._isUpdatingNetwork = true;
+
+        try {
+            const addressBook = await new AddressBookQuery()
+                .setFileId(FileId.ADDRESS_BOOK)
```
0xivanov: FYI, This would need to use the shard and realms

src/network/AddressBookQueryWeb.js
```diff
@@ -0,0 +1,341 @@
...
+                        } during attempt ${
+                            this._attempt
+                        }. Waiting ${delay} ms before next attempt: ${message}`,
+                    );
+                }
+
+                this._attempt += 1;
+
+                setTimeout(() => {
+                    void this._makeFetchRequest(
```
0xivanov: I don't really like the idea to make this recursive. A simple for loop using the attemps as index would do the job.

src/query/MirrorNodeContractEstimateQuery.js
```diff
@@ -12,7 +12,7 @@ import MirrorNodeContractQuery from "./MirrorNodeContractQuery.js";
  * to execute the same call on the main network. It's useful for determining the appropriate
  * gas limit before submitting an actual transaction.
  */
-export default class MirrorNodeContractCallQuery extends MirrorNodeContractQuery {
+export default class MirrorNodeContractEstimateQuery extends MirrorNodeContractQuery {
```
0xivanov, line 15: interesting.. and nice catch

src/client/WebClient.js
venilinvasilev: Thanks for flagging this. It uses now.

src/address_book/IPv4Address.js
```diff
@@ -90,6 +90,28 @@ export default class IPv4Address {
...
+     */
+    static _fromString(address) {
+        const parts = address.split(".");
+        if (parts.length !== 4) {
+            throw new Error("Invalid IPv4 address format");
+        }
+
+        const network = new IPv4AddressPart()
+            .setLeft(parseInt(parts[0], 10))
+            .setRight(parseInt(parts[1], 10));
```
ivaylonikolov7: isn't 10 the default one?

src/client/Client.js
```diff
@@ -130,7 +130,6 @@ export default class Client {
         this._requestTimeout = null;
 
         /**
-         * @private
```
ivaylonikolov7, line 133: Why expose this? And if we are going to expose this wouldn't it be better to do so by using a getter?

src/native.js
```diff
@@ -7,3 +7,4 @@
 export * from "./exports.js";
 
 export { default as Client } from "./client/NativeClient.js";
+export { default as AddressBookQuery } from "./network/AddressBookQuery.js";
```
ivaylonikolov7, line 10: Did some quick search and found that react native doesn't support gRPC requests too. Can we have `AddressBookQueryWeb` here too?

src/channel/WebChannel.js
ivaylonikolov7: I think this is fine. Local environment will always use http, and production will use https.

src/constants/ClientConstants.js
```diff
@@ -3,63 +3,63 @@ import AccountId from "../account/AccountId.js";
...
+    "node06.swirldslabs.com:443": new AccountId(9),
+    "node07.swirldslabs.com:443": new AccountId(10),
+    "node09.swirldslabs.com:443": new AccountId(12),
+    "node10.swirldslabs.com:443": new AccountId(13),
+    "node11.swirldslabs.com:443": new AccountId(14),
+    "node12.swirldslabs.com:443": new AccountId(15),
+    "node14.swirldslabs.com:443": new AccountId(17),
+    "node15.swirldslabs.com:443": new AccountId(18),
+    "node16.swirldslabs.com:443": new AccountId(19),
+    "node17.swirldslabs.com:443": new AccountId(20),
```
ivaylonikolov7, line 21: Can we make it to keep the `https://` in front of the address. I am worried that this might break the SDK somewhere else. I remember you had some issues with the `fetch` basically calling `http://http://" but seems like we could create some unnecessary headaches for us with this change.

src/native.js
venilinvasilev: Thanks for researching and flagging this. I think this would be great addition, but I would separate this change from this PR as it is quite big already. Furthermore this needs to be tested on Native environment which would slow down the process of getting these changes to being merged. WDYT? Would you like to create an issue for this?

src/constants/ClientConstants.js
venilinvasilev: We need to keep these like this, because this is the way that we are getting the addresses/domains from the address book query (without the "https://" schema). If I leave them here then I need to add some checks to transform the data for when we set the list from the hardcoded values including the schema and not when we are getting the list from the addressbok query response. Lets keep these consistent I will test this thoroughly.

src/network/AddressBookQueryWeb.js
venilinvasilev: No. This is the response schema defined from the mirror rest API

src/address_book/IPv4Address.js
venilinvasilev: Actually I will use Number instead

src/client/Client.js
venilinvasilev: We're exposing this internally for the sub-classes not for the end-user that's why it starts with "_"

src/network/AddressBookQueryWeb.js
venilinvasilev: Removed recursion

examples/frontend-examples/src/app/client/grpc-web-proxy/page.js
```diff
@@ -0,0 +1,350 @@
...
+    const [clientNetwork, setClientNetwork] = useState(initialNetwork);
+    const [isLoading, setIsLoading] = useState(false);
+    const [message, setMessage] = useState("");
+    const [messageType, setMessageType] = useState("info");
+    const [createdAccount, setCreatedAccount] = useState(null);
+
+    const client = useMemo(() => {
+        return WebClient.forNetwork(initialNetwork).setOperator(
+            AccountId.fromString(process.env.NEXT_PUBLIC_OPERATOR_ID),
+            PrivateKey.fromStringDer(process.env.NEXT_PUBLIC_OPERATOR_KEY),
```
ivaylonikolov7, line 44: We already have ` PrivateKey.fromStringED25519(process.env.NEXT_PUBLIC_OPERATOR_KEY),`

Are we going to stick to the already implemented way of setting up the client, wdyt? 

I can see your point that by using `fromStringDer` we could use both ECDSA and ED25519 key but we already have implemented one way. 

My opinion is that your way of handling this is better but we might have to change the setting of the client in the other page too in this PR?

venilinvasilev (dismissed):

venilinvasilev:
LGTM

src/client/WebClient.js
```diff
@@ -201,17 +215,73 @@ export default class WebClient extends Client {
...
+                    break;
+                case "testnet":
+                    this._mirrorNetwork.setNetwork(MirrorNetwork.TESTNET);
+                    break;
+                case "mainnet":
+                    this._mirrorNetwork.setNetwork(MirrorNetwork.MAINNET);
+                    break;
+                default:
+                    this._mirrorNetwork.setNetwork([mirrorNetwork]);
+            }
```
0xivanov, line 235: Can you extract this logic in the parent `Client.js`?

src/network/AddressBookQueryWeb.js
```diff
@@ -0,0 +1,379 @@
...
+ * or from the most recent address book if no file ID is specified. The response
+ * contains node metadata including IP addresses and ports for both node and mirror
+ * node services.
+ * @augments {Query<NodeAddressBook>}
+ */
+export default class AddressBookQueryWeb extends Query {
+    /**
+     * @param {object} props
+     * @param {FileId | string} [props.fileId]
+     * @param {number} [props.limit]
```
0xivanov, line 68: Could document what this limit does. Here and in the setter/getter.

src/network/AddressBookQueryWeb.js
```diff
@@ -0,0 +1,379 @@
...
+                        );
+                    }
+
+                    // Wait before next attempt
+                    await new Promise((resolve) => setTimeout(resolve, delay));
+                    continue;
+                }
+
+                // If we shouldn't retry or have exhausted attempts, reject
+                reject(new Error("failed to query address book"));
```
0xivanov: Could specify that we reached max attempts

test/unit/browser/WebClient.js
```diff
@@ -0,0 +1,1560 @@
...
+            const updatedNetwork = client.network;
+
+            const initialEntries = createNetworkAddressNodeSet(initialNetwork);
+            const updatedEntries = createNetworkAddressNodeSet(updatedNetwork);
+
+            expect(initialEntries).not.toEqual(updatedEntries);
+            expect(updatedEntries.size).toBe(0);
+        });
+
+        it("should change network to empty when mirror response has only nodes with empty domain_name in grpc_proxy_endpoint", async function () {
```
0xivanov, line 1467: Add a test where we have `forNetwork` and have response containing 1 node with valid proxy and 1 node with null/empty domain name proxy.
What would happen?

0xivanov: I think the result should be 1 node in the network. Could be tested with `forMainnet/Testnet`,etc.

src/client/WebClient.js
venilinvasilev: Good observation. Seems that NativeClient still has nothing to do with MirrorNetwork.
I would extract this when we add MirrorNetwork support to the NativeClient (maybe we need an issue for this).

test/unit/browser/WebClient.js
venilinvasilev: Nice catch. Seems that indeed I have missed this one!

venilinvasilev: It should have them now.

src/client/WebClient.js
venilinvasilev: I've created an issue for this: https://github.com/hiero-ledger/hiero-sdk-js/issues/3205

0xivanov (approved):
LGTM

0xivanov (commented):
Do not merge

0xivanov (approved):

PR #3133, opened 2025-05-29
feat: add shard and realm pesistence to the client

venilinvasilev:
**Description**:
Adds persistent realm and shard according to https://github.com/hiero-ledger/sdk-collaboration-hub/pull/28

### Changed
- `ClientConfiguration` type updated: The `Client` constructor now accepts `shard` and `realm` parameters directly.  
- `Client.forMirrorNetwork` now supports two additional arguments: `shard` and `realm`.
### Internal
- When a custom network is provided in the `Client` constructor, the `shard` and `realm` values are now automatically inferred from the specified network.
- An error is now thrown if the nodes defined in a custom network are not all in the same shard and realm, enforcing consistency and preventing misconfiguration.

Fixes https://github.com/hiero-ledger/hiero-sdk-js/issues/3153

Notes for reviewer:

Checklist

- [x]  Documented (Code comments, README, etc.)
- [x]  Tested (unit, integration, etc.)

### DIFF
--- a/src/client/Client.js
+++ b/src/client/Client.js
@@ -802,7 +802,7 @@ export default class Client {
      *
      * @param {{[key: string]: (string | AccountId)}} network
      */
-    static validateNetworkConsistency(network) {
+    static _validateNetworkConsistency(network) {
         let shard = undefined;
         let realm = undefined;
 
@@ -815,7 +815,7 @@ export default class Client {
                 realm = nodeRealm;
             } else if (shard !== nodeShard || realm !== nodeRealm) {
                 throw new Error(
-                    `All nodes must be within the same shard and realm. Found nodes in shard ${String(shard)}.${String(realm)} and ${String(nodeShard)}.${String(nodeRealm)}`,
+                    "Network is not valid, all nodes must be in the same shard and realm",
                 );
             }
         }
@@ -829,7 +829,7 @@ export default class Client {
      * @param {{[key: string]: (string | AccountId)}} network
      * @returns {{shard: number, realm: number}}
      */
-    static extractShardRealm(network) {
+    static _extractShardRealm(network) {
         const entries = Object.entries(network);
         if (entries.length === 0) {
             return { shard: 0, realm: 0 };
--- a/src/client/Client.js
+++ b/src/client/Client.js
@@ -803,21 +803,34 @@ export default class Client {
      * @param {{[key: string]: (string | AccountId)}} network
      */
     static _validateNetworkConsistency(network) {
-        let shard = undefined;
-        let realm = undefined;
+        if (Object.keys(network).length === 0) {
+            return;
+        }
 
-        for (const accountId of Object.values(network)) {
+        const [, nodeAccountId] = Object.entries(network)[0];
+
+        const accountIdStr = nodeAccountId.toString();
+
+        const [firstNodeShard, firstNodeRealm] = accountIdStr
+            .split(".")
+            .map(Number);
+
+        const isNetworkValid = Object.values(network).every((accountId) => {
             const accountIdStr = accountId.toString();
-            const [nodeShard, nodeRealm] = accountIdStr.split(".").map(Number);
 
-            if (shard === undefined && realm === undefined) {
-                shard = nodeShard;
-                realm = nodeRealm;
-            } else if (shard !== nodeShard || realm !== nodeRealm) {
-                throw new Error(
-                    "Network is not valid, all nodes must be in the same shard and realm",
-                );
-            }
+            const [currentShard, currentRealm] = accountIdStr
+                .split(".")
+                .map(Number);
+            return (
+                currentShard === firstNodeShard &&
+                currentRealm === firstNodeRealm
+            );
+        });
+
+        if (!isNetworkValid) {
+            throw new Error(
+                "Network is not valid, all nodes must be in the same shard and realm",
+            );
         }
     }
 
@@ -835,7 +848,9 @@ export default class Client {
             return { shard: 0, realm: 0 };
         }
 
-        const accountIdStr = entries[0][1].toString();
+        const [, firstNodeAccountId] = entries[0];
+
+        const accountIdStr = firstNodeAccountId.toString();
         const [shard, realm] = accountIdStr.split(".").map(Number);
 
         return { shard, realm };
--- a/src/client/Client.js
+++ b/src/client/Client.js
@@ -34,10 +34,12 @@ import { convertToNumber } from "../util.js";
 
 /**
  * @typedef {object} ClientConfiguration
- * @property {{[key: string]: (string | AccountId)} | string} network
+ * @property {{[key: string]: (string | AccountId)} | string} [network]
  * @property {string[] | string} [mirrorNetwork]
  * @property {Operator} [operator]
  * @property {boolean} [scheduleNetworkUpdate]
+ * @property {number} [shard]
+ * @property {number} [realm]
  */
 
 /**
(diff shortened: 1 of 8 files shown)


src/client/NodeClient.js
```diff
@@ -167,38 +163,63 @@ export default class NodeClient extends Client {
...
-        const addressBook = await new AddressBookQuery()
-            .setFileId(FileId.ADDRESS_BOOK)
-            .execute(client);
+    /**
+     * @param {string[] | string} mirrorNetwork - The mirror network address or addresses.
+     * @param {number | ClientConfiguration} realmOrProps - The realm number for the address book query.
+     * @param {number} shard - The shard number for the address book query.
+     * @returns {Promise<NodeClient>} A configured NodeClient instance.
+     */
+    static async forMirrorNetwork(mirrorNetwork, realmOrProps, shard = 0) {
```
0xivanov: It's easier to understand what's going on if we have these methods separated

src/client/NodeClient.js
```diff
@@ -136,7 +135,7 @@ export default class NodeClient extends Client {
      * tried.
      *
      * @param {{[key: string]: (string | AccountId)}} network
-     * @param {ClientConfiguration} [props]
+     * @param {BaseClientConfiguration} [props]
```
0xivanov: `WebClient` and `NativeClient` seem to not have this param.

src/client/Client.js
```diff
@@ -715,7 +744,9 @@ export default class Client {
 
         try {
             const addressBook = await CACHE.addressBookQueryConstructor()
-                .setFileId(FileId.ADDRESS_BOOK)
+                .setFileId(
+                    FileId.getAddressBookFileIdFor(this._realm, this._shard),
```
0xivanov: review all the changes in this PR and make sure it's shard -> realm and not realm -> shard

venilinvasilev: Should be fixed

src/client/NodeClient.js
venilinvasilev: Fixed

src/client/Client.js
```diff
@@ -780,4 +796,48 @@ export default class Client {
...
+
+    /**
+     * Extracts shard and realm values from a network configuration.
+     * Note: This method assumes the network is consistent (all nodes in same shard/realm).
+     * Use validateNetworkConsistency() first to ensure this.
+     *
+     * @param {{[key: string]: (string | AccountId)}} network
+     * @returns {{shard: number, realm: number}}
+     */
+    static extractShardRealm(network) {
```
0xivanov: Shouldn't we mark `extractShardRealm` and `validateNetworkConsistency` as internal?

src/client/Client.js
```diff
@@ -780,4 +796,48 @@ export default class Client {
...
+        for (const accountId of Object.values(network)) {
+            const accountIdStr = accountId.toString();
+            const [nodeShard, nodeRealm] = accountIdStr.split(".").map(Number);
+
+            if (shard === undefined && realm === undefined) {
+                shard = nodeShard;
+                realm = nodeRealm;
+            } else if (shard !== nodeShard || realm !== nodeRealm) {
+                throw new Error(
+                    `All nodes must be within the same shard and realm. Found nodes in shard ${String(shard)}.${String(realm)} and ${String(nodeShard)}.${String(nodeRealm)}`,
```
0xivanov: nitpick:
```suggestion
                    `Network is not valid, all nodes must be in the same shard and realm`,
```

src/file/FileId.js
```diff
@@ -163,18 +190,3 @@ export default class FileId {
         );
     }
 }
-
-/**
- * The public node address book for the current network.
- */
-FileId.ADDRESS_BOOK = new FileId(102);
```
0xivanov: These might be public and if so - revert this change.

src/client/NodeClient.js
```diff
@@ -177,19 +189,18 @@ export default class NodeClient extends Client {
...
-        // Execute an address book query to get the network nodes
-        const addressBook = await new AddressBookQuery()
-            .setFileId(FileId.ADDRESS_BOOK)
-            .execute(client);
+    static async forMirrorNetwork(mirrorNetwork, shard, realm) {
+        const client = new NodeClient({
+            mirrorNetwork,
+            shard,
+            realm,
+        }).setNetworkUpdatePeriod(10_000);
```
ivaylonikolov7:
``` javascript
const DEFAULT_NETWORK_UPDATE_PERIOD=10_000
const client = new N...
.setNetworkUpdatePeriod(DEFAULT_NETWORK_UPDATE_PERIOD)
```

This 10_000 number doesn't make sense otherwise.

src/client/Client.js
```diff
@@ -780,4 +796,48 @@ export default class Client {
...
+     * @returns {{shard: number, realm: number}}
+     */
+    static _extractShardRealm(network) {
+        const entries = Object.entries(network);
+        if (entries.length === 0) {
+            return { shard: 0, realm: 0 };
+        }
+
+        const accountIdStr = entries[0][1].toString();
+        const [shard, realm] = accountIdStr.split(".").map(Number);
```
ivaylonikolov7:
```javascript
// we assume that all nodes are in the same shard so we can get any accountId
const firstNode=entries[0];
const accountId = firstNode[1]
```

wdyt?

src/client/Client.js
```diff
@@ -780,4 +796,48 @@ export default class Client {
...
+     *
+     * @param {{[key: string]: (string | AccountId)}} network
+     */
+    static _validateNetworkConsistency(network) {
+        let shard = undefined;
+        let realm = undefined;
+
+        for (const accountId of Object.values(network)) {
+            const accountIdStr = accountId.toString();
+            const [nodeShard, nodeRealm] = accountIdStr.split(".").map(Number);
```
ivaylonikolov7: Rename the variables to `currentShard, currentRealm` and rename `shard` and `realm` to `lastShard` and `lastRealm`??

src/file/FileId.js
```diff
@@ -30,6 +30,33 @@ export default class FileId {
         this._checksum = null;
     }
 
+    /**
+     * @param {number} shard
+     * @param {number} realm
+     * @returns {FileId}
+     */
+    static getAddressBookFileIdFor(shard = 0, realm = 0) {
+        return new FileId({ num: 102, shard, realm });
```
ivaylonikolov7, line 39: Is it safe to assume that the addressbook will always be on the 102 num? 

Do you think it would be ok to add a property that makes it so we can add a third optional property in case the addressbook is not on 102 num?

@0xivanov @venilinvasilev ?

ivaylonikolov7: same for `getExchangeRatesFileIdFor` and `getExchangeRatesFileIdFor`

src/client/Client.js
ivaylonikolov7: also [array.prototype.every() is a perfect fit here ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every) in my opinion. We can use the new features that JS offers us.

src/client/NodeClient.js
venilinvasilev: This was a magic string before I made any changes and this is not the default network update period the default one is 24 hours.
Maybe we need to discuss this if we should update so frequently

src/file/FileId.js
venilinvasilev: I guess file id will always be the same. Here we are addressing the shard and realm management.

ivaylonikolov7 (approved):

src/client/NativeClient.js
```diff
@@ -75,6 +75,17 @@ export default class NativeClient extends Client {
                         );
                 }
             } else if (props.network != null) {
+                Client._validateNetworkConsistency(props.network);
```
0xivanov, line 78: (Ignore the line I am commenting on)
I don't see change in the client.fromConfig method which should now accept shard and realm in it's schema.

venilinvasilev: The change is in the Client constructor it accepts now shard and realm and sets them, have a look in the super class.

PR #3176, opened 2025-06-23
feat: add solo action integration

### LINKED ISSUE #2757: Add CI Support for testing with the Solo Project
### Problem

-

### Solution

-

### Alternatives

_No response_

ivaylogarnev-limechain:
**Description**:
- Sets up solo action integration
- use 0.63.7 version of local node
- remove treasury, alice and bob accounts from exaamples `.env` because `solo` doesn't use auto generated accounts so we need to set them up
- disable long term schedule transactions and long term schedule tests
- use 0.0.1002 account instead of 0.0.2

Fixes #2757

**Notes for reviewer**:

**Checklist**


.github/workflows/build.yml
```diff
@@ -109,16 +101,21 @@ jobs:
         id: build-sdk
         run: task build
 
-      - name: Start the local node
-        id: start-local-node
-        if: ${{ steps.build-sdk.conclusion == 'success' && !cancelled() && always() }}
+      - name: Prepare Hiero Solo
+        id: solo
+        uses: hiero-ledger/hiero-solo-action@71219540ac7f578e6ea4fc3c17575c0295e56163 # v0.7
```
Ndacyayisenga-droid: The lastest `hiero-solo-action` version is `v0.10.0` https://github.com/hiero-ledger/hiero-solo-action/releases

.github/workflows/build.yml
```diff
@@ -213,21 +205,23 @@ jobs:
         id: build-sdk
         run: task build
 
-      - name: Start the local node
-        id: start-local-node
-        if: ${{ !cancelled() && always() }}
+      - name: Prepare Hiero Solo
+        id: solo
+        uses: hiero-ledger/hiero-solo-action@71219540ac7f578e6ea4fc3c17575c0295e56163 # v0.7
```
Ndacyayisenga-droid: The lastest `hiero-solo-action` version is `v0.10.0` https://github.com/hiero-ledger/hiero-solo-action/releases

test/integration/ScheduleCreateIntegrationTest.js
```diff
@@ -16,7 +16,11 @@ import {
 import IntegrationTestEnv from "./client/NodeIntegrationTestEnv.js";
 import { createAccount } from "./utils/Fixtures.js";
 
-describe("ScheduleCreate", function () {
+/**
+ * Skipped because Solo does not support schedule transactions
+ */
+// eslint-disable-next-line vitest/no-disabled-tests
+describe.skip("ScheduleCreate", function () {
```
0xivanov: it does not support long term scheduled. I think most of these tests should pass

.github/workflows/build.yml
```diff
@@ -213,21 +205,23 @@ jobs:
...
+        uses: hiero-ledger/hiero-solo-action@a39acf8cfbaa2feb195a86530d0ab643a45aa541 # v0.10
+        with:
+          installMirrorNode: true
+          hieroVersion: v0.63.7
+
+      - name: Set Operator Account
         run: |
-          ${{ env.CG_EXEC }} npx @hashgraph/hedera-local start -d -—network local --balance=100000 --network-tag=0.63.7
-          sleep 30
+          echo "OPERATOR_KEY=${{ steps.solo.outputs.ed25519PrivateKey }}" >> $GITHUB_ENV
```
venilinvasilev, line 217: Can't we use the started solo from previous step somehow. AFAIR the solo network is terminated after the runner completes. Why have 2 paralel networks or I am missing something?

ivaylonikolov7: image

Different jobs. I wish we could create the solo instance in the first instance and use the same `solo` instance in the one that runs the examples. Both of these jobs(running the tests and running the examples) can't use the same instance.  Its not an issue as they run in parallel. Running the examples after the tests is going to take more time so it's better for them to run in parallel.

venilinvasilev (dismissed):
LGTM

.github/workflows/build.yml
ivaylonikolov7: bumped to 0.10.0

.github/workflows/build.yml
ivaylonikolov7: bumped to 0.10.0

PR #3180, opened 2025-06-24
feat: example of polling mirror node to simulate TopicMessageQuery

ivaylonikolov7:
**Description**:
Polling mirror node rest api to stimulate TopicMessageQuery

**Notes for reviewer**:

**Checklist**


ivaylonikolov7:
@CodiumAI-Agent /improve

examples/polling-mirror-node.js
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * @param {EventEmitter} dataEmitter
+ * @param {string} topicId
+ * @returns {Promise<void>}
+ */
+async function pollUntilReady(dataEmitter, topicId) {
+    let lastMessagesLength = 0;
+
+    // eslint-disable-next-line no-constant-condition
+    while (true) {
```
venilinvasilev: Suggestion: Consider using a setInterval stored in a constant so it can be cleared when needed. As it stands, this while (true) loop will run indefinitely once executed, which isn’t ideal for an example. If this is meant to demonstrate polling behavior, it's better to follow best practices by including a way to stop or clean up the polling loop.

examples/polling-mirror-node.js
```diff
@@ -0,0 +1,95 @@
...
+/* eslint-disable n/no-unsupported-features/node-builtins */
+
+/**
+ * @param {EventEmitter} dataEmitter
+ * @param {string} topicId
+ */
+function pollUntilReady(dataEmitter, topicId) {
+    let lastMessagesLength = 0;
+
+    setInterval(() => {
```
venilinvasilev: How about assigning this interval to a timer variable that is passed as a dependency trought the function so it can be cleared when polling is no longer needed (for example when the user navigates from a page). WDYT?

examples/polling-mirror-node.js
```diff
@@ -0,0 +1,95 @@
...
+                const newMessages = data.messages.slice(lastMessagesLength)[0];
+                const decodedMessage = Buffer.from(
+                    newMessages.message,
+                    "base64",
+                ).toString("utf-8");
+                dataEmitter.emit("newMessages", decodedMessage);
+                lastMessagesLength = currentMessagesLength;
+            }
+        })().catch(console.error); // Handle any errors from the async function
+    }, 1000);
```
venilinvasilev: How about assign this to a const variable e.g. something like POLLING_INTERVAL?

examples/polling-mirror-node.js
ivaylonikolov7: `eslint` cries that we have added a constant that doesn't get used.  I think the user can handle this and will set it to a constant and pass it to clearInterval when needed.

examples/run-all-examples.js
```diff
@@ -18,6 +18,7 @@ const excludedJSFile = [
     "create-update-delete-node.js",
     "batch-tx.js",
     "long-term-schedule-transaction.js",
+    "polling-mirror-node.js",
```
venilinvasilev: Leftover?

venilinvasilev (approved):
LGTM

Issue #2919, opened 2025-03-04
Jumbo transactions


0xivanov:
### Problem

Services will implement this https://github.com/hashgraph/hedera-improvement-proposal/pull/1086.

We need to stay tuned, for any support to give , and the remove the HFS code, to use directly the Ehtx

ivaylonikolov7:
resolved via #3187

PR #3187, opened 2025-06-30
feat: Jumbo Transaction

ivaylonikolov7:
**Description**:
- Adds integration tests that check if the jumbo transactions work as expected
- update the local node to the latest version network tag
- updates EthereumFlow size
- adds Frontend example with the jumbo tx

**Related issue(s)**:
#2919 

Fixes #
#2919 

**Checklist**

### DIFF
--- a/packages/proto/Taskfile.yml
+++ b/packages/proto/Taskfile.yml
@@ -19,7 +19,7 @@ tasks:
 
     install:submodules:
         vars:
-            stag: "v0.62.3"
+            stag: "v0.63.7"
         cmds:
             - git submodule update --init --recursive --remote
             - |
--- a/packages/proto/src/proto/services/event_core.proto
+++ b/packages/proto/src/proto/services/event_core.proto
@@ -66,14 +66,10 @@ message EventCore {
   proto.Timestamp time_created = 3;
 
   /**
-   * A list of EventDescriptors representing the parents of this event.<br/>
-   * The list of parents SHALL include zero or one self parents, and zero or more other parents.<br/>
-   * The first element of the list SHALL be the self parent, if one exists.<br/>
-   * The list of parents SHALL NOT include more than one parent from the same creator.
-   * <p>
-   * This field is deprecated and can be removed in release 0.62
+   * Used to be a list of EventDescriptors representing the parents of this event.<br/>
+   * This field has been moved to GossipEvent.
    */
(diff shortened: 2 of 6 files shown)


src/EthereumFlow.js
```diff
@@ -180,7 +181,9 @@ export default class EthereumFlow {
             ethereumTransaction
                 .setEthereumData(ethereumTransactionDataBytes)
                 .setCallDataFileId(this._callDataFileId);
-        } else if (ethereumTransactionDataBytes.length <= 5120) {
+            // in the consensus node config file, the maximum size of the call data file is 133120 bytes
+            // so we need to check if the call data is less than or equal to 133120 bytes
+        } else if (ethereumTransactionDataBytes.length <= 133120) {
```
0xivanov: I think it should be 128_000, since `133120` is the limit for the whole transaction.

test/integration/EthereumTransactionIntegrationTest.js
```diff
@@ -191,4 +195,163 @@ describe("EthereumTransactionIntegrationTest", function () {
...
+            .setFunction(
+                "test",
+                new ContractFunctionParameters().addBytes(
+                    new Uint8Array(1024 * 127).fill(1),
+                ),
+            )
+            .setContractId(contractId)
+            .execute(env.client);
+
+        const type = "02"; // 02 e tip eip1i
```
0xivanov:
```suggestion
        const type = "02";
```

src/EthereumFlow.js
```diff
@@ -33,6 +33,7 @@ import * as hex from "./encoding/hex.js";
 
 /**
  * Create a new Hedera™ transaction wrapped ethereum transaction.
+ * @deprecated - The implementation of Jumbo Transactions made this flow obsolete.
```
0xivanov:
```suggestion
 * @deprecated - use EthereumTransaction instead. With the introduction of jumbo transactions, it should always be less cost and more efficient to use EthereumTransaction instead.
```

0xivanov: Can you add test for ethereum flow, just to be sure it works with something like 50kb of calldata?

0xivanov (approved):
LGTM

andrewb1269hg (changes requested):
Can you please review the spacing in these files? Indents should be 2 spaces, not 4. I think the package.json and potentially other files have 4 space indents, resulting in large diffs.

PR #3200, opened 2025-07-01
feat(entity-id): add toEvmAddress method

venilinvasilev:
**Description**:
WIP
**Related issue(s)**:

Fixes #
https://github.com/hiero-ledger/hiero-sdk-js/issues/3154

**Notes for reviewer**:

**Checklist**


src/EntityIdHelper.js
```diff
@@ -224,6 +225,45 @@ export function fromSolidityAddress(address) {
     return [shard, realm, num];
 }
 
+/**
+ * Return the shard, realm, and num from a evm address.
+ *
```
ivaylonikolov7: nit: says it returns shard, realm and num but it returns the EVM address also.

src/EntityIdHelper.js
```diff
@@ -244,6 +284,26 @@ export function toSolidityAddress(address) {
...
 }
 
+/**
+ * Convert account num into a evm address.
+ *
+ * @param {Uint8Array<ArrayBufferLike> | null} evmAddress
+ * @param {Long} accountNum
+ * @returns {string}
+ */
+export function toEvmAddress(evmAddress, accountNum) {
```
ivaylonikolov7: It's confusing that the `toEvmAddress` accepts a parameter called `evmAddress`. Might be a better idea to make this function twice with overload. One of them accepts the `uint8array` representation of the evmAddress and the other one accepts `accountNum`

src/account/AccountId.js
```diff
@@ -299,6 +306,16 @@ export default class AccountId {
         }
     }
 
+    /**
+     * @returns {string}
```
0xivanov:
```suggestion
     * @returns {string} EVM-compatible address representation of the entity
```

src/account/AccountId.js
```diff
@@ -299,6 +306,16 @@ export default class AccountId {
...
+     */
+    toEvmAddress() {
+        return entity_id.toEvmAddress(
+            this.evmAddress?.toBytes() ?? null,
+            this.num,
+        );
+    }
```
ivaylonikolov7: What do you think of changing the `entity_id` import to `Entity_Id` in the whole file for better readability. Small case makes it seem like it's an instance of an `entity_id`. 🤔

ivaylonikolov7: I am aware that on other places we use it this way e.g. `hex` but in this case `EntityId` could be an instance and it was hard to read at first glance.

ivaylonikolov7 (commented):
```javascript
EVM_ADDRESS_LENGTH = 20
LONG_ZERO_PREFIX_LENGTH = 12
ENTITY_NUM_OFFSET = 12
```

If we could add these constants instead of the magic numbers please.

src/account/AccountId.js
ivaylonikolov7: I've noticed that in java it's called EntityIdHelper, sounds better imo.

venilinvasilev: I've renamed it to EntityIdHelper as I agree it sounds better and clearer.

src/account/AccountId.js
venilinvasilev: Done

src/EntityIdHelper.js
venilinvasilev: Made it more clear now, thanks!

src/EntityIdHelper.js
venilinvasilev: I've refactored this as suggested. Thanks!

ivaylonikolov7 (approved):

PR #3201, opened 2025-07-01
chore: move `react-native-get-random-values` to peerDependencies

ahmadi-akbar:
fix #3055

Issue #3055, opened 2025-04-16
[Intermediate]: Remove `react-native-get-random-values` from Cryptography Package


venilinvasilev:
## 🧩 Intermediate Friendly

This issue is a good fit for contributors who are comfortable navigating
package boundaries within the Hiero JS SDK and understanding how environment-specific
dependencies impact bundling and runtime behavior.

Intermediate Issues often involve:

- Cleaning up architectural boundaries  
- Removing unintended cross-environment coupling  
- Updating documentation alongside code changes  
- Auditing related areas to prevent regressions  

This task requires careful reasoning about environment abstraction and dependency management.

---

## 🐞 Problem Description

The `cryptography` package currently depends on `react-native-get-random-values`.

This dependency was originally introduced as a convenience workaround so that
users running `@hashgraph/sdk` in **Expo or React Native environments** would not
need to manually import `react-native-get-random-values` at the application root
(e.g., `App.js` / `App.tsx`).

While this simplified setup for some users, it introduced architectural issues:

- The cryptography package is no longer environment-agnostic.
- It breaks separation of concerns by including React Native-specific logic.
- It pulls unnecessary React Native dependencies into non-RN environments.
- It may negatively affect tree shaking and dependency graphs.
- It creates confusion for consumers expecting a pure cryptography layer.

The cryptography package should have **zero awareness of runtime environment**
(Node, browser, React Native, etc.).

---

## 💡 Expected Outcome

The cryptography package should be restored to being fully environment-independent.

Specifically:

- Remove `react-native-get-random-values` from `dependencies` and/or `peerDependencies`.
- Remove any `import "react-native-get-random-values"` statements.
- Ensure the package works correctly in Node and browser environments without React Native assumptions.
- Update documentation to clearly state that **React Native / Expo users must import `react-native-get-random-values` themselves before using the SDK.**

The responsibility for polyfilling `crypto.getRandomValues` in React Native should live in the application layer, not within the SDK.

---

## 🧠 Implementation Notes

Suggested approach:

1. Remove the dependency from `package.json`.
2. Remove all related imports from the cryptography package.
3. Audit for any implicit assumptions about React Native.
4. Run tests across:
   - Node environment
   - Browser build (if applicable)
   - Integration tests (Solo-based CI)
5. Update README or setup documentation with a clear note for React Native users.

Important constraints:

- Do not introduce environment detection logic inside cryptography.
- Preserve existing cryptographic behavior and test coverage.
[...]

PR #3203, opened 2025-07-01
chore: move react-native-get-random-values to peerDependencies

ivaylonikolov7:
**Original contributor: @ahmadi-akbar**

**Description**:
Makes `react-native-get-random-values` optional peer dependency.

Due to a problem with `pull` bot. I had to recreate #3201. 

**Related issue(s)**:
#3055

**Fixes**
#3055

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/cryptography/package.json
+++ b/packages/cryptography/package.json
@@ -65,7 +65,6 @@
         "crypto-js": "^4.2.0",
         "forge-light": "1.1.4",
         "js-base64": "^3.7.7",
-        "react-native-get-random-values": "^1.11.0",
         "spark-md5": "^3.0.2",
         "tweetnacl": "^1.0.3",
         "utf8": "^3.0.0"
@@ -73,7 +72,13 @@
     "peerDependenciesMeta": {
         "expo-crypto": {
             "optional": true
+        },
+        "react-native-get-random-values": {
+            "optional": true
         }
+    },
+    "peerDependencies": {
+        "react-native-get-random-values": "^1.11.0"
     },
     "devDependencies": {
         "@babel/cli": "^7.27.0",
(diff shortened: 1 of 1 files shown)


ivaylonikolov7:
@ahmadi-akbar I did some research on your solution. We could implement this in a future major version bump but for now this may not be possible as far as I understood

I believe merging this may introduce a breaking change, as existing expo users would need to update their code - for example, by installing an additional package. Upgrading to a new version could potentially break their current applications. Introducing a breaking change should be done only when bumping to a major version so currently this is not feasible.

@AdrianKBL, please correct me if I’m mistaken, but I assume this would affect you and other expo users as well?

franfernandez20:
> @ahmadi-akbar I did some research on your solution. We could implement this in a future major version bump but for now this may not be possible as far as I understood
> 
> I believe merging this may introduce a breaking change, as existing expo users would need to update their code - for example, by installing an additional package. Upgrading to a new version could potentially break their current applications. Introducing a breaking change should be done only when bumping to a major version so currently this is not feasible.
> 
> @AdrianKBL, please correct me if I’m mistaken, but I assume this would affect you and other expo users as well?

I understand the motivation behind this change, but yes, this will make the release not backward compatible for React Native and Expo users.

I suggest we clearly document this in the release notes, explicitly mentioning the impact on React Native users. Also, we should update the README (React Native section) to indicate that they now need to install this dependency manually.

ivaylonikolov7:
@franfernandez20 

Even if we document it, I’m not sure we can avoid potential community backlash, since maintainers would still need to update their projects. Kabila was the first team that came to mind when I thought of Expo users—hence why I reached out to you. Would it be an issue on your end if we made this package an optional peer dependency?

venilinvasilev:
First of all thank you for taking time to contribute to our repository @ahmadi-akbar! I've taken my time to review these changes.
Unfortunately I do not see how these changes, resolve the issue with the sdk having external dependencies which it should not have.
I've pulled this branch locally and confirmed that the sdk still has external dependencies like react/react-native etc...
To resolve this issue we need to remove the `react-native-get-random-values` dependency completely and change the sdk usage for expo applications thus introducing a breaking change. We can introduce breaking changes only in major versions as we are strictly following semantic versioning. WDYT? cc: @ahmadi-akbar

PR #3209, opened 2025-07-03
chore: Enable dual publishing in SDK_JS

### LINKED ISSUE #3208: Enable dual publishing on SDK-JS
Enable Dual Publishing on the SDK-JS flows

rbarker-dev:
## Important Notices

:rotating_light: NOT READY TO MERGE :rotating_light: 

:warning: TSC needs to approve dual publishing window before this PR is mergeable :warning:

## Description

This pull request updates the publishing workflow in `.github/workflows/publish_release.yaml` to support publishing packages under two different namespaces (`@hashgraph` and `@hiero-ledger`). The changes introduce additional environment variables and modify the publishing commands to handle namespace switching.

### Workflow Updates for Dual Namespace Publishing:

* **Environment Variables for Tokens**: Added `NPM_HG_TOKEN` and `NPM_HL_TOKEN` environment variables to replace the single `NPM_TOKEN` for publishing under different namespaces. (`[.github/workflows/publish_release.yamlL347-R386](diffhunk://#diff-4f398b9d652dfd8c5f7eaa797dc903b64a3735a99002c81f795d877bc43b4ba8L347-R386)`)

* **Namespace Switching Logic**: Updated the `run` commands for the `Publish Proto Release`, `Publish Cryptography Release`, and `Publish SDK Release` jobs to:
  - Use `jq` to modify the `package.json` file to switch the package name from the `@hashgraph` namespace to the `@hiero-ledger` namespace.
  - Publish the package under both namespaces by using the respective tokens. (`[.github/workflows/publish_release.yamlL347-R386](diffhunk://#diff-4f398b9d652dfd8c5f7eaa797dc903b64a3735a99002c81f795d877bc43b4ba8L347-R386)`)

### Related Issues

closes #3208

### Notes

- [ ] Tested

### DIFF
--- a/.github/workflows/publish_release.yaml
+++ b/.github/workflows/publish_release.yaml
@@ -344,22 +344,46 @@ jobs:
 
       - name: Publish Proto Release
         env:
-          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+          NPM_HG_TOKEN: ${{ secrets.NPM_HG_TOKEN }}
+          NPM_HL_TOKEN: ${{ secrets.NPM_HL_TOKEN }}
         working-directory: packages/proto
         if: ${{ needs.validate-release.outputs.proto-publish-required == 'true' && !cancelled() && !failure() }}
-        run: task -v publish -- ${{ steps.proto-publish.outputs.args }}
+        run: |
+          NPM_TOKEN="${NPM_HG_TOKEN}"
+          task -v publish -- ${{ steps.proto-publish.outputs.args }}
+          cp package.json tmp.package.json
+          jq '.name |= sub("@hashgraph";"@hiero-ledger")' tmp.package.json > package.json
+          NPM_TOKEN="${NPM_HL_TOKEN}"
+          task -v publish -- ${{ steps.proto-publish.outputs.args }}
+          mv tmp.package.json package.json
 
       - name: Publish Cryptography Release
         env:
-          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+          NPM_HG_TOKEN: ${{ secrets.NPM_HG_TOKEN }}
+          NPM_HL_TOKEN: ${{ secrets.NPM_HL_TOKEN }}
         working-directory: packages/cryptography
         if: ${{ needs.validate-release.outputs.crypto-publish-required == 'true' && !cancelled() && !failure() }}
-        run: task -v publish -- ${{ steps.crypto-publish.outputs.args }}
+        run: |
+          NPM_TOKEN="${NPM_HG_TOKEN}"
+          task -v publish -- ${{ steps.crypto-publish.outputs.args }}
+          cp package.json tmp.package.json
+          jq '.name |= sub("@hashgraph";"@hiero-ledger")' tmp.package.json > package.json
+          NPM_TOKEN="${NPM_HL_TOKEN}"
+          task -v publish -- ${{ steps.crypto-publish.outputs.args }}
+          mv tmp.package.json package.json
 
       - name: Publish SDK Release
         env:
-          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
-        run: task -v publish -- ${{ steps.sdk-publish.outputs.args }}
+          NPM_HG_TOKEN: ${{ secrets.NPM_HG_TOKEN }}
+          NPM_HL_TOKEN: ${{ secrets.NPM_HL_TOKEN }}
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/publish_release.yaml
```diff
@@ -344,22 +344,46 @@ jobs:
 
       - name: Publish Proto Release
         env:
-          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+          NPM_HG_TOKEN: ${{ secrets.NPM_HG_TOKEN }}
+          NPM_HL_TOKEN: ${{ secrets.NPM_HL_TOKEN }}
```
andrewb1269: Need the HL token created before testing.

ryjones: done

andrewb1269hg:
While you're in this file - I think the step `Validate Tag and Package Versions` has duplicate checks that are happening.

andrewb1269hg (changes requested):
While you're in this file - I think the step Validate Tag and Package Versions has duplicate checks that are happening.

andrewb1269hg (approved):

andrewb1269hg (approved):

ivaylonikolov7:
Thanks for your contribution. I have only one issue that I see with the new workflow.

Our current release process consists of 2 steps. 

1. We create a release PR, merging it to main 
2. Then we create a tag which we also push. 

This triggers the `publish release` process. I don't see a way we can pass this new variable `dual-publish-enabled` to `true` if we continue our current release process. If I understood correctly it would be always false/undefined by default. Would it be possible to use some kind of environment variable or change the default value of it?

rbarker-dev:
> Thanks for your contribution. I have only one issue that I see with the new workflow.
> 
> Our current release process consists of 2 steps.
> 
> 1. We create a release PR, merging it to main
> 2. Then we create a tag which we also push.
> 
> This triggers the `publish release` process. I don't see a way we can pass this new variable `dual-publish-enabled` to `true` if we continue our current release process. If I understood correctly it would be always false/undefined by default. Would it be possible to use some kind of environment variable or change the default value of it?

@ivaylonikolov7 when the dual publish period begins we can set it to true by default.

andrewb1269hg (dismissed):

.github/workflows/publish_release.yaml
```diff
@@ -304,66 +311,111 @@ jobs:
 
       - name: Calculate Proto Subpackage Publish Arguments
         id: proto-publish
-        working-directory: packages/proto
         if: ${{ needs.validate-release.outputs.proto-publish-required == 'true' && !cancelled() && !failure() }}
         run: |
           PUBLISH_ARGS="--access public --no-git-checks"
-          [[ "${{ github.event.inputs.dry-run-enabled }}" == "true" ]] && PUBLISH_ARGS="${PUBLISH_ARGS} --dry-run"
+          [[ "${{ inputs.dry-run-enabled }}" == "true" ]] && PUBLISH_ARGS="${PUBLISH_ARGS} --dry-run"
```
rbarker-dev: Update this so we aren't using `inputs.*` in the run stanza

.github/workflows/publish_release.yaml
```diff
@@ -304,66 +311,111 @@ jobs:
...
+            jq '.name |= sub("@hashgraph";"@hiero-ledger")' tmp.package.json > package.json
+            export NPM_TOKEN="${NPM_HL_TOKEN}"
+            task -v publish -- ${{ steps.sdk-publish.outputs.args }}
+            mv tmp.package.json package.json
+          fi
 
       - name: Generate Github Release
         uses: ncipollo/release-action@440c8c1cb0ed28b9f43e4d1d670870f059653174 # v1.16.0
-        if: ${{ github.event.inputs.dry-run-enabled != 'true' && !cancelled() && !failure() }}
+        if: ${{ inputs.dry-run-enabled != 'true' && !cancelled() && !failure() }}
```
rbarker-dev:
```suggestion
        if: ${{ inputs.dry-run-enabled != 'true' }}
```

andrewb1269hg (approved):

rbarker-dev:
> Ivaylo Nikolov 9:08 AM (slack sourced)
> I am not a DevOps guy but from our dev side i think it looks good. I saw that the proto/cryptography package will be updated to the new namespace too. I saw that there is environment flag that sets the dual publishing period. So I think it matches everything I expected to happen in this PR.
> If you need an approve from our side I think you got it

@ivaylonikolov7

PR #3295, opened 2025-08-14
feat: add web integration test setup

venilinvasilev:
## Summary
This PR adds **browser integration tests** for `hiero-sdk-js`, extending coverage beyond the existing Node.js environment.  
With Solo now integrated into CI and a gRPC proxy endpoint available, we can validate SDK behavior in both **Node.js** and **browser** environments.

## Changes
- **Test Environment**
  - Configured a browser-compatible test runner.
  - Enabled integration tests to run against Solo’s gRPC proxy endpoint.

- **Browser Integration Tests**
  - Added browser-specific test suite covering key SDK features (transactions, queries, account creation, etc.).
  - Adapted existing Node.js integration tests for browser execution.

- **Configuration**
  - Updated SDK test configs to support Node.js and browser environments.

## References
- [Solo Project Repository](https://github.com/hiero-ledger/solo)


ivaylonikolov7:
Great work so far. I’d like to propose adding two distinct React environments to test whether all transactions run successfully in both setups. This is important because the majority of our web clients use React with the SDK.

The two environments should differ in their module bundlers—one using Webpack and the other using Vite. We’ve encountered compatibility issues in the past with different bundlers, so this approach will help us proactively address potential problems.

This setup would also allow us to:

- Integrate WalletConnect with the SDK and verify proper behavior.

- Test functions like executeWithSigner and getReceiptWithSigner, aligning closely with real-world use cases our SDK users encounter.

- Mitigate future problems when we return tree-shaking to the SDK. 

I understand this would require considerable effort, might involve adding new development dependencies to the SDK, changes in the tests and some rework on the tests. However, the long-term benefit of reducing recurring compatibility issues makes this a worthwhile investment.

The current implementation is solid; however, I’d like to align it as closely as possible with the real-world use cases our users encounter.

test/integration/AccountBalanceIntegrationTest.js
```diff
@@ -58,6 +58,7 @@ describe("AccountBalanceQuery", function () {
     });
 
     it("an account that does not exist should return an error", async function () {
+        console.log({ env });
```
ivaylonikolov7:
```suggestion
```

test/integration/BatchTransactionIntegrationTest.js
```diff
@@ -16,7 +16,9 @@ import {
     TopicInfoQuery,
 } from "../../src/index.js";
 import IntegrationTestEnv from "./client/NodeIntegrationTestEnv.js";
-import { setTimeout } from "timers/promises";
+
+// Cross-environment sleep function
+const sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms));
```
ivaylonikolov7: we have `wait` function `util.js`

test/integration/BatchTransactionIntegrationTest.js
```diff
@@ -146,7 +148,13 @@ describe.skip("BatchTransaction", function () {
         async function () {
             const freezeTransaction = await new FreezeTransaction()
                 .setFileId(FileId.fromString("4.5.6"))
-                .setFileHash(Buffer.from("1723904587120938954702349857", "hex"))
+                .setFileHash(
+                    new Uint8Array(
+                        "1723904587120938954702349857"
+                            .match(/.{1,2}/g)
+                            .map((byte) => parseInt(byte, 16)),
```
ivaylonikolov7: Can we extract this to a variable with a proper name to make it more readable? Doesn't make sense what we are doing here.

test/integration/EthereumTransactionIntegrationTest.js
```diff
@@ -85,115 +85,114 @@ describe("EthereumTransactionIntegrationTest", function () {
...
+        // const chainId = hex.decode("012a");
+        // const nonce = new Uint8Array();
+        // const maxPriorityGas = hex.decode("00");
+        // const maxGas = hex.decode("d1385c7bf0");
+        // const gasLimit = hex.decode("0249f0");
+        // const value = new Uint8Array();
+        // const to = hex.decode(contractAddress);
+        // const callData = new ContractFunctionParameters()
+        //     .addString("new message")
+        //     ._build("setMessage");
```
ivaylonikolov7: Why is this test commented out? Can we add a comment expalining whats happening here if it's intended?

test/integration/FreezeTransactionIntegrationTest.js
```diff
@@ -16,7 +16,7 @@ describe("FreezeTransaction", function () {
         client = env.client;
     });
 
-    it("should be executable but not supported", async function () {
+    it.skip("should be executable but not supported", async function () {
```
ivaylonikolov7: Is this intended?

venilinvasilev:
> Could we at least use the `withSigner` methods (e.g., `executeWithSigner`, `getReceiptWithSigner`), since browsers interact with a signer rather than the client itself?
> 
> I will create issues that simulate the most common browser environments - `create-react-dapp`/`vite` dapp so we would cover most of the dapps that use the SDK? I will work on it the next sprint. After that I think then we can finish this issue. wdyt?

There are tests with freeze and executeWithSigner in the TransactionIntegration test file.

test/integration/FreezeTransactionIntegrationTest.js
venilinvasilev: No. Great catch. Removed it.

ivaylonikolov7 (approved):

PavelSBorisov (approved):
LGTM

PR #3299, opened 2025-08-19
feat(TCK): add contract transactions endpoints

mariodimitrovv:
**Description**:
Implement Contract transaction for the TCK.

SDK changes:
- Makes `setStakedAccountID` and `setStakedNodeID` explicitly mutually exclusive. Setting one sets the other to null.
- Makes `setTransferAccountId` and `setTransferContractId` explicitly mutually exclusive. Setting one sets the other to null.
- Adds the ability to remove autorenew account in `ContractUpdate` via setting it to `0.0.0`
- Added check in `ContractCreateTransaction` to reject setting gas less than 0.
- Added validations in `hex.js` for decoding invalid hex strings.

### DIFF
--- a/src/LedgerId.js
+++ b/src/LedgerId.js
@@ -40,11 +40,11 @@ export default class LedgerId {
             case "3":
                 return LedgerId.LOCAL_NODE;
             default: {
-                let ledgerIdDecoded = hex.decode(ledgerId);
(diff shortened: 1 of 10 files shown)


0xivanov (approved):
LGTM

PR #3376, opened 2025-09-18
fix: https enforcing in mirror rest url

venilinvasilev:
**Description**:
WIP


ivaylonikolov7 (dismissed):

test/unit/MirrorNodeContractQuery.js
```diff
@@ -67,4 +69,152 @@ describe("MirrorNodeContractQuery", function () {
...
+        afterEach(function () {
+            // Restore original fetch
+            if (typeof global !== "undefined") {
+                global.fetch = originalFetch;
+            } else {
+                window.fetch = originalFetch;
+            }
+        });
+
+        it("should use correct insecure HTTP scheme for port 80", async function () {
```
0xivanov: isn't that the same test as `"should preserve original port for non-localhost with HTTP port 80` ?

test/unit/node/AccountId.js
```diff
@@ -364,12 +369,12 @@ describe("AccountId", function () {
...
 
             // Call the method
-            const result = await accountId.populateAccountNum(mockClient);
+            const result = await accountId.populateAccountNum(client);
 
             // Verify fetch was called with correct URL
             expect(global.fetch.calledOnce).to.be.true;
             expect(global.fetch.firstCall.args[0]).to.equal(
-                "https://testnet.mirrornode.com/api/v1/accounts/123f681646d4a755815f9cb19e1acc8565a0c2ac",
+                `https://${MirrorNetwork.TESTNET}/api/v1/accounts/123f681646d4a755815f9cb19e1acc8565a0c2ac`,
```
0xivanov, line 377: if we call testnet here i'd move that into integration.

venilinvasilev: We're not calling it actually. The call is mocked as you can see at the beginning of the test.

test/unit/MirrorNodeContractQuery.js
venilinvasilev: It is. Removed it.

PR #3383, opened 2025-09-23
fix(transaction): respect max nodes after freeze

### LINKED ISSUE #3313: setMaxNodesPerTransaction issue
### Description

As a wallet, when we are executing transactions sent from a dapp the setMaxNodesPerTransaction() on our client does not seem to work, it still requires signing of every transaction in a transaction.

Could be because the transaction is frozen or something when coming from the dapp?

### Steps to reproduce

Using this code

```js
let temp_client = Client.forNetwork(temp_signer.network as any);
const transactionSigner = await this.SigningService.getTransactionSigner();
temp_client.setOperatorWith(temp_signer.accountID, temp_signer.publicKey, transactionSigner);
temp_client.setMaxNodesPerTransaction(1);
```

Trying to execute a transaction like this:

Image

### Additional context

_No response_

### Hedera network

_No response_

### Version

latest

### Operating system

None

venilinvasilev:
### Description

This PR fixes the issue where setMaxNodesPerTransaction() had no effect if called after a transaction was already frozen. Previously, when executing transactions originating from a dApp, the client would still attempt to sign and submit against multiple nodes, regardless of the configured maximum.

With this update, setMaxNodesPerTransaction() is now correctly respected even after the transaction has been frozen, ensuring consistent behavior and allowing dApps/wallets to reduce redundant signing prompts.

### Fixed
- Fixed an issue where `setMaxNodesPerTransaction()` had no effect if called after a transaction was frozen. The setting is now applied correctly, allowing dApps and wallets to reduce redundant signing prompts when executing transactions. 

### Motivation

This change enables wallet integrations and dApps to avoid unnecessary signing requests when sending transactions through a client with a restricted node set. It aligns the behavior with developer expectations and the documented API.

Fixes #3313

### DIFF
--- a/src/transaction/Transaction.js
+++ b/src/transaction/Transaction.js
@@ -1303,6 +1303,10 @@ export default class Transaction extends Executable {
     _applyMaxNodesPerTransactionLimit(client) {
         const maxNodes = client.maxNodesPerTransaction;
 
+        if (maxNodes <= 0 || this._nodeAccountIds.length <= maxNodes) {
+            return;
+        }
+
         if (this._logger) {
             this._logger.debug(
                 `Trimming frozen transaction from ${this._nodeAccountIds.length} nodes to ${maxNodes} nodes based on maxNodesPerTransaction setting`,
@@ -1724,10 +1728,7 @@ export default class Transaction extends Executable {
 
         // Apply maxNodesPerTransaction limit to already frozen transaction
         // This allows changing the node count even after freezing while preserving signatures
-        const maxNodes = client.maxNodesPerTransaction;
-        if (maxNodes > 0 && this._nodeAccountIds.length > maxNodes) {
-            this._applyMaxNodesPerTransactionLimit(client);
-        }
+        this._applyMaxNodesPerTransactionLimit(client);
 
         // Valid checksums if the option is enabled
         if (client.isAutoValidateChecksumsEnabled()) {


src/transaction/Transaction.js
```diff
@@ -1291,6 +1291,62 @@ export default class Transaction extends Executable {
...
+        // Clear and rebuild the transactions list since it's derived from signed transactions
+        this._transactions.clear();
+
+        // Update the node account IDs (we need to unlock, update, and relock)
+        const wasNodeAccountIdsLocked = this._nodeAccountIds.locked;
+        this._nodeAccountIds.locked = false;
+        this._nodeAccountIds.setList(trimmedNodeIds);
+        if (wasNodeAccountIdsLocked) {
+            this._nodeAccountIds.setLocked();
+        }
```
0xivanov:
```suggestion
       this._nodeAccountIds.locked = true;
```

since in `_beforeExecute` where this method is called the list should always be locked

src/transaction/Transaction.js
```diff
@@ -1667,6 +1723,13 @@ export default class Transaction extends Executable {
             this.freezeWith(client);
         }
 
+        // Apply maxNodesPerTransaction limit to already frozen transaction
+        // This allows changing the node count even after freezing while preserving signatures
+        const maxNodes = client.maxNodesPerTransaction;
+        if (maxNodes > 0 && this._nodeAccountIds.length > maxNodes) {
```
0xivanov:
```suggestion
        if (this._nodeAccountIds.length > maxNodes) {
```

src/transaction/Transaction.js
```diff
@@ -1291,6 +1291,62 @@ export default class Transaction extends Executable {
...
+     * @private
+     * @param {import("../client/Client.js").default<Channel, *>} client
+     */
+    _applyMaxNodesPerTransactionLimit(client) {
+        const maxNodes = client.maxNodesPerTransaction;
+
+        if (this._logger) {
+            this._logger.debug(
+                `Trimming frozen transaction from ${this._nodeAccountIds.length} nodes to ${maxNodes} nodes based on maxNodesPerTransaction setting`,
+            );
```
ivaylonikolov7, line 1313: nice 🙌

src/transaction/Transaction.js
venilinvasilev: Should be resolved now

src/transaction/Transaction.js
venilinvasilev: Done

src/transaction/Transaction.js
venilinvasilev: Actually it turned out that this check needs to be the way is this as it falls in the statement when setMaxNodesPerTransaction is disabled (-1 or 0) and strips all nodes from the frozen transaction. So I reverted it.

0xivanov: I see

src/transaction/Transaction.js
```diff
@@ -1667,6 +1722,13 @@ export default class Transaction extends Executable {
             this.freezeWith(client);
         }
 
+        // Apply maxNodesPerTransaction limit to already frozen transaction
+        // This allows changing the node count even after freezing while preserving signatures
+        const maxNodes = client.maxNodesPerTransaction;
+        if (maxNodes > 0 && this._nodeAccountIds.length > maxNodes) {
```
0xivanov: What do you think about moving this check in the `_applyMaxNodesPerTransactionLimit` function, just to have it as a guard if the func gets called from somewhere else?

PR #3388, opened 2025-09-25
fix: Update SDK to work with latest Expo version

### LINKED ISSUE #3354: Need support for Expo 53/54 React Native
### Description

Due to changes in metro, Expo 53 and 54 no longer resolve the react-native configuration properly.

```
The package at "node_modules/@hashgraph/sdk/lib/client/NodeClient.js" attempted to import the Node standard library module "fs".
It failed because the native React runtime does not include the Node standard library.
```

Need to update this library to support Expo ^53, React ^19, and ^0.79.5

### Steps to reproduce

https://docs.expo.dev/workflow/upgrading-expo-sdk-walkthrough/

Upgrade react-native example

### Additional context

_No response_

### Hedera network

_No response_

### Version

2.72

### Operating system

None

mariodimitrovv:
**Description**:
This PR updates the SDK to work with latest Expo version

**Related issue(s)**:

Fixes #3354

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,8 +65,8 @@
         "@ethersproject/bytes": "5.8.0",
         "@ethersproject/rlp": "5.8.0",
         "@grpc/grpc-js": "1.12.6",
-        "@hashgraph/cryptography": "file:.yalc/@hashgraph/cryptography",
-        "@hashgraph/proto": "file:.yalc/@hashgraph/proto",
+        "@hashgraph/cryptography": "1.11.0",
+        "@hashgraph/proto": "2.23.0",
         "ansi-regex": "6.2.2",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -22,6 +22,7 @@
     "exports": {
         "./package.json": "./package.json",
         ".": {
+            "react-native": "./src/native.js",
```
ivaylonikolov7: might be better to use "./lib/native.js" instead of the `src` folder.

mariodimitrovv: We’re actually already using ./src/native.js, so there’s no need to change it to ./lib/native.js.

ivaylonikolov7 (approved):

PR #3397, opened 2025-09-29
feat: add account balance query endpoint

mariodimitrovv:
**Description**:
Implement Account Balance Query Endpoint

**Checklist**


0xivanov (approved):

tck/response/account.ts
```diff
@@ -2,3 +2,7 @@ export interface AccountResponse {
     readonly accountId?: string;
     readonly status: string;
 }
+
+export interface GetAccountBalanceResponse {
+    readonly balance: string;
```
0xivanov: need to add token balances as well

mariodimitrovv: Fixed

tck/params/account.ts
```diff
@@ -48,3 +48,8 @@ export interface RemoveAllowancesParams {
     readonly ownerAccountId: string;
     readonly serialNumbers?: string[];
 }
+
+export interface GetAccountBalanceParams {
+    readonly accountId: string;
```
0xivanov:
```suggestion
    readonly accountId?: string;
```

PR #3418, opened 2025-10-03
feat: add setInnertransactions example

Reccetech:
Update example to show how to use setInnerTransactions method.

### DIFF
--- a/examples/batch-tx.js
+++ b/examples/batch-tx.js
@@ -42,6 +42,7 @@ async function main() {
 
     await executeBatchWithBatchify(client);
     await executeBatchWithManualInnerTransactionFreeze(client);
+    await executeBatchWithSetInnerTransactions(client);
     client.close();
 }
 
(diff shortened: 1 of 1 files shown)


venilinvasilev (approved):
LGTM

ivaylonikolov7:
Can we update the PR title so we can follow the conventional commits convention. Maybe something like `feat: add setInnertransactions example`? All our other commits follow the conventional commits methodology. If we change the PR title when we squash the PR it will be merged into main with proper conventional commit in `main`.

ivaylonikolov7:
Nit: I saw that you incremented the steps but there might be some duplication of `Step 1`: We have `Step 1:` once in the `main` function and then in the `sub-examples` functions.

Reccetech:
@ivaylonikolov7 
I updated the title. 
I reviewed the step numbering. Looks ok to me as in different examples the numbering starts again. 
Thanks

Issue #3419, opened 2025-10-06
Pagination Support for `updateNetwork()` and `AddressBookQuery`


venilinvasilev:
## Summary
There is currently an issue where the SDK only retrieves the first page of node entries from the Mirror Node API when updating the network. This causes incomplete node lists (only the first 10 nodes are returned), which impacts the accuracy of the network address book and gRPC endpoint discovery.

## Details
Upon investigation, it was found that the Mirror Node API documentation (Swagger) for the `/network/nodes` endpoint indicates a **default pagination limit of 25 nodes per page**, but the actual API response currently returns only **10 nodes per page**. The maximum allowed value for the `limit` query parameter is also 25.

Because the SDK’s `updateNetwork()` method only processes the first page of results, the total number of nodes loaded is incomplete — particularly affecting Testnet and Mainnet, which have more than 10 nodes.

## Proposed Solution
1. **Implement pagination handling** in the `updateNetwork()` method to retrieve and aggregate all pages of nodes from the Mirror Node API.  
2. **Extend the `AddressBookQuery`** class to accept pagination parameters (e.g., `limit`, `next`, or `pageToken`) so that developers can optionally fetch subsequent pages directly.  
3. **(Optional)**: Add unit tests to validate that all nodes are successfully retrieved across multiple pages.

## Example
The following request currently returns only the first 10 nodes:
```bash
curl -s "https://mainnet-public.mirrornode.hedera.com/api/v1/network/nodes?file.id=0.0.102" | jq '.nodes | length'
```

Even if the Mirror Node later provides more entries, the SDK will still only return 10 nodes unless pagination is implemented.

## Impact
- The SDK currently underreports available nodes.  
- This could lead to missing gRPC endpoints or outdated network topology information.  
- Affects both browser and Node.js environments (though the behavior may differ slightly depending on transport).

## Action Items
- [ ] Add pagination support to `updateNetwork()`.
- [ ] Extend `AddressBookQuery` to handle `next` pagination links.
- [ ] Add test coverage for multi-page address book retrieval.
- [ ] Update documentation to reflect new pagination behavior.

## Environment
- **SDK**: hiero-sdk-js  
- **File ID**: `0.0.102` (address book)
- **Mirror Node API**: `/api/v1/network/nodes`
- **Network**: Testnet / Mainnet

---

**References**
- Mirror Node API Swagger docs (pagination details)
- Related discussion on Slack (network node pagination issue)

### Steps to reproduce

-

[...]

PR #3421, opened 2025-10-07
fix: fetch all nodes from addressbook

venilinvasilev:
**Description**:

This PR fixes the pagination issue in AddressBookQuery for web environments. The query now fetches nodes from all pages of the Mirror Node API response instead of only the first page.

Changes

Implemented pagination handling in AddressBookQuery to request and combine all node pages from the Mirror Node API.

Ensured the aggregated node list is returned as a single result in web environments.

Minor cleanup and refactor for clarity.

Outcome

AddressBookQuery now correctly retrieves all nodes from all pages in the Mirror Node API, ensuring the full address book is returned.

Fixes [#3419](https://github.com/hiero-ledger/hiero-sdk-js/issues/3419)

**Checklist**

### DIFF
--- a/src/network/AddressBookQueryWeb.js
+++ b/src/network/AddressBookQueryWeb.js
@@ -72,7 +72,7 @@ export default class AddressBookQueryWeb extends Query {
     /**
      * @param {object} props
      * @param {FileId | string} [props.fileId]
-     * @param {number} [props.limit] - Page size limit (defaults to DEFAULT_PAGE_SIZE for optimal performance)
+     * @param {number} [props.limit] - Page size limit (defaults to 25 for optimal performance)
      */
     constructor(props = {}) {
         super();


src/network/AddressBookQueryWeb.js
```diff
@@ -65,7 +72,7 @@ export default class AddressBookQueryWeb extends Query {
     /**
      * @param {object} props
      * @param {FileId | string} [props.fileId]
-     * @param {number} [props.limit]
+     * @param {number} [props.limit] - Page size limit (defaults to DEFAULT_PAGE_SIZE for optimal performance)
```
venilinvasilev: I will put 25 instead of DEFAULT_PAGE_SIZE here as the constant is not exported and visible to the end user

ivaylonikolov7: I agree. When you do this I think we should merge this. wdyt @venilinvasilev ?

Issue #3338, opened 2025-09-03
HIP-1195: Hiero hooks and an application to allowances


ivaylonikolov7:
### Problem

SDK team should implement HIP-1195 according to the design proposal. 

> We propose hooks, programmable Hiero extension points that let users customize the behavior of their entities. In principle, hooks could be programmed in any language, but we begin with EVM hooks. Users program EVM hooks by writing contracts in a language like Solidity that compiles to EVM bytecode. EVM hooks are either pure (using neither storage nor external contracts); or lambdas (like event-driven code running in the cloud, that may keep state in a database to use state or call other external services). For any given entity, its user can create many hooks for that entity with different 64-bit hook ids. There is no limit on the number of hook ids than an entity can use; but its storage footprint, and hence rent, will increase proportionally. 

### Solution

Implement the methods, enums, classes, types designed in the design proposal for 1195

ivaylonikolov7:
Closed via #3553

PR #3422, opened 2025-10-07
feat: update protobuf files to support hooks

ivaylonikolov7:
**Description**:
- Bumps protobuf files to support hooks
- fixes a bug where if there are new paths in the `proto` files it doesnt flatten the import structure
- fixes a bug where proto files sometimes have a duplicated messages so i remove duplicated imports
- adds new status codes 
- adds new response codes

**Related issue(s)**:
#3338 

**Notes for reviewer**:
-

**Checklist**


packages/proto/src/proto/streams_record_stream_file.proto
```diff
@@ -27,7 +27,7 @@ option java_package = "com.hedera.services.stream.proto";
 option java_multiple_files = true;
 
 import "services_basic_types.proto";
-import "services_ethereum_transaction.proto";
+import "services_transaction.proto";
```
0xivanov, line 30: noting this change as it looks like a red flag to me

Issue #3435, opened 2025-10-10
Global gRPC Deadline


venilinvasilev:
### Summary

Currently, SDKs do not consistently handle gRPC request timeouts across implementations.  
While the **Java SDK** includes a configurable per-request `grpcDeadline`, other SDKs either lack this feature or require it to be manually set per transaction (e.g., via `Transaction.setGrpcDeadline()`).

This results in inconsistent behavior and potentially poor user experience - especially in cases where a node becomes unresponsive or experiences long response times (e.g., Cloudflare 522 timeout scenarios).

The goal of this issue is to introduce a **global, SDK-wide `grpcDeadline` setting** that defines a maximum timeout per gRPC request, after which the SDK marks the node as unhealthy and automatically rotates to the next available node.

---

### Background

- When a node is slow or unresponsive, the SDK can currently hang for extended periods (e.g., 30–40 seconds) before failing.  
- Inconsistent timeout handling across SDKs leads to different behavior between languages, making it harder for developers to reason about network reliability and retry behavior.  
- The Java SDK already provides a `Client.setGrpcDeadline(Duration)` setting, which applies a consistent timeout to all requests.  
- Other SDKs rely on per-transaction configuration or lack timeout enforcement entirely.  

---

### Proposal

Introduce a **global `grpcDeadline` configuration** at the client level for all SDKs:

```ts
const client = Client.forMainnet();
client.setGrpcDeadline(5_000); // 5 seconds per gRPC request
```

#### Behavior
- Applies to **all outgoing gRPC requests** (queries, transactions, etc.).
- If a request exceeds this deadline:
  - The SDK **aborts the request**.
  - Marks the node as **temporarily unhealthy**.
  - **Rotates** to the next healthy node automatically.
- Can be **overridden per transaction or query**, maintaining backward compatibility with `Transaction.setGrpcDeadline()` (or equivalent).

---

### Relationship with `requestTimeout`

The SDK already defines a `requestTimeout` property on the client, which controls the **total time budget** for a complete **Transaction or Query execute operation** — including:
- Attempting the first node.
- Failing due to a node error.
- Applying retry/backoff logic.
- Retrying with another healthy node.

To ensure consistent and predictable behavior:
- `requestTimeout` **must be greater than or equal to** the per-request `grpcDeadline`.
- The SDK should enforce this constraint at runtime or initialization (e.g., throw or adjust automatically if misconfigured).
[...]

PR #3437, opened 2025-10-10
feat(client): add grpc deadline

venilinvasilev:
**Description**:
Introduce client-level timeout controls and align JS SDK behavior with cross-SDK design:

Client.grpcDeadline (ms): new client-level default per-RPC deadline with setGrpcDeadline(int) / getGrpcDeadline(). Default 10,000 ms. Used for warmup and as fallback when a Transaction doesn’t set its own deadline.

Client requestTimeout (ms): ensure presence across SDKs; default 120,000 ms for the full Transaction/Query execute() lifecycle (retries + backoff).

Behavior change (JS): when a transaction-level grpcDeadline is exceeded, the SDK now marks the node unhealthy and rotates to the next node instead of throwing a fatal error that terminates execute().

Browser (gRPC-web/Envoy): add initial HTTP healthcheck to the proxy with timeout = grpcDeadline; cache result per-node to skip repeated checks.

Validation: both grpcDeadline and requestTimeout must be positive, non-zero integers.

### Added
- **Client-level `grpcDeadline` (ms)** with `setGrpcDeadline(int)` / `getGrpcDeadline()`; default **10,000 ms**. Used for warmup and as fallback per-RPC deadline when a Transaction doesn’t specify one.
- **Client-level `requestTimeout` (ms)** to bound the full `execute()` operation (retries + backoff); default **120,000 ms**. 

### Changed
- **Deadline exceed behavior (JS)**: if a **transaction-level** `grpcDeadline` is exceeded, the SDK now **marks the node unhealthy and rotates to the next node** instead of terminating `execute()` with a fatal error.
- **Executable** now applies the **client-level `grpcDeadline`** as a fallback when no per-transaction deadline is set. 
- **Browser (gRPC-web via Envoy)**: initial **HTTP healthcheck** to the proxy with timeout = `grpcDeadline`; cached per-node.

### DIFF
--- a/manual/CONFIGURATION.md
+++ b/manual/CONFIGURATION.md
@@ -273,7 +273,9 @@ const timeInMs = 10000;
 client.setMaxExecutionTime(10000); // Set the gRPC deadline to 10 seconds (10000 milliseconds)
 ```
 
-**Important**: When configuring timeouts, ensure that `requestTimeout` is always larger than `grpcDeadline`. The SDK will throw an error if you attempt to set `grpcDeadline` greater than or equal to `requestTimeout`, or if you set `requestTimeout` less than or equal to `grpcDeadline`. This validation ensures that individual gRPC requests have sufficient time to complete within the overall operation timeout.
+**Important**: When configuring timeouts, ensure that `requestTimeout` is always larger than `grpcDeadline`. The SDK will display deprecation warnings if you attempt to set `grpcDeadline` greater than or equal to `requestTimeout`, or if you set `requestTimeout` less than or equal to `grpcDeadline`. These warnings will become errors in the next major version. This validation ensures that individual gRPC requests have sufficient time to complete within the overall operation timeout.
+
+**Deprecation Notice**: Invalid timeout configurations currently show warnings but will throw errors in the next major version. Please update your timeout configurations to avoid future breaking changes.
 
 ### Node Management
 
--- a/src/client/Client.js
+++ b/src/client/Client.js
@@ -175,7 +175,11 @@ export default class Client {
 
         // Validate that requestTimeout is larger than grpcDeadline after both are set
         if (this._requestTimeout <= this._grpcDeadline) {
-            throw new Error("requestTimeout must be larger than grpcDeadline");
+            console.warn(
+                `DEPRECATION WARNING: requestTimeout (${this._requestTimeout}ms) should be larger than grpcDeadline (${this._grpcDeadline}ms). ` +
+                    `This configuration may cause operations to fail unexpectedly. ` +
+                    `This will throw an error in the next major version. Please adjust your timeout values.`,
+            );
         }
 
         /** @internal */
@@ -707,7 +711,11 @@ export default class Client {
             throw new Error("requestTimeout must be a positive number");
         }
         if (requestTimeout <= this._grpcDeadline) {
-            throw new Error("requestTimeout must be larger than grpcDeadline");
+            console.warn(
+                `DEPRECATION WARNING: requestTimeout (${requestTimeout}ms) should be larger than grpcDeadline (${this._grpcDeadline}ms). ` +
+                    `This configuration may cause operations to fail unexpectedly. ` +
+                    `This will throw an error in the next major version. Please adjust your timeout values.`,
+            );
         }
         this._requestTimeout = requestTimeout;
         return this;
@@ -733,7 +741,11 @@ export default class Client {
             throw new Error("grpcDeadline must be a positive number");
         }
         if (grpcDeadline >= this._requestTimeout) {
-            throw new Error("grpcDeadline must be smaller than requestTimeout");
+            console.warn(
+                `DEPRECATION WARNING: grpcDeadline (${grpcDeadline}ms) should be smaller than requestTimeout (${this._requestTimeout}ms). ` +
+                    `This configuration may cause operations to fail unexpectedly. ` +
+                    `This will throw an error in the next major version. Please adjust your timeout values.`,
+            );
         }
         this._grpcDeadline = grpcDeadline;
         return this;
--- a/test/unit/Client.js
+++ b/test/unit/Client.js
@@ -77,48 +77,97 @@ describe("Client deadline configuration", function () {
         ).to.throw("requestTimeout must be a positive number");
     });
 
-    it("throws when grpcDeadline is set to be larger than or equal to requestTimeout", function () {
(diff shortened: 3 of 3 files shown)


ivaylonikolov7:
related issue #3435

src/channel/NativeChannel.js
```diff
@@ -27,6 +39,89 @@ export default class NativeChannel extends Channel {
...
+                const grpcStatus = response.headers.get("grpc-status");
+                const grpcMessage = response.headers.get("grpc-message");
+
+                // If gRPC headers exist, the proxy is running and processing requests
+                if (grpcStatus != null || grpcMessage != null) {
+                    // Mark this connection as ready
+                    this._isReady = true;
+                    return; //  Healthy - gRPC-Web proxy is responding
+                }
+            }
```
ivaylonikolov7, line 104: Not sure if this is possible but what would happen if the reponse returns 300  or one of the other statuses that indicate that the resource has been moved. 

Do you think this would be a valid case which we might need to handle?

src/channel/Channel.js
```diff
@@ -73,6 +75,28 @@ export default class Channel {
...
+         */
+        this._grpcDeadline = grpcDeadline;
+    }
+
+    /**
+     * Set the gRPC deadline for this channel
+     * @param {number} deadline - The deadline in milliseconds, or null to clear
+     */
+    setGrpcDeadline(deadline) {
+        this._grpcDeadline = deadline;
```
ivaylonikolov7, line 91: Should we add validation to prevent `grpcDeadline` from being larger than `requestTimeout`? This would ensure that individual request timeouts don't exceed the total transaction timeout (including retries and node rotation).

src/channel/NodeChannel.js
```diff
@@ -14,16 +14,15 @@ export default class NodeChannel extends Channel {
     /**
      * @internal
      * @param {string} address
-     * @param {number=} maxExecutionTime
+     * @param {number=} grpcDeadline
      */
-    constructor(address, maxExecutionTime) {
-        super();
+    constructor(address, grpcDeadline) {
+        super(grpcDeadline);
```
ivaylonikolov7, line 21: Could potentially break some of our users. Sorry 😂 We still have the setter and the getter in this node channel but we don't use it anymore. Maybe we can make it backwards compatible and check which one is set and get the bigger number wdyt?

ivaylonikolov7: Also we can add that we are deprecating this method, add a console.log for it and then mention it in the discussion about the v3 of the SDK?

src/channel/WebChannel.js
```diff
@@ -10,15 +10,122 @@ import Channel, { encodeRequest, decodeUnaryResponse } from "./Channel.js";
...
+
+        const shouldUseHttps = !(
+            this._address.includes("localhost") ||
+            this._address.includes("127.0.0.1")
+        );
+
+        const address = shouldUseHttps
+            ? `https://${this._address}`
+            : `http://${this._address}`;
+
```
ivaylonikolov7: wdyt about adding this in a separate util function. I'm noticing that we use this pattern more and more and maybe its time to separate it?

src/channel/NativeChannel.js
```diff
@@ -27,6 +39,89 @@ export default class NativeChannel extends Channel {
...
+        if (this._isReady) {
+            return; // Health check already passed for this address
+        }
+
+        const shouldUseHttps = !(
+            this._address.includes("localhost") ||
+            this._address.includes("127.0.0.1")
+        );
+
+        const address = shouldUseHttps
```
ivaylonikolov7, line 63: I am noticing this is a function that is pretty much repeating itself and not following the DRY principle. I think it would be great if we can extract it to the base class and also separate it into smaller functions. This way we could test it easier into smaller chunks and it would be more human readible. This way we wouldn't have the `execute` in `Executable` which is famous for being longer than the Bible. 

I remember reading in the book of Martin Fowler about refactoring principles "if a function is longer than 50 lines its probably a code smell".

ivaylonikolov7 (commented):
Another issue we’ve noticed is that we might be performing redundant health checks on the gRPC web proxies and gRPC servers. For example, consider a client. Her name is Elon. Elon’s setup includes a single node(so it has only one channel in it), and when she sends 20 transactions simultaneously, each transaction triggers its own health check. This results in 20 health checks being executed in a very short time span instead of just one. Such behavior could make the server interpret Elon’s activity as suspicious or malicious and consequently refuse to process her requests.

This could make it so his transactions wouldn't pass because of these health checks. We could have a boolean flag if we are waiting for a response from the server for the health check.

p.s Please rebase your PR. some merge conflicts showed. Thanks 🙌

src/channel/NodeChannel.js
venilinvasilev: NodeChannel isn't exported. It won't break any users.

src/channel/NativeChannel.js
venilinvasilev: We cannot extract it to the base class because it won't be useful for the NodeChannel as there we're not performing http requests as you may know. However your concern is still valid for the scope of WebChannel so I will extract these as private methods which can be used here and in the actual request part.
Thanks!

src/channel/NativeChannel.js
venilinvasilev: I am not aware, but I will handle it anyway.

src/channel/NativeChannel.js
venilinvasilev: Done

src/channel/WebChannel.js
venilinvasilev: Done

src/channel/Channel.js
venilinvasilev: Good point, this was in the design proposal as well, my bad :D. Thanks.

venilinvasilev: Added validation.

src/channel/NativeChannel.js
venilinvasilev: Done

src/channel/NodeChannel.js
venilinvasilev: But it seemed that only setMaxExecutionTime was an exposed setter on the NodeClient so I made an internal change to it so set grpcDeadline instead of the unexposed internal _maxExecutionTime (which could and was removed). Thanks!

ivaylonikolov7 (approved):

PR #3457, opened 2025-10-16
chore(release): v2.75.0

ivaylonikolov7:
### Fixed

- Fixed pagination issue in `AddressBookQuery` for web environments where only the first page of nodes from the Mirror Node API was being retrieved instead of all available pages. [#3421](https://github.com/hiero-ledger/hiero-sdk-js/pull/3421)


venilinvasilev (approved):
LGTM

PR #3477, opened 2025-10-26
feat: hooks examples

ivaylonikolov7:
**Description**:
- Hooks examples 
- Adds exports for Hooks

**Checklist**

### DIFF
--- a/examples/transfer-transaction-hooks-example.js
+++ b/examples/transfer-transaction-hooks-example.js
@@ -75,22 +75,22 @@ async function main() {
             hook: new LambdaEvmHook({ contractId: hookContractId }),
         });
 
-        const senderpPrivateKey = PrivateKey.generateECDSA();
+        const senderPrivateKey = PrivateKey.generateECDSA();
         const { accountId: senderAccountId } = await (
             await (
                 await new AccountCreateTransaction()
-                    .setKeyWithoutAlias(senderpPrivateKey.publicKey)
+                    .setKeyWithoutAlias(senderPrivateKey.publicKey)
                     .setInitialBalance(new Hbar(10))
                     .addHook(hookDetails)
                     .freezeWith(client)
-                    .sign(senderpPrivateKey)
+                    .sign(senderPrivateKey)
             ).execute(client)
         ).getReceipt(client);
 
         const { accountId: receiverAccountId } = await (
             await new AccountCreateTransaction()
                 .setKeyWithoutAlias(PrivateKey.generateECDSA().publicKey)
-                .setMaxAutomaticTokenAssociations(100)
+                .setMaxAutomaticTokenAssociations(-1)
                 .setInitialBalance(new Hbar(10))
                 .addHook(hookDetails)
                 .execute(client)
@@ -106,10 +106,10 @@ async function main() {
                     .setDecimals(2)
                     .setInitialSupply(10000)
                     .setTreasuryAccountId(senderAccountId)
-                    .setAdminKey(senderpPrivateKey.publicKey)
-                    .setSupplyKey(senderpPrivateKey.publicKey)
+                    .setAdminKey(senderPrivateKey.publicKey)
+                    .setSupplyKey(senderPrivateKey.publicKey)
                     .freezeWith(client)
-                    .sign(senderpPrivateKey)
+                    .sign(senderPrivateKey)
             ).execute(client)
         ).getReceipt(client);
 
@@ -126,10 +126,10 @@ async function main() {
                     .setTokenSymbol("ENT")
                     .setTokenType(TokenType.NonFungibleUnique)
                     .setTreasuryAccountId(senderAccountId)
-                    .setAdminKey(senderpPrivateKey.publicKey)
-                    .setSupplyKey(senderpPrivateKey.publicKey)
+                    .setAdminKey(senderPrivateKey.publicKey)
+                    .setSupplyKey(senderPrivateKey.publicKey)
                     .freezeWith(client)
-                    .sign(senderpPrivateKey)
+                    .sign(senderPrivateKey)
             ).execute(client)
         ).getReceipt(client);
 
@@ -143,7 +143,7 @@ async function main() {
                     .setTokenId(nftTokenId)
                     .addMetadata(metadata)
                     .freezeWith(client)
-                    .sign(senderpPrivateKey)
+                    .sign(senderPrivateKey)
             ).execute(client)
         ).getReceipt(client);
 


examples/account-hooks-example.js
```diff
@@ -0,0 +1,189 @@
...
+        process.env.OPERATOR_KEY == null ||
+        process.env.HEDERA_NETWORK == null
+    ) {
+        throw new Error(
+            "Environment variables OPERATOR_ID, HEDERA_NETWORK, and OPERATOR_KEY are required.",
+        );
+    }
+
+    const operatorId = AccountId.fromString(process.env.OPERATOR_ID);
+    const operatorKey = PrivateKey.fromStringED25519(process.env.OPERATOR_KEY);
```
0xivanov: ecdsa

examples/account-hooks-example.js
```diff
@@ -0,0 +1,189 @@
...
+         * Step 2: Demonstrate creating an account with hooks.
+         */
+        console.log("\n=== Creating Account with Hooks ===");
+        console.log("Creating account with lambda EVM hook...");
+
+        const accountKey = PrivateKey.generate();
+        const accountPublicKey = accountKey.publicKey;
+
+        console.log(`private key = ${accountKey.toString()}`);
+        console.log(`public key = ${accountPublicKey.toString()}`);
```
0xivanov: why we print this? does it provide any useful info?

examples/account-hooks-example.js
```diff
@@ -0,0 +1,189 @@
...
+        // Verify the account by querying account info
+        try {
+            const accountInfo = await new AccountInfoQuery()
+                .setAccountId(accountId)
+                .execute(client);
+
+            console.log("Account ID:", accountInfo.accountId.toString());
+            console.log("Account balance:", accountInfo.balance.toString());
+        } catch (error) {
+            console.error("Failed to query account info:", error);
```
0xivanov: what do we verify? Hooks are not present in this query

examples/contract-hooks-example.js
```diff
@@ -0,0 +1,169 @@
...
+         * Step 3: Demonstrate adding hooks to an existing contract.
+         */
+        console.log("\n=== Adding Hooks to Existing Contract ===");
+        console.log("Adding hooks to existing contract...");
+
+        const adminKey = client.operatorPublicKey;
+
+        // Hook 3: Basic lambda hook with no storage updates (using ID 3 to avoid conflict with existing hook 1)
+        const basicHook = new LambdaEvmHook({ contractId: hookContractId });
+        const hook3 = new HookCreationDetails({
```
0xivanov:
```suggestion
        const hookWithID3 = new HookCreationDetails({
```

0xivanov: I got confused because of the naming, can be done everywhere

examples/lambda-sstore-example.js
```diff
@@ -0,0 +1,154 @@
...
+
+        console.log("Hook contract created with ID:", contractId.toString());
+
+        // Create account with lambda hook
+        console.log("Creating account with lambda hook...");
+        const accountKey = PrivateKey.generate();
+        const accountPublicKey = accountKey.publicKey;
+
+        console.log(`private key = ${accountKey.toString()}`);
+        console.log(`public key = ${accountPublicKey.toString()}`);
```
0xivanov: same as the other comment

examples/transfer-transaction-hooks-example.js
```diff
@@ -0,0 +1,276 @@
...
+        await (
+            await (
+                await new AccountCreateTransaction()
+                    .setKeyWithoutAlias(operatorKey.publicKey)
+                    .setInitialBalance(new Hbar(1))
+                    .addHook(hookDetails)
+                    .freezeWith(client)
+                    .sign(operatorKey)
+            ).execute(client)
+        ).getReceipt(client);
```
0xivanov: we don't do anything with this account?

examples/transfer-transaction-hooks-example.js
```diff
@@ -0,0 +1,276 @@
...
+
+        // Use existing accounts (operator as sender, node 0.0.3 as receiver)
+        const senderAccountId = operatorId;
```
0xivanov: having the sender = operator means you basically do not need the hooks, since the sender will always sign.
Can make the result of the newly created account the sender.

Also, having hbar, nft, ft transfers together is not as readable. Id suggest separating them via methods.

examples/transfer-transaction-hooks-example.js
```diff
@@ -0,0 +1,276 @@
...
+        );
+        console.log(
+            "To actually execute this transaction, the hooks (IDs 1001-1004) must exist on the network.",
+        );
+        console.log(
+            "The transaction would be executed with: transferTx.execute(client)",
+        );
+
+        /*
+         * Step 3: Execute a simple transfer without hooks that actually works
```
0xivanov: why do we need this? It even verifies that we don't need the hooks in this instance which is counter-intuitive

examples/transfer-transaction-hooks-example.js
ivaylonikolov7: No, i just need it to create the hook.

0xivanov: how do you invoke the hook when you don't use the account id of the created account?

examples/account-hooks-example.js
ivaylonikolov7: We need the other PR from Mario to use ECDSA here. 0.0.2 is ED25519 and the examples use account 0.0.2 to run these examples.

examples/transfer-transaction-hooks-example.js
ivaylonikolov7: I separated them in different transaction wdyt? I am not sure if separating a single transaction in a separate method would improve readability.

examples/account-hooks-example.js
ivaylonikolov7: merged Mario's PR about using ECDSA in all our examples and fixed Hook examples in his PR

examples/contract-hooks-example.js
```diff
@@ -0,0 +1,169 @@
...
+         */
+        console.log("\n=== Deleting Hooks from Contract ===");
+        console.log("Deleting hooks from contract...");
+
+        try {
+            await (
+                await (
+                    await new ContractUpdateTransaction()
+                        .setContractId(contractWithHooksId)
+                        .addHookToDelete(Long.fromNumber(1)) // Delete hook created during contract creation
```
venilinvasilev, line 149: Isn't it a good idea to have these ids in variables?

ivaylonikolov7: i am not sure if it won't add unnecessary noise to the example. I think the comment explains why we add these here.

examples/transfer-transaction-hooks-example.js
```diff
@@ -0,0 +1,270 @@
...
+        const hookContractId = hookContractCreateReceipt.contractId;
+
+        const hookId = Long.fromInt(1);
+        const hookDetails = new HookCreationDetails({
+            extensionPoint: HookExtensionPoint.ACCOUNT_ALLOWANCE_HOOK,
+            hookId: hookId,
+            hook: new LambdaEvmHook({ contractId: hookContractId }),
+        });
+
+        const senderpPrivateKey = PrivateKey.generateECDSA();
```
0xivanov:
```suggestion
        const senderPrivateKey = PrivateKey.generateECDSA();
```

0xivanov (approved):
LGTM

0xivanov (approved):

PR #3535, opened 2025-11-19
feat: hip-1195

ivaylonikolov7:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


src/account/AccountCreateTransaction.js
```diff
@@ -652,10 +665,36 @@ export default class AccountCreateTransaction extends Transaction {
                     : null,
             stakedNodeId: this.stakedNodeId,
             declineReward: this.declineStakingRewards,
+            hookCreationDetails: this._hooks.map((hook) => hook._toProtobuf()),
```
venilinvasilev, line 668: Is this ok if it is an empty list. Directly resolve this comment if yes.

src/account/AccountUpdateTransaction.js
```diff
@@ -533,6 +555,57 @@ export default class AccountUpdateTransaction extends Transaction {
...
+        return this;
+    }
+
+    /**
+     * @returns {Long[]}
+     */
+    get hooksToDelete() {
+        return this._hooksToBeDeleted;
+    }
+
```
venilinvasilev, line 608: I assume we do not have a clear all hooks option here as we need specifiers for the hook we need to remove. If yes please resolve the comment directly.

src/account/AccountUpdateTransaction.js
```diff
@@ -613,6 +686,11 @@ export default class AccountUpdateTransaction extends Transaction {
                 this.declineStakingRewards != null
                     ? { value: this.declineStakingRewards }
                     : null,
+
+            hookIdsToDelete: this._hooksToBeDeleted,
```
venilinvasilev, line 690: Don't we need to map these toProtobuf here. It seems that TS would yell, but I am confused.

ivaylonikolov7: `hookIdsToDeleted` is Long[] so its fine that we dont have to call `toProtobuf`.

src/account/AccountUpdateTransaction.js
ivaylonikolov7: I think we communicated that we wont have a clear method in the collaboration hub. I cant find clear methods in the design doc.

PR #3540, opened 2025-11-20
feat: implement ContractCallQuery endpoint

mariodimitrovv:
**Description**:
Implement `ContractCallQuery` endpoints

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/tck/methods/account.ts
+++ b/tck/methods/account.ts
@@ -12,6 +12,8 @@ import {
     TokenId,
     EvmAddress,
     AccountBalanceQuery,
+    AccountInfoQuery,
+    AccountInfo,
 } from "@hiero-ledger/sdk";
 import Long from "long";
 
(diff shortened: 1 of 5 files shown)


tck/methods/contract.ts
```diff
@@ -286,3 +292,45 @@ export const executeContract = async ({
...
+
+    if (contractId != null) {
+        query.setContractId(contractId);
+    }
+
+    if (gas != null) {
+        query.setGas(Long.fromString(gas));
+    }
+
+    if (functionParameters != null) {
```
venilinvasilev, line 318: Why do we have this check structured like this - first if functionParameters is. not null we set function parameters then in an else if block we check for function name? Shouldn't we check for both with if blocks?

mariodimitrovv: Fixed

PR #3610, opened 2026-01-05
feat: use pnpm link for linking packages

venilinvasilev:
**Description**:
This PR switches outdated dependency **yalc** for linking packages used for local development to **pnpm link**

**Related issue(s)**:

Fixes 
[#3354](https://github.com/hiero-ledger/hiero-sdk-js/issues/3534)

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,8 +106,8 @@
         "@types/utf8": "3.0.1",
         "@typescript-eslint/eslint-plugin": "5.48.2",
         "@typescript-eslint/parser": "5.62.0",
-        "@vitest/browser": "3.1.2",
-        "@vitest/coverage-v8": "3.1.2",
+        "@vitest/browser": "3.2.4",
+        "@vitest/coverage-v8": "3.2.4",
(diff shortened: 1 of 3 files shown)


mariodimitrovv (approved):

.github/workflows/react_native.yml
```diff
@@ -54,15 +54,15 @@ jobs:
...
-        run: npm install -g yarn
-
       - name: Build @hiero-ledger/sdk
         run: task build
 
       - name: Install detox
         run: npm install -g detox-cli
 
+      - name: Install React Native example deps
+        run: pnpm install --filter react-native-example --no-frozen-lockfile
```
PavelSBorisov: Why are we using `--no-frozen-lockfile` here?

venilinvasilev: Great catch :). We're using it in our taskfiles as well. We removed this flag so dependencies are consistent.

ivaylonikolov7 (approved):

PR #3698, opened 2026-02-10
ci: optimizations

ivaylonikolov7:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/proto/Taskfile.yml
+++ b/packages/proto/Taskfile.yml
@@ -84,7 +84,7 @@ tasks:
         cmds:
             - pnpm exec pbjs -r hashgraph -t static-module -w es6 -p=src/proto --force-long --no-beautify --no-convert --no-delimited --no-verify -o src/proto.js src/proto/**/*.proto
             - pnpm exec pbts -n hashgraph -o src/proto.d.ts src/proto.js
-            - perl -pi -e 's#(?<!api\.)proto\.#hashgraph.proto.#g' src/proto.d.ts
+            - node scripts/fix-proto-namespaces.js
             - task: format
             - task: lint
             - npx babel src -d lib


packages/proto/Taskfile.yml
```diff
@@ -82,9 +82,9 @@ tasks:
 
     build:standalone:
         cmds:
-            - npx pbjs -r hashgraph -t static-module -w es6 -p=src/proto --force-long --no-beautify --no-convert --no-delimited --no-verify -o src/proto.js src/proto/*.proto
-            - npx pbts -n hashgraph -o src/proto.d.ts src/proto.js
-            - node scripts/fix-proto-namespaces.js
+            - pnpm exec pbjs -r hashgraph -t static-module -w es6 -p=src/proto --force-long --no-beautify --no-convert --no-delimited --no-verify -o src/proto.js src/proto/**/*.proto
+            - pnpm exec pbts -n hashgraph -o src/proto.d.ts src/proto.js
+            - perl -pi -e 's#(?<!api\.)proto\.#hashgraph.proto.#g' src/proto.d.ts
```
venilinvasilev: This does not work on windows, that's why I used the js script instead

packages/proto/Taskfile.yml
```diff
@@ -136,7 +137,7 @@ tasks:
                 if [ "$(git rev-parse --is-shallow-repository)" = "true" ]; then
                     git fetch --unshallow --tags
                 fi &&
-                git checkout {{.stag}}
+                git checkout {{.proto}}
```
venilinvasilev, line 140: Could you elaborate this change

packages/proto/Taskfile.yml
ivaylonikolov7: Oh yeah, my bad. Good catch!

packages/proto/Taskfile.yml
ivaylonikolov7: The variable `stag` doesn't exist. It should be `.proto` so I guess it fixes a bug because it didn't work as expected.

PR #3846, opened 2026-03-10
feat(hip-1137): implementation

venilinvasilev:
**Description**:
Implementation of hip-1137

Fixes #
https://github.com/hiero-ledger/hiero-sdk-js/issues/3829
**Notes for reviewer**:

**Checklist**


src/client/Client.js
```diff
@@ -344,6 +344,14 @@ export default class Client {
         return this._mirrorNetwork.mirrorRestApiBaseUrl;
     }
 
+    /**
+     * @returns {string}
+     * @throws {Error} When no mirror network is configured or available
+     */
+    get mirrorRestJavaApiBaseUrl() {
+        return this._mirrorNetwork.mirrorRestJavaApiBaseUrl;
```
ivaylonikolov7: I think it misses a check that throws an error when network is missing?

rbarker-dev (dismissed):

src/node/RegisteredNode.js
```diff
@@ -0,0 +1,138 @@
...
+     * @returns {IStateRegisteredNode}
+     */
+    _toProtobuf() {
+        return {
+            registeredNodeId: this.registeredNodeId,
+            adminKey: this.adminKey._toProtobufKey(),
+            description: this.description != null ? this.description : null,
+            serviceEndpoint: this.serviceEndpoints.map((endpoint) =>
+                endpoint._toProtobuf(),
+            ),
```
ivaylonikolov7:
```javascript
    toBytes() {
        const body = {
            registeredNodeId: this.registeredNodeId,
            adminKey: this.adminKey._toProtobufKey(),
            description: this.description != null ? this.description : null,
            serviceEndpoint: this.serviceEndpoints.map((endpoint) =>
                endpoint._toProtobuf(),
            ); 
        return HieroProto.com.hedera.hapi.node.state.addressbook.RegisteredNode.encode(
body
        ).finish();
    }
```
wdyt?

src/node/RegisteredNodeAddressBook.js
```diff
@@ -0,0 +1,41 @@
...
+     * @param {IStateRegisteredNode[]} registeredNodes
+     * @returns {RegisteredNodeAddressBook}
+     */
+    static _fromProtobuf(registeredNodes) {
+        return new RegisteredNodeAddressBook({
+            registeredNodes: registeredNodes.map((registeredNode) =>
+                RegisteredNode._fromProtobuf(registeredNode),
+            ),
+        });
+    }
```
ivaylonikolov7: why do we have `_fromProtobuf` but not `_toProtobuf` ?

src/node/RegisteredNodeUpdateTransaction.js
```diff
@@ -0,0 +1,403 @@
...
+
+import Long from "long";
+import Key from "../Key.js";
+import Transaction, {
+    TRANSACTION_REGISTRY,
+} from "../transaction/Transaction.js";
+import RegisteredServiceEndpoint from "./RegisteredServiceEndpoint.js";
+
+const DESCRIPTION_MAX_BYTES = 100;
+const SERVICE_ENDPOINTS_MAX_LENGTH = 50;
```
ivaylonikolov7: these are repeated can we add them in a `src/constant/` so if they are changed its easier for us

src/node/RegisteredNodeUpdateTransaction.js
```diff
@@ -0,0 +1,403 @@
...
+            throw new TypeError(
+                "serviceEndpoints must not be null or undefined.",
+            );
+        }
+
+        if (serviceEndpoints.length === 0) {
+            throw new Error("ServiceEndpoints list must not be empty.");
+        }
+
+        if (serviceEndpoints.length > SERVICE_ENDPOINTS_MAX_LENGTH) {
```
ivaylonikolov7: Could this be a blocker for the TCK. I remember previously we had conversations if this should be handled on SDK level or by consensus node?

src/query/FeeEstimateQuery.js
```diff
@@ -428,7 +428,9 @@ export default class FeeEstimateQuery {
                 if (response.status === 400) {
                     // 400 = malformed transaction (INVALID_ARGUMENT). Do not retry.
                     throw new Error(
-                        `HTTP 400 Bad Request${errorDetail ? `: ${errorDetail}` : ""}`,
+                        `HTTP 400 Bad Request${
```
ivaylonikolov7, line 435: I thought that these formatting changes were only on my side. Maybe we can fix them in the next release not in this PR?

src/client/Client.js
venilinvasilev: This functionallity has been removed from the changes in latest commits.

src/node/RegisteredNode.js
venilinvasilev: What do you mean I have the toBytes method which calls the this._toProtobuf and endcodes it at the end of the file?

ivaylonikolov7:
I think we can also remove the retry mechanism for gRPC errors. You can remove the `RST_STREAM` import, the `MirrorError` import, the gRPC code switch.

src/node/RegisteredNodeUpdateTransaction.js
venilinvasilev: removed as these we're part of a validation that needed to happen on consensus node side.

src/node/RegisteredNodeAddressBook.js
venilinvasilev: we should have from and to protobuf methods tor the stuff we need like transactions.

ivaylonikolov7 (approved):

PavelSBorisov (approved):

PR #3870, opened 2026-03-16
feat: migration to ethers v6

Neurone:
**Description**:
This PR migrates the project dependency from ethers v5 to ethers v6.

**Related issue(s)**:

Fixes: N/A

**Notes for reviewer**:

Ethers v5 contains dependencies that included some minor vulnerabilities that will not be fixed (i.e. elliptic). Tweaking with those dependencies is complex and risky. Migrating to ethers v6 solves the issue entirely.

**Checklist**

### DIFF
--- a/src/contract/ContractFunctionResult.js
+++ b/src/contract/ContractFunctionResult.js
@@ -8,7 +8,9 @@ import * as hex from "../encoding/hex.js";
 import * as utf8 from "../encoding/utf8.js";
 import * as util from "../util.js";
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-import { ParamType, defaultAbiCoder } from "@ethersproject/abi";
+import { AbiCoder, ParamType } from "ethers";
+
+const defaultAbiCoder = AbiCoder.defaultAbiCoder();
 import Long from "long";
 import ContractNonceInfo from "./ContractNonceInfo.js";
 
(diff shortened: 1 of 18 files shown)


src/contract/ContractFunctionResult.js
```diff
@@ -1000,10 +1002,55 @@ export default class ContractFunctionResult {
     /**
      * @description Decode the data according to the array of types, each of which may be a string or ParamType.
      * @param {Array<string | ParamType>} types
-     * @returns {string | any}
```
ivaylonikolov7, line 1003: Wouldnt that return be a breaking change for users as previosuly it returned string or any and now it returns Array ?

src/contract/ContractFunctionResult.js
```diff
@@ -1000,10 +1002,55 @@ export default class ContractFunctionResult {
...
+     * @returns {Array<string | number | BigNumber>}
```
venilinvasilev, line 1005: I am worried that this is one is theoretically a breaking change of the public API, even though the previous signature was actually incorrect - this has never returned a string, but rather the Result type of decode which is a list again. From my side I see no potential issues - @ivaylonikolov7 WDYT?

Neurone: Yeah, I confirm that was to me a _false_ return type (in fact we don't have tests covering that return type), so I see no issues in updating that. I addition, I noticed a small reduction in the code coverage of that file, due to a single line that is essentially dead code in the `_convertEthersValue`. I fixed that, so if you see this code outdated it's because of that.

PR #3871, opened 2026-03-16
chore: upgrade TypeScript to 5.9.3 and fix TS 5.9 Uint8Array compatib…

### LINKED ISSUE #3479: Typescript bump incompatibility
### Description

TypeScript 5.9.3 introduced stricter type definitions for TypedArray and Buffer.

This change breaks compatibility with certain libraries (e.g., hashgraph/proto), resulting in errors like:

Type 'Buffer' is not assignable to type 'Uint8Array<ArrayBufferLike>'
In TS 5.9.3, Uint8Array now defaults to Uint8Array<ArrayBuffer>, while Node’s Buffer is typed as Uint8Array<ArrayBufferLike>.
Since these types are no longer compatible, encoding transactions fails in Node.js environments (Not sure If happens the same in FrontEnd Environments).

## :bulb: Potential fix

A possible solution would be to introduce a dual type definition inside of proto.encode() (Might it applies to more methods) that supports both Node and browser environments, e.g.:

`Buffer | Uint8Array`

This would make the SDK Node-friendly without requiring us to refactor all our existing DApps.

reference: https://github.com/microsoft/TypeScript/issues/62240
possible solution: https://github.com/paulmillr/scure-base/issues/53#issuecomment-3435534798

### Steps to reproduce

-

### Additional context

_No response_

### Hedera network

_No response_

### Version

latest

### Operating system

None

ivaylonikolov7:
Description: 
- Upgrade TypeScript to 5.9.3 and fix resulting type compatibility issues caused by TS 5.9 making Uint8Array generic.
- Upgrade typescript from 5.7.2 to 5.9.3
- Upgrade @types/node from 24.0.8 to 25.5.0 to resolve TextEncoder/TextDecoder type conflicts

Fixes: #Should resolve #3479

### DIFF
--- a/src/cryptography/sha384.browser.js
+++ b/src/cryptography/sha384.browser.js
@@ -8,8 +8,8 @@ export async function digest(data) {
     // https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest
     // This will be executed in a browser environment so the crypto object should be available if its
     // in secure context.
-    // eslint-disable-next-line n/no-unsupported-features/node-builtins
     const input =
-        data.buffer instanceof ArrayBuffer ? data : new Uint8Array(data); // copy SharedArrayBuffer-backed data into plain ArrayBuffer
+        data.buffer instanceof ArrayBuffer ? data : new Uint8Array(data);
+    // eslint-disable-next-line n/no-unsupported-features/node-builtins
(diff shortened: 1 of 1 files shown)


venilinvasilev (dismissed):
LGTM

PR #3874, opened 2026-03-17
chore(release): beta features

ivaylonikolov7:
**Description**:
A PR where we can merge our features to and cherry pick them to main for a release whenever we need to. 

Fixes #

**Notes for reviewer**:

**Checklist**


venilinvasilev:
I think that it would be better to have beta branch for each individual feature, so we could easily cherry pick the changes. Example lets say we have 2 features one would be deployed to prod earlier and the one would be uploaded to prod later. We merge those 2 features in the beta branch and release a beta with the 2 features for testing.
Then on the next iteration we need to release one of those features only. It would be pain to cherry pick changes - especially if there are any overlaps in file changes.
I suggest having every feature in a separate beta banch with some convention like feat/hooks-beta-release or something.

Issue #3701, opened 2026-02-11
Orphaned deadline timers in Executable prevent Node.js process exit after Client.close()


vbichkovsky:
### Problem

After `Client.close()`, the Node.js process stays alive for ~10 seconds before exiting. This affects any short-lived process that uses the SDK: e2e tests, CLI tools, serverless functions, etc.

### Root cause

`Executable._execute()` races every gRPC call against a deadline timer ([`src/Executable.js` L521-538](https://github.com/hiero-ledger/hiero-sdk-js/blob/32533f570157b2540f45c5c0d25fb341183cfb1b/src/Executable.js#L521-L538)):

```js
if (this._grpcDeadline != null) {
    promises.push(
        new Promise((_, reject) =>
            setTimeout(
                () => reject(new GrpcServiceError(GrpcStatus.DeadlineExceeded)),
                this._grpcDeadline,
            ),
        ),
    );
}
// ...
response = await Promise.race(promises);
```

When the gRPC call succeeds, `Promise.race` resolves — but the losing `setTimeout` is **never cleared**. The timer ID is not captured, so `clearTimeout` cannot be called. Each successful gRPC call (execute, getReceipt, etc.) leaks one ref'd timer with `DEFAULT_GRPC_DEADLINE` (10 000 ms) delay. These timers keep the Node.js event loop alive after `Client.close()`.

### Reproduction

```js
import { Client, TransferTransaction, Hbar, AccountId, PrivateKey } from "@hashgraph/sdk";

const client = Client.forTestnet();
client.setOperator(AccountId.fromString("0.0.XXXX"), PrivateKey.fromStringDer("..."));

const tx = new TransferTransaction()
  .addHbarTransfer(AccountId.fromString("0.0.XXXX"), Hbar.fromTinybars(0));

const response = await tx.execute(client);
await response.getReceipt(client);
console.log("done");

client.close();
// Process hangs here for ~10 seconds instead of exiting immediately
```

### Profiling data

Using `async_hooks` to track timers after `client.close()`:

```
Timeout: delay=10000ms, refed=true, destroyed=false, callback=()=>reject(new GrpcServiceError(GrpcStatus.DeadlineExceeded))
Timeout: delay=10000ms, refed=true, destroyed=false, callback=()=>reject(new GrpcServiceError(GrpcStatus.DeadlineExceeded))
[...]

PR #3877, opened 2026-03-18
fix: node js doesnt exit until timers are finished

ivaylonikolov7:
**Description**:
The timer ID was never saved, so clearTimeout was never called.

Even in the happy path (gRPC responds in 200ms), the 10-second countdown was still ticking in the background — orphaned, with no way to cancel it. Every gRPC call leaked one timer like this.

Node.js won't exit while there are active timers, so after client.close() the process would just sit there waiting for all those 10-second countdowns to expire naturally before finally quitting.

The fix just saves the timer ID so it can be cancelled the moment 

**Related issue(s)**: #3701
 
Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/src/Executable.js
+++ b/src/Executable.js
@@ -375,10 +375,13 @@ export default class Executable {
 
         try {
             // Whichever settles first wins: the real response or the deadline error.
-            return await Promise.race([
-                this._execute(channel, request),
-                deadlinePromise,
-            ]);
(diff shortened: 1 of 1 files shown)


venilinvasilev:
Closing this as it is covered in https://github.com/hiero-ledger/hiero-sdk-js/pull/3880

PR #3880, opened 2026-03-20
chore(refactor): executable

venilinvasilev:
**Description**:
This PR refactors the execute in Executable class, so it is better readable and maintainable.

Fixes #
https://github.com/hiero-ledger/hiero-sdk-js/issues/3867

**Checklist**

### DIFF
--- a/src/Executable.js
+++ b/src/Executable.js
@@ -639,12 +639,6 @@ export default class Executable {
             return false;
         }
 
-        console.error(
-            `Attempting to execute a transaction against node ${nodeAccountId.toString()}, which is not included in the Client's node list. Please review your Client configuration.`,
-        );
-
-        this._nodeAccountIds.advance();
-
         return true;
     }
 
@@ -826,6 +820,11 @@ export default class Executable {
                     this._nodeAccountIds.current,
                 )
             ) {
+                console.error(
+                    `Attempting to execute a transaction against node ${this._nodeAccountIds.current.toString()}, which is not included in the Client's node list. Please review your Client configuration.`,
+                );
+
+                this._nodeAccountIds.advance();
                 continue;
             }
 
@@ -921,7 +920,7 @@ export default class Executable {
                 status.toString() !== Status.Success.toString();
 
             if (isError) {
-                persistentError = new Error(status.toString());
+                persistentError = status;
             }
 
             // Determine by the executing state what we should do
--- a/test/unit/Executable.js
+++ b/test/unit/Executable.js
@@ -59,18 +59,33 @@ describe("Executable", function () {
         expect(beforeExecuteCalled).to.be.true;
     });
 
-    it("_throwIfRequestTimedOut throws with the current node account ID", function () {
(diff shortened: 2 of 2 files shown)


src/Executable.js
```diff
@@ -550,108 +583,272 @@ export default class Executable {
...
-        const LOCAL_NODE_ATTEMPTS = 1000;
-        const maxAttempts = isLocalNode
-            ? LOCAL_NODE_ATTEMPTS
-            : client._maxAttempts ?? this._maxAttempts;
+    /**
+     * @private
+     * @param {number} executeOperationStartTime
+     * @returns {void}
+     */
+    _throwIfRequestTimedOut(executeOperationStartTime) {
```
ivaylonikolov7: Not sure if this is a good idea to be separated in a different function.

src/Executable.js
```diff
@@ -550,108 +583,272 @@ export default class Executable {
...
+     *
+     * @template {Channel} ChannelT
+     * @template {MirrorChannel} MirrorChannelT
+     * @param {import("./client/Client.js").default<ChannelT, MirrorChannelT>} client
+     * @param {number=} requestTimeout
+     * @returns {Promise<OutputT>}
+     */
+    async execute(client, requestTimeout) {
+        await this._setupExecution(client, requestTimeout);
+        const isLocalNode = client.isLocalNetwork;
```
ivaylonikolov7, line 788: I don't think we need to extract this to a varaible/constant 🤔

src/Executable.js
```diff
@@ -550,108 +583,272 @@ export default class Executable {
...
-            // Get the log ID for the request.
-            const logId = this._getLogId();
-            if (this._logger) {
-                this._logger.debug(
-                    `[${logId}] Node AccountID: ${node.accountId.toString()}, IP: ${node.address.toString()}`,
-                );
+            const executionNode = this._getExecutionNode(client);
+
+            if (executionNode == null) {
+                continue;
```
ivaylonikolov7: Isn't this dead code because this.getExecutionNode throws an error if its null?

src/Executable.js
```diff
@@ -752,112 +886,86 @@ export default class Executable {
...
+            );
+
+            const isError =
                 status.toString() !== Status.Ok.toString() &&
-                status.toString() !== Status.Success.toString()
-            ) {
+                status.toString() !== Status.Success.toString();
+
+            if (isError) {
                 persistentError = status;
```
ivaylonikolov7: Is it a good idea to keep both type Error and Status. I know its not part of this refactor but I just saw we assign both Errors and Status to this constant and it feels weird to me, wdyt? Maybe we can have handleStatus(status) which returns of type Error ?

ivaylonikolov7: Which seems even weirder to have a function that returns an Error but I guess its a werid niche case where returning an Error might be appropriate.

ivaylonikolov7: Sorry for the spam. No need for the handle function -  `persistentError= new Error(status.toString)`?

PR #3913, opened 2026-04-01
fix(ethereum): add proper access list support for tx types 1, 2 and 4

### LINKED ISSUE #3885: Add proper support for access list in tx type 1, 2 and 4 (Pectra)
### Problem

Creating `EthereumTransactionData` from a raw transaction is currently possible, and the access list appears to be one of the decoded fields. However, the current implementation of the `fromBytes` methods expects the access list to be an array of strings, which is not compliant with Ethereum standards. Each access list item should include not only an address but also an array of associated storage keys.

This inconsistency currently results in an error:
```
TypeError: text.startsWith is not a function
    at Object.decode (/Users/mariusz.jasuwienas/Projects/sdk-accesslist-issue-sample/node_modules/@hiero-ledger/sdk/lib/encoding/hex.cjs:25:20)
    at /Users/mariusz.jasuwienas/Projects/sdk-accesslist-issue-sample/node_modules/@hiero-ledger/sdk/lib/EthereumTransactionDataEip1559.cjs:90:66
    at Array.map (<anonymous>)
    at EthereumTransactionDataEip1559.fromBytes (/Users/mariusz.jasuwienas/Projects/sdk-accesslist-issue-sample/node_modules/@hiero-ledger/sdk/lib/EthereumTransactionDataEip1559.cjs:90:53)
    at Cache._ethereumTransactionDataEip1559FromBytes (/Users/mariusz.jasuwienas/Projects/sdk-accesslist-issue-sample/node_modules/@hiero-ledger/sdk/lib/EthereumTransactionDataEip1559.cjs:133:99)
    at EthereumTransactionData.fromBytes (/Users/mariusz.jasuwienas/Projects/sdk-accesslist-issue-sample/node_modules/@hiero-ledger/sdk/lib/EthereumTransactionData.cjs:36:31)
    at /Users/mariusz.jasuwienas/Projects/sdk-accesslist-issue-sample/scripts/create-submit-accesslist.js:25:31
```
when attempting to create EthereumTransactionData from raw transaction bytes when the access list is not empty.

To reproduce the issue, run the following script.

1. Prerequisites:

**node.js v24.13.1** 

```bash
nvm use 24.13.1 
```

2. Install required libraries:
```bash
npm install @hiero-ledger/sdk@2.81.0
npm install ethers@5.7.2
```
(or use the following package.json and type `npm install`)
```json
{
  "name": "sdk-accesslist-issue-sample",
  "version": "1.0.0",
  "dependencies": {
    "@hiero-ledger/sdk": "^2.81.0",
    "ethers": "^5.7.2"
  }
}
```
3. Create sample script to replicate this issue (paste this content into the file `scripts/accesslist-problem-showcase.js`):
```js
const { EthereumTransactionData } = require("@hiero-ledger/sdk");
const { Wallet, utils: { arrayify } } = require("ethers");

const wallet = new Wallet(Wallet.createRandom().privateKey);

const tx = {
  type: 1,
  chainId: 296,
  nonce: 0,
  gasPrice: 0,
  gasLimit: 21_000,
  to: "0x0000000000000000000000000000000000000000",
  value: 0,
  data: "0x",
  accessList: [
    {
      address: wallet.address,
      storageKeys: [],
    },
  ],
};

wallet.signTransaction(tx)
    .then((signed) => {
      EthereumTransactionData.fromBytes(arrayify(signed));
    });
```
and run:

```bash
node scripts/accesslist-problem-showcase.js
```
You can modify the JavaScript code above by removing the access list to verify that it works in that case, confirming that the issue only occurs when the access list is provided.

### Solution

Add proper support for the access list by expecting the correct format it has after RLP decoding of the raw transaction:

https://github.com/hiero-ledger/hiero-sdk-js/blob/main/src/EthereumTransactionDataEip1559.js#L86

```js
            accessList: /** @type {string[]} */ (decoded[8]).map((v) =>
                hex.decode(v),
            ),
```
`v` can be an array here, which doesn’t have `startsWith`, so it cannot be passed into the `decode` function, which explicitly expects `v` to be a `string`.

### Alternatives

Please let me know if I’m using your script correctly and whether you plan to address this issue on your end, or if I should make adjustments on my side to pass the access list in a different format.

venilinvasilev:
## Summary
  - Fix TypeError: text.startsWith is not a function when decoding Ethereum
  transactions with non-empty access lists                                 
  - Properly decode access list entries as [address, storageKeys[]] tuples per
  EIP-2930 spec                                                               
  - Applied across EIP-2930 (type 1), EIP-1559 (type 2), and EIP-7702 (type 4)  
  - Closes #3885

### DIFF
--- a/src/EthereumTransactionDataEip1559.js
+++ b/src/EthereumTransactionDataEip1559.js
@@ -4,6 +4,16 @@ import EthereumTransactionData from "./EthereumTransactionData.js";
 import CACHE from "./Cache.js";
 
 /**
+ * @typedef {[Uint8Array, Uint8Array[]]} AccessListItem - [address, storageKeys[]]
+ */
+
+/**
+ * @typedef {object} AccessListItemJSON
+ * @property {string} address
+ * @property {string[]} storageKeys
+ */
+
+/**
  * @typedef {object} EthereumTransactionDataEip1559JSON
  * @property {string} chainId
  * @property {string} nonce
@@ -13,7 +23,7 @@ import CACHE from "./Cache.js";
  * @property {string} to
  * @property {string} value
  * @property {string} callData
- * @property {string[]} accessList
+ * @property {AccessListItemJSON[]} accessList
  * @property {string} recId
  * @property {string} r
  * @property {string} s
@@ -31,7 +41,7 @@ export default class EthereumTransactionDataEip1559 extends EthereumTransactionD
      * @param {Uint8Array} props.to
      * @param {Uint8Array} props.value
      * @param {Uint8Array} props.callData
-     * @param {Uint8Array[]} props.accessList
+     * @param {AccessListItem[]} props.accessList
      * @param {Uint8Array} props.recId
      * @param {Uint8Array} props.r
      * @param {Uint8Array} props.s
@@ -61,8 +71,9 @@ export default class EthereumTransactionDataEip1559 extends EthereumTransactionD
             throw new Error("empty bytes");
         }
 
-        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-        const decoded = /** @type {string[]} */ (decodeRlp(bytes.subarray(1)));
+        const decoded = /** @type {unknown[]} */ (
+            /** @type {unknown} */ (decodeRlp(bytes.subarray(1)))
+        );
 
         if (!Array.isArray(decoded)) {
             throw new Error("ethereum data is not a list");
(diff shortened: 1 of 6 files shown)


ivaylonikolov7:
Two suggestions that I have: 

 - Extract named typedefs for the tuple types — Add `@typedef {[Uint8Array, Uint8Array[]]} AccessListItem` (in all three files) and `@typedef {[Uint8Array, Uint8Array, Uint8Array, Uint8Array, Uint8Array, Uint8Array]} AuthorizationItem` (in EIP-7702). Then use `AccessListItem[]` and     
  `AuthorizationItem[]`in the constructor JSDoc instead of the long inline tuple annotations. I am aware that this PR is about accessList. While reviewing it I found that the `AuthroizationItem` I have implemented is barely readable and I have no idea what does these properties mean 😅  - [Uint8Array, Uint8Array, Uint8Array, Uint8Array, Uint8Array, Uint8Array]
- Replace @ts-ignore / eslint-disable with proper type casts — Instead of suppressing the type checker, cast the decoded RLP output through unknown to the expected shape, e.g. /** @type {Array<[string, string[]]>} */ (/** @type {unknown} */ (decoded[8])) for access lists. Once the  outer array is properly typed, the individual hex.decode() calls don't need their own @type casts or lint suppressions anymore, so all of those can be removed. 

Other than that LGTM.

ivaylonikolov7 (approved):

PR #3930, opened 2026-04-09
feat: add contractId support to HookEntityId

ddeskov-limechain:
## Summary

`HookEntityId` represents the owner of a hook — either an account or a contract. The protobuf definition (`services_basic_types.proto`) defines it as a `oneof` with both `account_id` and `contract_id` fields:

```protobuf
message HookEntityId {
    oneof entity_id {
        AccountID account_id = 1;
        ContractID contract_id = 2;
    }
}
```

The current JS SDK implementation only supports `accountId`. This PR adds the missing `contractId` field, bringing the JS SDK to parity with the Java SDK (`HookEntityId.java`) and Go SDK (`hook_id.go`), which both already support both fields.

### `src/hooks/HookEntityId.js`

- Import `ContractId` from `../contract/ContractId.js`
- Add `_contractId` private field (initialized to `null`) with JSDoc
- Add `@param {ContractId} [props.contractId]` to constructor
- Add `setContractId(contractId)` method — sets the contract ID and clears `_accountId` (oneof enforcement)
- Update `setAccountId(accountId)` to also clear `_contractId` (oneof enforcement)
- Add `get contractId()` getter
- Update `_toProtobuf()` to serialize `contractId`
- Update `_fromProtobuf()` to deserialize `contractId`

### `test/unit/HookEntityId.js`

28 new test cases added (32 → 60 total), covering:

- **Constructor**: create with `contractId`, `ContractId.fromString()`, shard/realm/num
- **`setContractId`**: set and chain, overwrite, different formats
- **Oneof behavior**: setting `contractId` clears `accountId` and vice versa, constructor with both props (`contractId` wins since it's processed second), toggling between the two
- **Getter**: get `contractId`, return `null` when unset, updated after setter
- **`_toProtobuf`**: convert with `contractId`, both fields null
- **`_fromProtobuf`**: create from protobuf with `contractId`, null handling, property preservation
- **Round-trip serialization**: `contractId` survives protobuf round-trip
- **Edge cases**: very large contract numbers, `ContractId` with EVM address
- **Integration with `ContractId`**: `fromString`, property preservation, equality

### Oneof mutual exclusion enforced in setters

Since `HookEntityId` IS the oneof wrapper itself (not a class that happens to contain a oneof), mutual exclusion is enforced directly in the setters: `setContractId` clears `_accountId`, and `setAccountId` clears `_contractId`. This matches the Java SDK's behavior, where the same clearing logic applies. The Go SDK achieves the same effect through separate factory functions (`NewHookEntityIdWithAccountId` / `NewHookEntityIdWithContractId`).

### Constructor with both props

If both `accountId` and `contractId` are passed in the constructor, `contractId` wins because it's processed second (its setter clears `accountId`). This is a natural consequence of the oneof enforcement and is documented with a dedicated test case.

## How to verify

```bash
# Run HookEntityId unit tests
npx vitest run test/unit/HookEntityId.js --config test/vitest-node.config.ts

# Run all hook-related unit tests
npx vitest run test/unit/HookEntityId.js test/unit/HookId.js test/unit/HookStoreTransaction.js --config test/vitest-node.config.ts
```

All 126 tests pass (60 HookEntityId + 35 HookId + 31 HookStoreTransaction).

### DIFF
--- a/src/hooks/HookEntityId.js
+++ b/src/hooks/HookEntityId.js
@@ -1,4 +1,5 @@
 import AccountId from "../account/AccountId.js";
+import ContractId from "../contract/ContractId.js";
 
 /**
  * The id of an entity using a hook.
@@ -8,6 +9,7 @@ class HookEntityId {
      *
      * @param {object} props
      * @param {AccountId} [props.accountId]
+     * @param {ContractId} [props.contractId]
      */
     constructor(props = {}) {
         /**
@@ -16,8 +18,18 @@ class HookEntityId {
          */
         this._accountId = null;
 
+        /**
+         * @private
+         * @type {?ContractId}
+         */
+        this._contractId = null;
+
         if (props.accountId != null) {
             this.setAccountId(props.accountId);
+        }
+
+        if (props.contractId != null) {
+            this.setContractId(props.contractId);
         }
     }
 
@@ -27,6 +39,7 @@ class HookEntityId {
      */
     setAccountId(accountId) {
         this._accountId = accountId;
+        this._contractId = null;
         return this;
     }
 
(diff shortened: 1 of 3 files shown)


venilinvasilev (dismissed):
LGTM

ivaylonikolov7 (approved):

PR #3974, opened 2026-04-28
feat: HIP-1313

ivaylonikolov7:
**Description**:
Implement HIP-1313


src/transaction/Transaction.js
```diff
@@ -221,6 +222,17 @@ export default class Transaction extends Executable {
...
         this._isThrottled = false;
+
+        /**
+         * Whether to use high-volume throttles for this transaction.
+         * When true, enables high-volume throttles and pricing for entity creation.
+         * Only affects supported transaction types; otherwise, it is ignored.
+         * Undefined for transaction types that do not support high-volume.
+         *
+         * @private
+         * @type {boolean | undefined}
```
venilinvasilev: nit: Why can't we just have it boolean and assigned to false by default, do we need to have it with an undefined state at all?

src/transaction/Transaction.js
```diff
@@ -844,6 +857,29 @@ export default class Transaction extends Executable {
...
+        return this;
+    }
+
+    /**
+     * Get whether high-volume throttles are enabled for this transaction.
+     *
+     * @returns {boolean}
+     */
+    get highVolume() {
+        return this._highVolume ?? false;
```
venilinvasilev: we could just return this_.highVolume if we set it to false by default right?

venilinvasilev (approved):
LGTM

PavelSBorisov (approved):

PR #4049, opened 2026-05-11
test(channel): add unit tests for NativeChannel

Utkal059:
Add Vitest unit-test coverage for `NativeChannel`, which previously had none. Tests mock `fetch`, `FileReader`, and the native base64 encoder so the suite runs in a standard Node/Vitest environment without a browser or live gRPC-Web endpoint.

- Add `test/unit/channel/NativeChannel.js` (24 cases, 315 lines)
- Cover constructor: `_address` stored, `_isReady` defaults to `false`, custom `grpcDeadline` honoured
- Cover `close()`: confirmed no-op returning `undefined`
- Cover `_waitForReady()`: cached `_isReady` fast path; `https://` for remote and `http://` for `localhost` / `127.0.0.1`; past-deadline throws `Timeout`; `grpc-status` and `grpc-message` headers each mark the channel ready; HTTP 200 with no gRPC headers throws `Unavailable`; `AbortError` maps to `Timeout`; existing `GrpcServiceError` re-thrown unchanged; generic network error maps to `Unavailable`
- Cover `_createUnaryClient()`: `GrpcServiceError` propagated to callback; URL built as `https://<addr>/proto.<svc>/<method>` for remote and `http://...` for localhost; non-OK HTTP returns `HttpError`; both `application/octet-stream` and `application/grpc-web+proto` base64 prefixes decoded; unknown MIME prefix surfaces a descriptive `Error`; fully successful round-trip returns a `Uint8Array`

**Related issue(s)**:

N/A — coverage improvement, no linked issue.

**Notes for reviewer**:

The `FileReader` stub fires `load` / `loadend` via `queueMicrotask` rather than synchronously inside `readAsDataURL`. This matches real browser behaviour and is required because `NativeChannel.js` assigns `reader.onloadend` *after* calling `readAsDataURL(blob)` — a synchronous stub would fire before the handler is attached and every decode test would hang.

No production code changed; this PR is test-only.

**Checklist**

### DIFF
--- a/test/unit/NativeChannel.js
+++ b/test/unit/NativeChannel.js
@@ -1,0 +1,316 @@
+// SPDX-License-Identifier: Apache-2.0
+import NativeChannel from "../../src/channel/NativeChannel.js";
+import GrpcServiceError from "../../src/grpc/GrpcServiceError.js";
+import GrpcStatus from "../../src/grpc/GrpcStatus.js";
+import HttpError from "../../src/http/HttpError.js";
+
+// minimal valid gRPC-Web DATA frame (header 0x00 + length 1 + 1 payload byte)
+const VALID_GRPC_FRAME = new Uint8Array([0x00, 0x00, 0x00, 0x00, 0x01, 0x00]);
+
+vi.mock("../../src/encoding/base64.native.js", () => ({
+    encode: vi.fn(() => "bW9jaw=="),
+    decode: vi.fn(() => VALID_GRPC_FRAME),
+}));
+
+// ─── helpers ──────────────────────────────────────────────────────────────
+
+const makeFetchResponse = (status, grpcHeaders = {}) => ({
+    ok: status >= 200 && status < 300,
+    status,
+    headers: { get: (key) => grpcHeaders[key] ?? null },
+    blob: vi.fn().mockResolvedValue(new Blob()),
+    text: vi.fn().mockResolvedValue(""),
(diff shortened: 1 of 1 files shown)


test/unit/channel/NativeChannel.js
```diff
@@ -0,0 +1,317 @@
+// SPDX-License-Identifier: Apache-2.0
+import { vi, describe, it, expect, beforeEach, afterEach } from "vitest";
```
venilinvasilev: No need to import these

venilinvasilev:
Hi @Utkal059, thanks for taking the time to contribute!
The test file should be at test/unit/NativeChannel.js rather than test/unit/channel/NativeChannel.js - we don't mirror the source directory structure in tests. The only permitted subdirectories are test/unit/node/ and test/unit/browser/ for environment-specific tests; everything else goes directly under test/unit/.

Sorry the original issue had the wrong path  it's been corrected now!

Utkal059:
Thanks @venilinvasilev! Moved the file to test/unit/NativeChannel.js and removed the explicit Vitest globals import. Also updated the relative import paths accordingly. Force-pushed — please let me know if anything else needs adjusting!

ivaylonikolov7:
Thank you for the help! 

You need to import `vi`. `task build` is failing.

ivaylonikolov7:
closing this due to inactivity

Issue #3590, opened 2025-12-15
[Advanced]: Standardize Mirror Node Ingress Endpoint for Local & CI Environments


venilinvasilev:
## 🧠 Advanced

This issue is well-suited for contributors who are very familiar with the
Hiero JS SDK architecture, configuration patterns, and environment handling
across Node and Web targets.

Advanced Issues often involve:

- Shaping SDK configuration abstractions  
- Coordinating behavior across multiple environments (Node/Web)  
- Reasoning about long-term maintainability and fragility  
- Updating CI, defaults, tests, and documentation together  

This task impacts local development, CI reliability, and future-proofing
against Mirror Node internal changes.

---

## 🐞 Problem Description

**Scope:** Local development and CI only  
**Dependency:** Solo exposing Mirror Node via ingress  
Related: https://github.com/hiero-ledger/hiero-solo-action/issues/120  
Mirror Node migration context: https://github.com/hiero-ledger/hiero-mirror-node/issues/1699  

The Mirror Node team is migrating endpoints from the JS REST API service
to the Java API service. When the SDK directly targets specific Mirror Node
service ports in local or CI setups, it becomes fragile — especially when
endpoints move, are consolidated, or are removed.

Currently:

- `hiero-sdk` points to specific Mirror Node service ports in local setups  
- Node and Web environments use different ports:
  - Node → gRPC  
  - Web → HTTP  
- Endpoint migrations require SDK configuration updates  
- CI and local development depend on internal service topology  

This tight coupling creates unnecessary churn and increases maintenance
overhead when Mirror Node internals change.

Solo will expose Mirror Node via a **single ingress endpoint**, which should
serve as the stable entry point for local and CI environments.

The SDK should treat this ingress endpoint as the canonical base URL,
rather than binding to specific service ports.

---

## 💡 Proposed / Expected Outcome

Update the SDK’s local and CI configuration model to support a **single
Mirror Node base endpoint**, provided via Solo ingress.

The SDK should:

- Accept **one Mirror Node endpoint configuration**
- Internally determine transport usage:
  - gRPC for Node environments  
  - HTTP for Web environments  
- Avoid directly targeting individual service ports in local/CI setups  
- Remain unaffected by internal Mirror Node service migrations  

The design should:

- Preserve existing public configuration APIs where possible  
- Avoid introducing environment-specific configuration duplication  
- Cleanly abstract transport differences behind a stable endpoint  

This change ensures that:

[...]

PR #4096, opened 2026-05-27
feat: add retry logic for transient mirror node errors

Utkal059:
## Summary
Adds exponential backoff retry logic to `MirrorNodeContractQuery.performMirrorNodeRequest` to handle transient mirror node errors (HTTP 500/503/504 and network-level failures).

## Changes
- `src/query/MirrorNodeContractQuery.js` — retry loop with up to 5 attempts, doubling backoff from 250ms to 8s max; non-retryable 4xx errors thrown immediately
- `src/network/AddressBookQueryWeb.js` — normalize error message format to match
- `test/unit/MirrorNodeContractQuery.js` — 4 new unit tests covering retry on 503, retry on 504, exhaustion after 5 attempts, no retry on 400

## Testing
13/13 unit tests passing locally.

Related to #3590

### DIFF
--- a/src/query/MirrorNodeContractQuery.js
+++ b/src/query/MirrorNodeContractQuery.js
@@ -1,4 +1,17 @@
 import ContractFunctionParameters from "../contract/ContractFunctionParameters.js";
+import {
+    isRetryableNetworkError,
+    readErrorDetail,
+} from "../network/mirrorRestRetry.js";
+
+/** @constant {number} Initial retry backoff in milliseconds. */
+const INITIAL_BACKOFF_MS = 250;
+
+/** @constant {number} Maximum retry attempts for transient mirror errors. */
+const MAX_ATTEMPTS = 5;
+
+/** @constant {number} Maximum retry backoff in milliseconds. */
+const MAX_BACKOFF_MS = 8000;
 
 /**
  * @typedef {import("../contract/ContractId").default} ContractId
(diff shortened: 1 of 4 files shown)


Utkal059:
Hi @venilinvasilev — this addresses #3590 by wrapping `performMirrorNodeRequest` in an exponential backoff retry loop. Retries on HTTP 500/503/504 and transient network-level failures (up to 5 attempts, doubling from 250ms to 8s max). Non-retryable 4xx responses throw immediately without retrying. Also added `mirrorRestRetry.js` as a small utility module to keep the retry helpers reusable. 4 unit tests cover the main scenarios — retry on 503, retry on 504, exhaustion after 5 attempts, and no-retry on 400. All 13 tests pass locally. Happy to adjust retry parameters or the retryable status code set if needed.

src/query/MirrorNodeContractQuery.js
```diff
@@ -236,22 +249,62 @@ export default class MirrorNodeContractQuery {
...
+                if (
+                    response.status === 500 ||
+                    response.status === 503 ||
+                    response.status === 504
+                ) {
+                    lastError = new Error(
+                        `HTTP ${response.status}${errorDetail ? `: ${errorDetail}` : ""}`,
+                    );
+                    if (attempt < MAX_ATTEMPTS) {
+                        await _sleep(backoff);
```
ivaylonikolov7: why do we have this here. Can't we add it to the isRetryableNetworkError and this way we won't have repeating code?

ivaylonikolov7 (commented):
Also can we add setters and getter for max attempts and max backoff similar to the other transaction?

Utkal059:
Thanks for the suggestion! I have already added setMaxAttempts(), getMaxAttempts(), setMaxBackoff(), and getMaxBackoff() in the latest commit (8996795), following the same pattern as the other transactions. Let me know if anything needs adjusting.

src/query/MirrorNodeContractQuery.js
Utkal059: yes in the latest commit (8996795), I've consolidated the sleep logic into _fetchWithRetry so it no longer repeats. isRetryableNetworkError stays a pure boolean predicate — adding a side effect like await _sleep() inside it would make it harder to test and reason about. The if (attempt < MAX_ATTEMPTS) guard is kept to avoid an unnecessary delay after the final attempt. Let me know if you'd prefer a different structure!

src/query/MirrorNodeContractQuery.js
```diff
@@ -269,3 +407,11 @@ export default class MirrorNodeContractQuery {
         );
     }
 }
+
+/**
+ * @param {number} ms
+ * @returns {Promise<void>}
+ */
+function sleep(ms) {
+    return new Promise((resolve) => setTimeout(resolve, ms));
```
ivaylonikolov7, line 416: We already have the same function which you can import. No need to redeclare it.

src/query/MirrorNodeContractQuery.js
ivaylonikolov7: I don't like it that we have repeating code in the try and the catch blocks.

ivaylonikolov7:
Closing due to inactivity. You can reopen it if you want to continue working on it.

PR #4167, opened 2026-06-16
fix(deps): align protobufjs to 8.2.0 and reconcile pnpm lockfile

Fantomasa:
## Problem

The root `package.json` declares `protobufjs@8.2.0` (from #4084), but `pnpm.overrides`, `packages/proto`, and `pnpm-lock.yaml` were left at `8.0.1`. Under **pnpm v11** — which the `hiero-consensus-node` Block Node / SDK-TCK regression uses via `npm install -g pnpm` — a frozen install fails:

```
[ERR_PNPM_OUTDATED_LOCKFILE] ... pnpm-lock.yaml is not up to date with <ROOT>/package.json
  - protobufjs (lockfile: 8.0.1, manifest: 8.2.0)
```

This repo's own CI uses pnpm **v9.15.5**, which is lenient about the mismatch and reads overrides from `package.json`, so it never caught it.

## Change

Align everything on **protobufjs 8.2.0** and regenerate the lockfile:
- bump `protobufjs` to `8.2.0` in `packages/proto` and root `pnpm.overrides`
- regenerate `pnpm-lock.yaml` (protobufjs 8.2.0 throughout)

pnpm v11 no longer reads `pnpm.overrides`/`resolutions` from `package.json` and removed `onlyBuiltDependencies`, so the equivalent config is also declared in `pnpm-workspace.yaml` (its v11 home), while the `package.json` copies are kept for pnpm v9:
- `overrides`: `bn.js`, `protobufjs` (mirrors `package.json` for v11)
- `allowBuilds`: explicitly approve the dependency build scripts v11 gates by default

## Verification

| Check | Result |
|---|---|
| `pnpm install --frozen-lockfile` (pnpm 9.15.5) | ✅ |
| `pnpm install --frozen-lockfile` (pnpm 11.7.0) | ✅ no lockfile / config / ignored-builds errors |
| `task build` | ✅ |
| generated proto code (`proto.js`/`proto.d.ts`) | ✅ unchanged |
| node unit tests (`vitest run`) | ✅ |
| browser unit tests (`vitest run`) | ✅ 1604 passed |

protobufjs `8.0.1 → 8.2.0` is a minor bump with no declared breaking changes (the breaking `7→8` jump was already adopted).

### DIFF
--- a/package.json
+++ b/package.json
@@ -144,7 +144,7 @@
     },
     "pnpm": {
         "overrides": {
-            "protobufjs": "8.0.1"
+            "protobufjs": "8.2.0"
         }
     }
 }
--- a/packages/proto/package.json
+++ b/packages/proto/package.json
@@ -74,7 +74,7 @@
     "ansi-regex": "6.2.2",
     "ansi-styles": "6.2.3",
     "debug": "4.4.1",
-    "protobufjs": "8.0.1",
+    "protobufjs": "8.2.0",
     "strip-ansi": "7.1.2"
   },
   "publishConfig": {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -6,7 +6,7 @@ settings:
 
 overrides:
   bn.js: 5.2.3
-  protobufjs: 8.0.1
+  protobufjs: 8.2.0
 
 importers:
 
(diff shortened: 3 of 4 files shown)


ivaylonikolov7 (approved):

ivaylonikolov7:
Maybe a bump in the CI would be great, wdyt @Fantomasa

Fantomasa:
I'd prefer to do it as a separate follow-up PR rather than fold it in here, since the ordering matters: v11 stops reading pnpm.overrides/onlyBuiltDependencies from package.json, and this PR is precisely what moves that config into pnpm-workspace.yaml (overrides + allowBuilds). So this PR needs to land first, then the CI bump is safe.

PR #4230, opened 2026-07-10
fix: auto-estimate gas for ContractCallQuery

### LINKED ISSUE #2848: Inaccurate fee estimation
Cross linking [this issue here,](https://github.com/hashgraph/hedera-services/issues/4854) it seems still not clear if the issue is on the SDK or service side.

### Description

Executing `ContractCallQuery` with an automatic cost computation, the cost of the tx is sometimes miscalculated, causing the precheck to fail with `INSUFFICIENT_TX_FEE`:

```bash
transaction 0.0.11205@1675091501.521603607 failed precheck with status INSUFFICIENT_TX_FEE
```

This issue has been around for some years, and because of this the SDKs usually apply a multiplier (1.1) to work around that problem.

There are different open issues on this topic, and I think we should address it at the services level, not as a workaround in the SDKs.

We should already have the code that can do the correct computation, the precheck function. If that function can understand the fees are insufficient, the same should be possible for the estimate function.

Even though a precise estimation would not have been possible, as an extreme measure, I suggest increasing the result of the estimate function directly, lifting the burden to fix this to the several SDKs out there, and having at the same time a coherent response across the board.

### Steps to reproduce

Here is an example of a smart contract call where the cost estimation fails:

- Contract Identifier: 0.0.3350566
- Contract Source Code: https://github.com/hashgraph/guardian/blob/dbef5986469989da3b98dbd5a20ed2d3ef60344e/contracts/RetirementContract.sol
- Function Name: 'getPair
- First Parameter: 00000000000000000000000000000000002f9fde
- Second Parameter: 0000000000000000000000000000000000000000

I also deployed another new simple smart contract ([0.0.5133653](https://hashscan.io/testnet/contract/0.0.5133653)) you can test with this code:

```JavaScript
require("dotenv").config();
const { Client, ContractCallQuery } = require("@hashgraph/sdk");

async function main() {
  let client = Client.forTestnet();
  client.setOperator(process.env.OPERATOR_ID, process.env.OPERATOR_KEY);
  let contractId = "0.0.5133653";

  // query with explicit gas
  let query = await new ContractCallQuery()
    .setContractId(contractId)
    .setFunction("retrieve")
    .setGas(23474);

  let txResponse = await query.execute(client);
  console.log("Response with explicit gas:", txResponse.bytes);
  console.log();

  // query with computed gas
  query = await new ContractCallQuery()
    .setContractId(contractId)
    .setFunction("retrieve");

  txResponse = await query.execute(client); // ERROR: failed precheck with status INSUFFICIENT_GAS
  console.log("Response with estimated gas:", txResponse.bytes);
  process.exit(0);
}

main().catch((error) => {
  if (process.env.DEBUG) {
    console.log(error);
    process.exit(-1);
  }
});
```
Consider Hashscan is able to use the function without issue via the JSON RPC relay.

### Additional context

Tickets related to the same topic:

- https://github.com/hashgraph/hedera-sdk-js/issues/1425

### Hedera network

testnet

### Version

v0.34.1

### Operating system

None

ddeskov-limechain:
# fix: auto-estimate gas for `ContractCallQuery`

Fixes #2848

## What

A `ContractCallQuery` executed without `setGas()` failed with `PrecheckStatusError: ... INSUFFICIENT_GAS` (historically also reported as `INSUFFICIENT_TX_FEE`). The SDK now estimates the gas via the mirror node before the query is sent, and raises a clear, actionable error when estimation is not possible.

## Why

Per `services_contract_call_local.proto`, `gas` on `contractCallLocal` is **caller-supplied and required** — the node consumes the entire offered gas when computing the fee. The SDK never supplied it:

1. `query.execute(client)` with no query payment calls `Query._beforeExecute` (`src/query/Query.js:302`), which calls `this.getCost(client)`.
2. `getCost` wraps the query in a `CostQuery` and sends it with `responseType: COST_ANSWER`.
3. `CostQuery._makeRequestAsync` reuses the wrapped query's `_onMakeRequest(header)`, which serializes `gas: this._gas`. With `setGas()` never called, `_gas` is `null` and protobuf encodes it as `gas=0`.
4. `gas=0` is below the EVM intrinsic floor (~21000), so the node rejects **even the cost probe** with `INSUFFICIENT_GAS`.

The node behaves per spec; the bug was on the SDK side. The protocol delegates gas estimation to the caller, typically via a mirror node simulation — and the SDK already ships `MirrorNodeContractEstimateQuery` for exactly that.

Because the node rejects `gas=0` unconditionally (probing 0, 1, and 21000 all yield `INSUFFICIENT_GAS`, as verified in the issue thread), **no currently-working call path changes behavior**. A path that always failed now either succeeds (mirror node reachable) or fails with a far better message.

## Changes

**`src/contract/ContractCallQuery.js`**
- New private `_estimateGasIfNotSet(client)`: when `_gas` is `null` and `_contractId` is set, run `MirrorNodeContractEstimateQuery` (reusing the already-built calldata and forwarding the sender) and assign the result to `_gas`.
- Overrides `_beforeExecute` and `getCost` to call it. Both are needed:
  - `_beforeExecute` covers `execute()`, **including** the `setQueryPayment()` path that skips `getCost` internally (`src/query/Query.js:295`) but still needs gas for the real `contractCallLocal`.
  - `getCost` covers callers invoking `query.getCost(client)` directly, so gas is set before the `CostQuery` is serialized.
  - The helper is idempotent (`_gas != null` early-returns), so `execute()` → `_beforeExecute` → internal `getCost` estimates exactly once.
- If estimation fails for any reason (no mirror network, HTTP error, missing call data, non-finite/non-positive result), it throws an `Error` naming `setGas()` and including the underlying cause, instead of letting the opaque `INSUFFICIENT_GAS` precheck error surface.

**`src/query/MirrorNodeContractQuery.js`**
- Adds public `setFunctionParameters(Uint8Array)` for raw, already-ABI-encoded call data. The class previously only had `setFunction(name, params)`, which rebuilds the calldata; this lets a caller pass bytes straight through to `callData` without reaching into private fields.

**Deliberate non-choices**
- **No multiplier** on the mirror estimate. It is already padded (issue data: mirror estimated 25136 vs 23474 actually used), and `Query.getCost` already applies a ×1.1 buffer to the *cost* (`src/query/Query.js:169`). `_gas` equals the mirror node's returned estimate exactly.
- `_contractId == null` skips estimation entirely, so a malformed query still surfaces `INVALID_CONTRACT_ID` from the node as before.
- No changes to `CostQuery`, `Query._beforeExecute`, `ContractExecuteTransaction`, or any other query. No new dependencies.

## Notes for reviewers

Two behavior changes worth an explicit look:

1. **A mirror node HTTP request now happens when gas is unset.** When `setGas()` *is* called, no request is made and serialization is byte-for-byte unchanged.
2. **The error type changes for the no-gas case.** A query with neither gas nor call data can't be estimated, so it now throws a plain `Error` naming `setGas()` rather than a `PrecheckStatusError` carrying `INSUFFICIENT_GAS`. This path never succeeded before, so nothing working regresses — but it does mean code matching on `Status.InsufficientGas` for this case sees a different error.

This is also why two existing integration tests were updated rather than left alone:
- `"should return error when the gas is not set"` asserted `INSUFFICIENT_GAS` for precisely the scenario this PR fixes. (It was also vacuous — the assertion sat inside a `catch`, so it passed whether or not the query threw.) It is now the acceptance test: the query must succeed and return `"Hello from Hedera."`.
- `"should error when gas is not set"` sets neither gas nor a function, so there is no call data to estimate from. It now asserts the clear SDK error and is renamed `"should error when neither gas nor the function to call is set"`.

The other two tests in that file are untouched and unaffected: one sets gas explicitly, the other sets no contract ID.

`src/contract` → `src/query` is a new import direction; `dpdm` confirms it introduces no circular dependency.

## Verification

Unit tests, types, lint, and build all pass locally. **Integration tests were not run** — they require Solo / a local node, which was not available in my environment. The new mirror-node estimation path is therefore covered by unit tests and by direct exercise of the real call paths, but has not been executed end-to-end against a live network. Please confirm via CI.

- Full unit suite: **1889 passed**, 1 skipped, 150 files — no regressions.
  ```
  npx vitest run --config=test/vitest-node.config.ts     # == task test:unit:node
  ```
- **6 new unit tests** in `test/unit/ContractCallQuery.js` (estimates when unset; no mirror call when gas is set; no mirror call without a contract ID; forwards contract ID / call data / sender; clear error on estimation failure; clear error on an invalid estimate) and **1** in `test/unit/MirrorNodeContractQuery.js` for the new setter.
- The unit tests exercise the helper directly, so the real call paths were driven separately to confirm the wiring: `super.getCost` and `super._beforeExecute` each observe `gas = 25136`; an explicit `setGas(75000)` issues **zero** mirror requests; and `_onMakeRequest` outside `execute` still serializes `gas: null`, so `toBytes()` is unchanged.
- `npx tsc` — **0 errors under `src/`**.
- `npx dpdm src/index.js --circular true` — no circular dependency.
- `npx eslint "src/**/*.js"` — 180 errors / 112 warnings, **identical to the baseline on `main`**; this change adds none. `prettier --check` clean.
- `npx babel src -d lib --out-file-extension .cjs && npx rollup -c` (i.e. `task build:compile`) succeeds.

## Pull Request Readiness

Per the checklist in [`CONTRIBUTING.md`](./CONTRIBUTING.md#pull-request-readiness):

- [x] Includes tests to exercise the new behaviour
- [x] Code is documented, especially public and user-facing constructs
- [x] Local run of `task build` succeeds
- [x] Unit tests pass: `task test:unit:node`
- [ ] Integration tests pass with Solo: `task solo:setup && task test:integration:node` — **not run locally** (no Solo/local node available); relying on CI
- [x] Git commit message is detailed and includes context behind the change
- [x] If the change is related to an existing Bug Report or Feature Request, please include its issue number — #2848


ddeskov-limechain:
@Fantomasa please review

Fantomasa (approved):
LGTM

PR #4233, opened 2026-07-13
fix: remove react-native-get-random-values from cryptography package

ddeskov-limechain:
**Description**:

Remove `react-native-get-random-values` from the cryptography package so it stays environment-agnostic. React Native / Expo users must now import the polyfill themselves at the application entry point before importing the SDK.

* Remove `react-native-get-random-values` from the cryptography package dependencies
* Remove the polyfill import from `polyfills.native.js` (the `process.nextTick` shim is kept)
* Update the README React Native section to instruct users to import `react-native-get-random-values` at the app root before the SDK
* Add the explicit polyfill import to both React Native example apps, which previously relied on the SDK importing it
* Regenerate `pnpm-lock.yaml` (removal-only diff: the package and the `react-native` peer tree it pulled in)

**Related issue(s)**:

Fixes #3055

**Notes for reviewer**:

* All 162 cryptography node unit tests pass; the change does not touch browser or Node code paths (the polyfill import only existed in the React Native entry point).
* Audited the cryptography package for other React Native references: the remaining `*.native.js` files and the `react-native` field in `package.json` are the package's own environment entry points resolved by RN bundlers, not environment detection logic, so they stay.
* The lockfile was regenerated with pnpm 9.15.5 (the version pinned in CI) to keep the diff free of unrelated churn.
* This is a breaking change for React Native / Expo apps that relied on the implicit import — they must now add `import "react-native-get-random-values";` before importing the SDK (both example apps show the pattern).

**Checklist**

- [x] Documented (Code comments, README, etc.)
- [x] Tested (unit, integration, etc.)

PR #4229, opened 2026-07-09
refactor(cryptography): migrate primitives to Noble & Scure (#3054)

ddeskov-limechain:
## Description

Migrates the `@hiero-ledger/cryptography` package from its custom and
environment-specific cryptographic implementations to the audited,
zero-dependency **Noble** & **Scure** ecosystem.

The package previously carried three parallel implementations of most
primitives — a Node (`crypto`) version, a `*.browser.js` version
(SubtleCrypto / `crypto-js`), and a `*.native.js` version (`crypto-js` /
`spark-md5`) — selected at bundle time via `package.json` `browser` /
`react-native` maps. Noble/Scure are isomorphic (pure JS, WebCrypto only for
CSPRNG), so this collapses each primitive to a **single implementation** and
removes the environment branching entirely.

Closes #3054

## Migration

| Concern | Before | After |
| --- | --- | --- |
| SHA-256 | `crypto` / `crypto-js` / SubtleCrypto | `@noble/hashes/sha2` |
| HMAC (SHA-256/384/512) | `crypto` / `crypto-js` / SubtleCrypto | `@noble/hashes/hmac` |
| PBKDF2 | `crypto` / `crypto-js` / SubtleCrypto | `@noble/hashes/pbkdf2` |
| Ed25519 (sign/verify/keygen) | `tweetnacl` | `@noble/curves/ed25519` |
| secp256k1 | `@noble/curves` (pre-existing) | `@noble/curves` (unchanged) |
| CSPRNG | `tweetnacl` | `@noble/hashes/utils` (`randomBytes`) |
| BIP-32 (secp256k1 CKD) | hand-rolled | `@scure/bip32` (`HDKey`) |
| BIP-39 (mnemonic → seed) | hand-rolled | `@scure/bip39` (`mnemonicToSeed`) |
| AES-128 CTR/CBC | `crypto` / `crypto-js` / SubtleCrypto | `@noble/ciphers/aes` |
| MD5 (legacy PEM KDF) | `spark-md5` | `@noble/hashes/legacy` |
| Keccak-256 | ~430-line hand-roll | `@noble/hashes/sha3` |
| base64 | `Buffer` / `js-base64` | `@scure/base` |
| utf8 | `Buffer` / `utf8` | `TextEncoder` / `TextDecoder` |
| hex | `Buffer` | consolidated pure JS |

## Dependency changes (`packages/cryptography`)

[...]

PR #4226, opened 2026-07-08
fix(deps): force brace-expansion and fast-uri to patched versions

Fantomasa:
## What

Fixes two **high** severity transitive-dependency vulnerabilities via version-scoped pnpm overrides:

- [SNYK-JS-BRACEEXPANSION-17706650](https://security.snyk.io/vuln/SNYK-JS-BRACEEXPANSION-17706650) (CVE-2026-13149) — inefficient algorithmic complexity (ReDoS) in `brace-expansion`'s `expand()`. Fixed upstream in `1.1.16`, `2.1.2`, `5.0.7`.
- [SNYK-JS-FASTURI-17675102](https://security.snyk.io/vuln/SNYK-JS-FASTURI-17675102) — Interpretation Conflict in `fast-uri`. Fixed upstream in `3.1.3` / `4.0.1`.

This supersedes the automated Snyk PRs **#4215**, **#4214**, and **#4212**. Snyk splits them per "project" (`examples/`, `examples/react-native-example/`, `packages/cryptography/`), but those projects have no separate lockfile — they are all importers in the single root `pnpm-lock.yaml`, so one set of workspace-wide overrides covers every one of them.

## Why not just take the Snyk PRs

Both packages are **transitive** dependencies. The Snyk PRs bumped direct devDependencies to try to pull a fixed copy, and each **could not update the lockfile** (Snyk warned as much):

- #4215: `eslint` `^9` → `^10` (Snyk-flagged **high** breaking-change risk) + `typedoc` in `examples/package.json`.
- #4214: `eslint` `^9` → `^10` + `react-native` `0.81.5` → `0.85.0` in `examples/react-native-example/package.json`.
- #4212: `eslint` `9` → `10` + `@babel/cli` `7` → `8` + `@typescript-eslint/*` `8.55` → `8.56.1` in `packages/cryptography/package.json`.

Those approaches take on risky major bumps (ESLint, Babel, React Native) unrelated to the actual CVEs, and — because they only touch one manifest — leave the **other** vulnerable copies in the shared lockfile untouched (`brace-expansion` `1.1.12`/`1.1.15`/`2.0.2`/`2.1.1`/`5.0.2`/`5.0.6`, `fast-uri` `3.1.0`).

## Changes

Version-scoped pnpm overrides so **each release line moves only to its own patched release** (no cross-major forcing):

| package | before (in lockfile) | after |
|---------|----------------------|-------|
| brace-expansion 1.x | 1.1.12 / 1.1.15 | **1.1.16** |
[...]

PR #4257, opened 2026-07-21
chore(release): v2.86.1

### LINKED ISSUE #2173: Bundle issue in browser environment with esbuild

Fantomasa:
Stable release v2.86.1

### Fixed
- The SDK can now be bundled for the browser with esbuild-based tooling (Vite, modern React). A `browser` condition was added to the package `exports` map so bundlers no longer resolve the Node.js entry point and fail on `fs`/`util`/`tls`/`net` and `@grpc/grpc-js`. Fixes #2173. [#4256](https://github.com/hiero-ledger/hiero-sdk-js/pull/4256)

### Changed
- Migrated `@hiero-ledger/cryptography` primitives to the Noble and @scure libraries, collapsing the separate browser/native/default implementations into a single portable one and removing the `react-native-get-random-values` dependency. [#4229](https://github.com/hiero-ledger/hiero-sdk-js/pull/4229) [#4233](https://github.com/hiero-ledger/hiero-sdk-js/pull/4233)

### Security
- Forced `brace-expansion` and `fast-uri` to patched versions to address advisories. [#4226](https://github.com/hiero-ledger/hiero-sdk-js/pull/4226)
- Upgraded `vite` to `7.3.6` and forced `esbuild` to `>=0.28.1` to address advisories. [#4223](https://github.com/hiero-ledger/hiero-sdk-js/pull/4223)

## Sub-packages
- `@hiero-ledger/cryptography`: **1.19.0 → 1.20.0** — co-released so the Noble/@scure refactor above actually publishes (the published 1.19.0 predates it and cannot be overwritten).
- `@hiero-ledger/proto`: unchanged (`2.31.0` — only dev-tooling bumps since the last release).

## Notes
- `v2.86.0` was merged to `main` but never tagged/published, so npm's latest is still `2.85.0`; this release supersedes it and includes everything since `2.85.0`.
- Address books refreshed via `task update:addressbooks`.


ddeskov-limechain (approved):
LGTM

PR #4259, opened 2026-07-22
fix: revert "chore(deps): bump @noble/curves from 1.8.1 to 2.2.0"

Fantomasa:
## Description

Reverts #4239 (commit ef7580ab), restoring `@noble/curves` **1.8.1** in `packages/cryptography`, and adds guards so this class of break cannot recur silently.

### Why

`@noble/curves` v2 is **ESM-only** — `"type": "module"` and its exports map has no `require` condition. Our CJS build (`lib/*.cjs`, produced by Babel) transpiles `import ... from "@noble/curves/ed25519.js"` into `require(...)`, which throws at load time:

```
Error [ERR_REQUIRE_ESM]: require() of ES Module @noble/curves@2.2.0/ed25519.js
from @hiero-ledger/cryptography@1.20.0/lib/Ed25519PublicKey.cjs not supported.
```

This breaks CommonJS consumers of `@hiero-ledger/sdk@2.86.1` / `@hiero-ledger/cryptography@1.20.0` running under **ts-node** or **Node < 20.19** (native `require(esm)` in newer Node masks it elsewhere).

It was caught by the nightly [SDK TCK Regression in hiero-consensus-node](https://github.com/hiero-ledger/hiero-consensus-node/actions/runs/29881093619/job/88804320755): the TCK sdk-server (ts-node, CJS mode) resolves `@hiero-ledger/sdk@^2.85.0` → 2.86.1, crashed on startup, and all 122 TCK tests failed with `ECONNREFUSED 127.0.0.1:8544`.

### Why CI did not catch it

The `Common JS` workflow runs on Node 22/24, where `require()` of ESM works natively — so `require("@hiero-ledger/sdk")` succeeded in CI while ts-node and older Node consumers crash.

### Changes

- **Revert** `@noble/curves` to 1.8.1 (sources byte-identical to pre-#4239 state; minimal lockfile change validated with `pnpm install --frozen-lockfile`)
- **Strict CJS smoke test** in `common_js_test`: `node --no-experimental-require-module -e "require(…sdk…)"` + key generation/signing — reproduces the exact TCK failure on the broken state, passes with the revert
- **Dependabot ignore** for `@noble/curves >=2.0.0` so the same bump is not re-proposed until the CJS build strategy supports ESM-only deps

### Verification

- CJS smoke test: fails with `ERR_REQUIRE_ESM` on current main, passes with this branch
- Unit tests: cryptography 169/169, SDK 1896/1896 passing
- `pnpm install --frozen-lockfile` passes with the updated lockfile

### Related Issues

- Fixes the SDK TCK Regression failure in the nightly CITR Ext Test Suite (consensus-node run 29881093619)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,3 +16,7 @@ updates:
           - dependency-name: "*"
             update-types:
                 ["version-update:semver-minor", "version-update:semver-patch"]
+          # @noble/curves v2+ is ESM-only and breaks the CommonJS (lib/*.cjs)
+          # build of @hiero-ledger/cryptography — see #4259
+          - dependency-name: "@noble/curves"
+            versions: [">=2.0.0"]
--- a/common_js_test/Taskfile.yml
+++ b/common_js_test/Taskfile.yml
@@ -23,4 +23,8 @@ tasks:
         deps:
             - install
         cmds:
+            # Strict CJS smoke test: modern Node can require() ESM natively, which
+            # hides ESM-only dependencies sneaking into the lib/*.cjs builds. The
+            # flag restores strict behavior (what ts-node and Node <20.19 enforce).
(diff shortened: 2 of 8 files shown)


agadzhalov (approved):

andrewb1269 (approved):
Approve dependabot file.

PR #4265, opened 2026-07-22
fix: add missing length/bounds checks in Transaction.fromBytes

### LINKED ISSUE #4260: Add missing length/bounds checks in Transaction fromBytes (existing FIXMEs)
Two long-standing FIXMEs in `src/transaction/Transaction.js` around byte deserialization:

1. ~L266 — after decoding `TransactionList`, the code checks `list.length === 0` but the FIXME notes we should also verify the input bytes length is greater than 0, otherwise the empty-list check can misclassify the input.
2. ~L410 — `const body = bodies[0];` is accessed without a length check (FIXME on the line above). A subsequent `body == null` guard catches the undefined case, but the access should be made explicitly safe and the error message accurate.

These are potential production runtime/robustness gaps in a core deserialization path rather than test-only cleanup. Scope: add the checks, keep error messages descriptive, add unit tests with empty/malformed byte inputs.

ddeskov-limechain:
## Description

Closes #4260 — resolves the two long-standing FIXMEs in `Transaction.fromBytes` around byte deserialization.

### Root cause

1. `fromBytes` decoded the input as a `TransactionList` and treated `list.length === 0` as "the bytes are a single `Transaction`" — but empty input also produces an empty list, so empty bytes were misclassified and fell through to a misleading generic error at the end of the method.
2. `const body = bodies[0]` was accessed without a length check. The subsequent `body == null` guard happened to catch the `undefined` case, but the access was implicitly unsafe and the combined error message ("No transaction found in bytes or failed to decode TransactionBody") did not say which of the two things actually went wrong.

### What this PR changes

- `fromBytes` now rejects empty input up front with a clear error (`cannot deserialize a transaction from empty bytes`), which makes the empty-list → "not a TransactionList" inference sound.
- The `bodies[0]` access is now behind an explicit `bodies.length === 0` check, with separate, accurate error messages for "no transactions found in bytes" vs "failed to decode TransactionBody: data not set".
- Both FIXME comments are removed.

### Tests

New unit tests in `test/unit/Transaction.js`:
- `fromBytes` fails for empty bytes with the new explicit error
- `fromBytes` fails for a `TransactionList` entry with no body (`no transactions found in bytes`)
- `fromBytes` fails for malformed/garbage bytes

Full node unit suite passes (152 files, 1899 tests).

### DIFF
--- a/src/transaction/Transaction.js
+++ b/src/transaction/Transaction.js
@@ -242,6 +242,12 @@ export default class Transaction extends Executable {
      * @returns {Transaction}
      */
     static fromBytes(bytes) {
+        if (bytes.length === 0) {
+            throw new Error(
+                "cannot deserialize a transaction from empty bytes",
+            );
+        }
+
         /** @type {HieroProto.proto.ISignedTransaction[]} */
         const signedTransactions = [];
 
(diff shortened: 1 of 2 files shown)


Fantomasa (approved):
LGTM

PR #4266, opened 2026-07-22
test: document the 7 skipped custom fee integration tests

### LINKED ISSUE #4261: Fix or document 7 skipped custom fee integration tests
`test/integration/CustomFeesIntegrationTest.js` has 7 `it.skip` tests covering custom fee edge cases:

- L204 — transfer of custom fee schedule token to a non-associated fee-collecting account
- L299 — custom fractional fee greater than 1
- L328 — fee schedule update when no fee schedule is set
- L359 — fee schedule update signed with a key other than the fee schedule key
- L596 — clearing custom fees when none are present
- L654 — recursive custom fees
- L757 — more than 20 balance changes in a single transfer

For each: either fix the test so it runs, or document why the behavior is a known limitation and remove/annotate the test accordingly.

ddeskov-limechain:
## Description

Closes #4261 — each of the 7 `it.skip` tests in `test/integration/CustomFeesIntegrationTest.js` now carries a comment documenting exactly why it is skipped and (where applicable) what would be needed to re-enable it.

### Investigation results, per test

| Test | Why it can't pass as written |
| --- | --- |
| L204 — transfer to non-associated fee collector | The fee is a fixed **HBAR** fee; collectors only need token association for token-denominated fees (that case is covered by the active test "cannot create custom fee with un-associated token fee collector"). The update succeeds, and `INVALID_CUSTOM_FEE_COLLECTOR` signals a missing/invalid collector account, not a missing association. |
| L299 — fractional fee > 1 | Consensus nodes don't reject a fraction > 1 at creation time. Moreover, the expected `Status.InvalidCustomFractionalFeesSum` **does not exist in the SDK's Status enum** — it evaluates to `undefined`, so the assertion could never match a real status. |
| L328 — fee schedule update with no fee schedule set | Adding fees to a token that has a fee schedule key but no fees yet is a supported operation — the update succeeds. `CUSTOM_SCHEDULE_ALREADY_HAS_NO_FEES` only applies to clearing fees when none exist. |
| L359 — update signed with wrong key | Wrong expected status: the network returns `INVALID_SIGNATURE` for a missing fee-schedule-key signature; `INVALID_CUSTOM_FEE_SCHEDULE_KEY` is a creation-time status for an unusable key. The comment documents the path to re-enable (assert `INVALID_SIGNATURE`, verify against a network). |
| L596 — clearing custom fees when none present | Expanded the existing one-line note: this is no longer an error on current networks, so `CUSTOM_SCHEDULE_ALREADY_HAS_NO_FEES` is never returned. |
| L654 — recursive custom fees | The two-token fee cycle settles within the network's allowed custom fee nesting depth, so `CUSTOM_FEE_CHARGING_EXCEEDED_MAX_RECURSION_DEPTH` cannot be reproduced. |
| L757 — more than 20 balance changes | Current consensus node configurations allow more balance adjustments per transfer than the historical limit of 20 this test targeted, so `CUSTOM_FEE_CHARGING_EXCEEDED_MAX_ACCOUNT_AMOUNTS` is not returned. |

### Notes

- Comments-only change — no test logic was modified and no tests were removed, so reviewers can still see exactly what each scenario intended.
- The tests remain skipped rather than deleted so the documented expectations can be revisited if network behavior changes (e.g. the L359 status fix can be verified against solo/local-node in a follow-up).
- Prettier passes; the eslint `parserOptions.project` error on this file pre-exists on `main` for all integration test files linted individually.

### DIFF
--- a/test/integration/CustomFeesIntegrationTest.js
+++ b/test/integration/CustomFeesIntegrationTest.js
@@ -201,6 +201,14 @@ describe("CustomFees", function () {
         await env.close();
     });
 
+    // Skipped: this scenario is not an error on the network. The fee used here
+    // is a fixed HBAR fee (no denominating token), and a fee collector only
+    // needs to be associated with the token when the fee is denominated in
+    // that token — that case is covered by the active test
+    // "cannot create custom fee with un-associated token fee collector" below.
+    // The fee schedule update therefore succeeds, and
+    // INVALID_CUSTOM_FEE_COLLECTOR (which signals a missing or invalid
+    // collector account, not a missing association) is never returned.
     it.skip("User cannot transfer a custom fee schedule token to a fee collecting account that is not associated with it", async function () {
         const env = await IntegrationTestEnv.new();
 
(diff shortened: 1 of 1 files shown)


Fantomasa (approved):
LGTM

Issue #4105, opened 2026-05-29
Support gRPC-Web and Custom Auth Headers in Client for Hashsphere


Torakushi:
# Support gRPC-Web and Custom Auth Headers in WebClient|NodeClient for Hashsphere

## Summary

We are using the Hiero SDK `NodeClient` for Hedera Mainnet and Testnet to craft and broadcast transactions, query state, and perform other network operations.

However, we are running into compatibility issues when connecting to Hashsphere.

### 1. NodeClient does not work with Hashsphere consensus nodes

`NodeClient` relies on gRPC connections to communicate with consensus nodes. On Hashsphere, the consensus node gRPC endpoints are private and not publicly reachable.

Instead, Hashsphere exposes a gRPC-Web proxy for consensus node access. Since `NodeClient` does not support gRPC-Web, it cannot be used to communicate with Hashsphere consensus nodes.

As a result, applications must use `WebClient` to connect through the gRPC-Web proxy.

### 2. WebClient does not support custom authentication headers

Hashesphere requires an authentication token to be included in requests to consensus nodes via gRPC-Web.

Currently, `WebClient` does not provide a mechanism to inject custom headers (e.g. `Authorization`) into requests. This prevents applications from authenticating against Hashsphere's gRPC-Web endpoints.

## Current Workaround

Our current workaround is to create a custom channel implementation that extends `WebChannel` and injects the required headers before requests are sent.

This allows us to avoid monkey-patching `globalFetch`, but it requires maintaining custom SDK integration code for a fairly common use case.

## Proposed Enhancement

It would be helpful if the SDK provided:

1. Native gRPC-Web support where applicable.
2. A way to configure custom request headers for `Node/WebClient` / `Web/NodeChannel` (e.g. authentication tokens or other middleware requirements).

## Benefits

* Enables first-class Hashsphere support.
* Avoids custom channel implementations and SDK workarounds.
* Supports other deployments that require authenticated gRPC-Web access.
* Provides greater flexibility for enterprise and managed network environments.

PR #4268, opened 2026-07-22
feat: add mirror node support to WebClient and NativeClient

### LINKED ISSUE #4263: Mirror node support for WebClient and NativeClient
Mirror node queries (e.g. topic subscriptions, mirror-node-backed queries) are currently only supported on the Node client. `WebClient` and `NativeClient` consumers have no mirror node support, limiting SDK reach for dApp (browser) and mobile developers.

Scope: evaluate transport options (gRPC-Web / REST) for mirror node access from browser and React Native environments, then implement mirror network support in `WebClient` and `NativeClient`. Related: #4105 (gRPC-Web + custom auth headers) may share transport groundwork.

ddeskov-limechain:
## Description

Closes #4263 — mirror node queries (topic subscriptions and other mirror-backed streaming queries) now work on `WebClient` and `NativeClient`, not just the Node client.

### Transport evaluation

The config plumbing already existed (`WebMirrorNetwork` constants including the local-node gRPC-Web mirror port `5551`, `setMirrorNetwork` on both clients) — only `_createMirrorNetworkChannel()` threw `"mirror support is not supported in browsers/native"`. Both clients already speak gRPC-Web for the consensus network (`WebChannel`/`NativeChannel`), so gRPC-Web was chosen over REST for the mirror transport too: it reuses the existing framing helpers, works with the same proxy setup users already need, and supports true server streaming (which REST polling would only approximate). This also lays the transport groundwork shared with #4105.

### What this PR adds

- **`src/channel/GrpcWebFrameParser.js`** — shared incremental parser for gRPC-Web framed bodies (frames may be split across or share transport chunks), plus trailers parsing (`grpc-status`/`grpc-message`).
- **`src/channel/WebMirrorChannel.js`** — browser mirror channel: binary gRPC-Web (`application/grpc-web+proto`) over a streaming `fetch` response body, with `AbortController`-based cancellation and header-level error detection for proxies that fail before the stream starts.
- **`src/channel/NativeMirrorChannel.js`** — React Native mirror channel: gRPC-Web **text** (base64) mode over `XMLHttpRequest` incremental `responseText`, since React Native's `fetch` cannot stream response bodies. Includes a `Base64StreamDecoder` that handles both continuous base64 and independently padded per-frame segments, split at arbitrary chunk boundaries.
- Wires both into `_createMirrorNetworkChannel()` and tightens the client JSDoc type params from `Client<Channel, *>` to the concrete mirror channel types.

Streams that end without a trailers frame are surfaced as `UNAVAILABLE` (code 14), which `TopicMessageQuery`'s retry handler treats as retryable, so subscriptions reconnect after dropped connections.

### Tests

New unit tests in `test/unit/GrpcWebMirrorChannels.js` (12 tests):
- Frame parser: single/multiple frames per chunk, frames split across chunks including mid-header, trailers with/without status.
- `Base64StreamDecoder`: arbitrary split boundaries and mid-stream padding.
- `WebMirrorChannel` against a stubbed streaming `fetch`: message delivery + clean end, non-zero trailer status as `MirrorError`, missing trailers → retryable code 14.
- `NativeMirrorChannel` against a fake `XMLHttpRequest`: streamed base64 chunks decode to the right frames, correct URL/method, missing trailers → code 14.

Full node unit suite passes (153 files, 1908 tests). `tsc` and eslint report no issues for the new/changed files.

### Notes

- As with the existing unary gRPC-Web transport, the configured mirror address must be gRPC-Web capable (e.g. Envoy in front of the mirror node gRPC API, or the local-node mirror gRPC-Web port). The public `*.mirrornode.hedera.com:443` endpoints speak native gRPC; browser/native users point at a proxy, exactly as they already do for consensus nodes.


Fantomasa (dismissed):
LGTM

src/channel/NativeMirrorChannel.js
```diff
@@ -0,0 +1,247 @@
...
+
+        const consumeResponseText = () => {
+            const text = xhr.responseText;
+
+            if (text.length > seenLength) {
+                const chunk = text.slice(seenLength);
+                seenLength = text.length;
+                handleFrames(frameParser.feed(base64Decoder.feed(chunk)));
+            }
+        };
```
Fantomasa, line 159: **Bug: pre-stream proxy failures are fed into the base64 decoder.**

Unlike `WebMirrorChannel`, nothing checks `xhr.status` or the `grpc-status`/`grpc-message` response headers. The default mirror endpoints hit exactly this: `testnet.mirrornode.hedera.com:443` answers a gRPC-Web POST with HTTP 415, headers `grpc-status: 13` / `grpc-message: Content-Type 'application/grpc-web-text' is not supported`, and a **plain-text** body (verified live). That body isn't base64, so on RN's strict decode path every `onprogress` throws uncaught; `onload` then reports a retryable 14 → ~20 futile retries ending in a misleading `stream ended without receiving trailers` instead of the real status (13, correctly non-retryable).

Suggest checking headers before consuming the body (mirrors the web channel's header handling), and guarding the decode so undecodable bodies fail the stream through `error()` instead of throwing inside an XHR event handler — this also contains the `decodeURIComponent` edge in `parseGrpcWebTrailers`:

```suggestion
        let headersChecked = false;

        // Errors the proxy reports before the stream starts arrive as an
        // HTTP error status or grpc-status response headers instead of a
        // trailers frame
        const checkResponseHeaders = () => {
            if (headersChecked) {
                return;
            }
            headersChecked = true;

            const grpcStatus = xhr.getResponseHeader("grpc-status");

            if (
                xhr.status < 200 ||
                xhr.status >= 300 ||
                (grpcStatus != null && grpcStatus !== "0")
            ) {
                ended = true;
                this._requests.delete(xhr);
                xhr.abort();
                error({
                    code: grpcStatus != null ? parseInt(grpcStatus, 10) : 2,
                    details:
                        xhr.getResponseHeader("grpc-message") ??
                        `received HTTP status ${xhr.status}`,
                });
            }
        };

        const consumeResponseText = () => {
            const text = xhr.responseText;

            if (text.length > seenLength) {
                const chunk = text.slice(seenLength);
                seenLength = text.length;

                try {
                    handleFrames(frameParser.feed(base64Decoder.feed(chunk)));
                } catch (err) {
                    // A body that fails to decode (e.g. a plain-text error
                    // page) must fail the stream instead of throwing
                    // uncaught inside an XHR event handler
                    if (!ended) {
                        ended = true;
                        this._requests.delete(xhr);
                        xhr.abort();
                        error(
                            err instanceof Error
                                ? err
                                : new Error(String(err)),
                        );
                    }
                }
            }
        };
```

Needs the two small wiring changes below in `onprogress`/`onload`.

src/channel/NativeMirrorChannel.js
```diff
@@ -0,0 +1,247 @@
...
+        this._requests.add(xhr);
+
+        xhr.open("POST", url);
+        xhr.responseType = "text";
+        xhr.setRequestHeader("content-type", "application/grpc-web-text");
+        xhr.setRequestHeader("accept", "application/grpc-web-text");
+        xhr.setRequestHeader("x-user-agent", `${SDK_NAME}/${SDK_VERSION}`);
+        xhr.setRequestHeader("x-grpc-web", "1");
+
+        xhr.onprogress = consumeResponseText;
```
Fantomasa: Wire the header check in before consuming (see the `checkResponseHeaders` suggestion above):

```suggestion
        xhr.onprogress = () => {
            checkResponseHeaders();

            if (!ended) {
                consumeResponseText();
            }
        };
```

Fantomasa (commented):
Solid PR overall. Two issues in `NativeMirrorChannel` to fix before merge — the inline comments carry committable suggestions, verified locally (new regression test red→green, 15/15 tests pass, `tsc` clean):

1. **`Base64StreamDecoder` breaks when a chunk boundary splits `==` padding.** React Native's strict base64 polyfill throws uncaught inside `onprogress` and the chunk's bytes are lost. Node and Chromium decoders are lenient, so CI can't catch this.
2. **No HTTP status / `grpc-status` header check before decoding the body.** Pre-stream proxy errors — e.g. the default public endpoints answer gRPC-Web with 415 + `grpc-status: 13` + a plain-text body — are fed into the base64 decoder: uncaught throws, ~20 futile retries, and a misleading final error instead of failing fast like `WebMirrorChannel`.

Note: the three source suggestions belong together (the `onprogress`/`onload` ones reference `checkResponseHeaders`) — apply them as one batch.

Non-blocking follow-ups, fine as separate issues: map web HTTP errors to retryable codes (currently code 2, never retried — a transient proxy 503 kills the subscription for good), a time-to-first-byte deadline, treat header `grpc-status: 0` as a clean `end()`, guard `decodeURIComponent` in trailers parsing, dedupe `handleFrames` into `GrpcWebFrameParser.js`.

src/channel/NativeMirrorChannel.js
```diff
@@ -0,0 +1,247 @@
...
+
+        xhr.onload = () => {
+            this._requests.delete(xhr);
+            consumeResponseText();
```
Fantomasa: Same for the final flush on `onload` (the missing-trailers block that follows stays unchanged):

```suggestion
        xhr.onload = () => {
            this._requests.delete(xhr);
            checkResponseHeaders();

            if (!ended) {
                consumeResponseText();
            }
```

src/channel/NativeMirrorChannel.js
```diff
@@ -0,0 +1,247 @@
...
+        const parts = [];
+
+        for (;;) {
+            const padding = this._pending.indexOf("=");
+
+            if (padding !== -1) {
+                let segmentEnd = padding + 1;
+                if (this._pending[segmentEnd] === "=") {
+                    segmentEnd += 1;
+                }
```
Fantomasa: **Bug (React Native): a chunk boundary that splits `==` padding crashes the strict decoder.**

Decoding eagerly on the first `=` means a chunk ending between the two `=` of a `==` pad (arbitrary XHR flush points; `==` occurs whenever a frame's length ≡ 1 mod 3) decodes `"…AQ="` now and a lone `"="` on the next feed. js-base64's strict polyfill — the path taken on RN without global `atob`/`Buffer` — throws `TypeError: malformed base64.` on both, uncaught inside `onprogress`, and the chunk's bytes are lost because `seenLength` has already advanced.

Node (lenient `Buffer` path) and Chromium (`atob(_tidyB64(...))` strips `=`) both tolerate it, so only the RN production path fails — tests below pin it.

Compute the segment length from the `=` offset and wait for the full group:

```suggestion
                // A '=' at offset p means the segment is p rounded up to
                // the next multiple of 4 characters long. Wait until the
                // whole group has arrived — strict decoders (React
                // Native's base64 polyfill) throw on partial padding.
                const segmentEnd = padding + (4 - (padding % 4));

                if (this._pending.length < segmentEnd) {
                    break;
                }
```

test/unit/GrpcWebMirrorChannels.js
```diff
@@ -0,0 +1,304 @@
...
+    });
+
+    it("decodes independently padded segments mid-stream", function () {
+        const decoder = new Base64StreamDecoder();
+        // two frames base64-encoded separately, as proxies may do
+        const encoded = b64([1, 2, 3, 4, 5]) + b64([6, 7]);
+
+        expect([...decoder.feed(encoded)]).to.deep.equal([1, 2, 3, 4, 5, 6, 7]);
+    });
+});
```
Fantomasa, line 143: Regression tests for the padding-split fix. The first fails against the current decoder (`expected Uint8Array[ 65, 66, 67, 1 ] to have a length of +0 but got 4` — it decoded a group whose padding hadn't fully arrived, the exact input that makes RN's strict base64 polyfill throw):

```suggestion
    });

    it("holds back a padded group until all padding characters arrive", function () {
        const decoder = new Base64StreamDecoder();
        // "QUJDAQ==" — while only one of the two '=' has arrived the
        // group must not be decoded: strict decoders (React Native's
        // base64 polyfill) throw on partial padding
        const encoded = b64([65, 66, 67, 1]);
        expect(encoded.endsWith("==")).to.equal(true);

        expect(decoder.feed(encoded.slice(0, -1))).to.have.length(0);
        expect([...decoder.feed(encoded.slice(-1))]).to.deep.equal([
            65, 66, 67, 1,
        ]);
    });

    it("decodes '=='-padded segments split at every boundary", function () {
        const encoded = b64([65, 66, 67, 1]) + b64([6, 7]);

        for (let split = 0; split <= encoded.length; split += 1) {
            const decoder = new Base64StreamDecoder();
            const out = [
                ...decoder.feed(encoded.slice(0, split)),
                ...decoder.feed(encoded.slice(split)),
            ];
            expect(out).to.deep.equal([65, 66, 67, 1, 6, 7]);
        }
    });
});
```

test/unit/GrpcWebMirrorChannels.js
```diff
@@ -0,0 +1,304 @@
...
+
+describe("NativeMirrorChannel", function () {
+    class FakeXMLHttpRequest {
+        static instances = [];
+
+        constructor() {
+            FakeXMLHttpRequest.instances.push(this);
+            this.responseText = "";
+            this.headers = {};
+        }
```
Fantomasa, line 250: The fake needs `status`/`getResponseHeader` for the header-check path (the defaults keep the existing tests green):

```suggestion
            this.headers = {};
            this.status = 200;
            this.responseHeaders = {};
        }

        getResponseHeader(name) {
            return this.responseHeaders[name.toLowerCase()] ?? null;
        }
```

test/unit/GrpcWebMirrorChannels.js
```diff
@@ -0,0 +1,304 @@
...
+                () => reject(new Error("should not end cleanly")),
+            );
+        });
+
+        FakeXMLHttpRequest.instances[0].onload();
+
+        const err = await done;
+        expect(err.code).to.equal(14);
+    });
+});
```
Fantomasa, line 367: Covers the fail-fast path against the response shape the public mirror endpoints actually return (415 + `grpc-status: 13` headers + plain-text body):

```suggestion
    });

    it("fails fast with the header grpc-status when the proxy rejects the request", async function () {
        const channel = new NativeMirrorChannel("mirror.example.com:443");

        const done = new Promise((resolve, reject) => {
            channel.makeServerStreamRequest(
                "ConsensusService",
                "subscribeTopic",
                new Uint8Array([0]),
                () => {},
                resolve,
                () => reject(new Error("should not end cleanly")),
            );
        });

        // e.g. the public mirror endpoints answer gRPC-Web requests with
        // 415 + grpc-status headers and a plain-text (non-base64) body
        const xhr = FakeXMLHttpRequest.instances[0];
        xhr.status = 415;
        xhr.responseHeaders["grpc-status"] = "13";
        xhr.responseHeaders["grpc-message"] = "Content-Type not supported";
        xhr.push("Content-Type 'application/grpc-web-text' is not supported");

        const err = await done;
        expect(err).to.deep.equal({
            code: 13,
            details: "Content-Type not supported",
        });
        expect(xhr.aborted).to.equal(true);
    });
});
```

Fantomasa (approved):
LGTM

Issue #3952, opened 2026-04-23
[Good First Issue]: exclude fully deprecated classes from vitest coverage reports


venilinvasilev:
### 🆕🐥 First-Time Friendly

This issue is especially welcoming for people who are new to contributing to the **Hiero JS SDK**.

We know that opening your first pull request can feel like a big step.  
Issues labeled **Good First Issue** are designed to make that experience easier, clearer, and more comfortable.

No prior knowledge of Hiero, Hedera, or distributed ledger technology is required.  
Just a basic familiarity with JavaScript and Git is more than enough to get started.

> **📢 Note (2026-08-12):** this issue has been rewritten to match current `main`. Since it was opened: **7 of the originally listed 12 files have been removed from the SDK entirely**, there are now **six** vitest configs instead of four (two new `dual-mode` integration configs), and four of the configs already carry a `coverage.exclude` array (for the generated addressbook files) — so the deprecated files are **appended** to it rather than adding a new array. The list below was re-verified against `main` by scanning every `src/**/*.js` file for a class-level `@deprecated` tag.

### 👾 Description of the Issue

Several classes in the SDK are **fully deprecated** — the entire class is marked `@deprecated`, not just individual methods. These files will never receive new tests, so including them in coverage reports unfairly lowers the overall coverage percentage.

The following **5 files** are fully deprecated and should be excluded from coverage:

| File | Reason |
|------|--------|
| `src/contract/ContractStateChange.js` | `@deprecated - Use mirror node for contract traceability instead` |
| `src/contract/StorageChange.js` | `@deprecated - Use mirror node for contract traceability instead` |
| `src/token/TokenNftsUpdateTransaction.js` | `@deprecated - Use TokenUpdateNftsTransaction instead` |
| `src/transaction/NodeAccountIdSignatureMapLegacy.js` | `@deprecated` |
| `src/transaction/SignatureMapLegacy.js` | `@deprecated` |

> **Note:** Files with only *some* deprecated methods (e.g., `Client.js`, `Mnemonic.js`) should **not** be excluded — the class itself is still active.

### 💡 Proposed Solution

Add the 5 files above to the `coverage.exclude` array of all **six** vitest config files:

[...]

PR #4269, opened 2026-07-22
fix: remove long-deprecated classes and methods

### LINKED ISSUE #4264: Deprecation sweep: remove long-deprecated classes and methods
Candidates for removal in the next major-compatible window (all long-deprecated, verified still present on main):

- `src/account/LiveHashAddTransaction.js`, `src/account/LiveHashDeleteTransaction.js` (+ `LiveHash.js`, `LiveHashQuery.js`)
- `src/system/SystemDeleteTransaction.js`, `src/system/SystemUndeleteTransaction.js`
- `src/EthereumFlow.js`
- `AccountId.toSolidityAddress()` / `fromSolidityAddress()` (src/account/AccountId.js:292) and equivalents on other entity IDs
- `src/account/AccountAllowanceAdjustTransaction.js`

Scope: confirm deprecation notices have been in place for the required window, remove the classes/methods, update exports and docs. Related: #3952 (exclude deprecated classes from coverage) becomes moot for removed classes.

ddeskov-limechain:
## Description

Closes #4264 — removes the long-deprecated classes and methods listed in the issue. **Breaking change** (commit is marked `chore!`), intended for the next major-compatible release window. Net −3,024 lines.

### Removed

| API | Deprecated since | Replacement |
| --- | --- | --- |
| `LiveHashAddTransaction`, `LiveHashDeleteTransaction`, `LiveHashQuery`, `LiveHash` | 2022 (live hash operations have returned `NOT_SUPPORTED` from consensus nodes for years) | none — feature removed from HAPI |
| `AccountInfo.liveHashes` | carried the always-empty `LiveHash` data | none |
| `SystemDeleteTransaction`, `SystemUndeleteTransaction` | 2022 | none — privileged ops removed from HAPI |
| `EthereumFlow` | 2022 | `EthereumTransaction` (jumbo transactions) |
| `AccountAllowanceAdjustTransaction` | January 2022 | `AccountAllowanceApproveTransaction` / `AccountAllowanceDeleteTransaction` |
| `toSolidityAddress()` / `fromSolidityAddress()` on `AccountId`, `ContractId`, `DelegateContractId`, `FileId`, `TokenId`, `TopicId` | 2022+ | `toEvmAddress()` / `fromEvmAddress()` |

### Deliberately kept — needs a follow-up

- **`ScheduleId.toSolidityAddress()` / `fromSolidityAddress()`**: these were never annotated `@deprecated` and `ScheduleId` has **no `toEvmAddress()` replacement**, so removing them now would strip a non-deprecated API with no migration path. They should first gain `toEvmAddress()` + a deprecation notice, then be removed in a later window.
- **`EntityIdHelper.toSolidityAddress` / `fromSolidityAddress`**: not deprecated, publicly exported via `EntityIdHelper`, and still the implementation the remaining callers (`ScheduleId`) use.

### Ripple effects handled

- Removed the 8 exports from `src/exports.js` (shared by node/browser/native entry points) and the now-unused `hex`/`isLongZeroAddress` imports in `AccountId`/`ContractId`.
- Deleted the corresponding integration tests (`LiveHashIntegrationTest`, `SystemIntegrationTest`, `EthereumFlowIntegrationTest`) and `EthereumFlowMocking` unit test.
- tck: dropped `liveHashes` from `mapAccountInfoResponse` and the `LiveHashResponse` interface.
- Migrated remaining `toSolidityAddress()` call sites to `toEvmAddress()`: `EthereumTransactionIntegrationTest`, the frontend jumbo example, and two `ContractId` unit tests (equivalent behavior for the evm-address and shard-0/realm-0 long-zero cases they exercise).
- Registry note: bytes containing `systemDelete`/`systemUndelete`/`cryptoAddLiveHash`/`cryptoDeleteLiveHash` transaction bodies can no longer be parsed by `Transaction.fromBytes` (their `TRANSACTION_REGISTRY` entries are gone) — inherent to removing the classes.
- `RequestType`/`Status` enum names (protocol-level constants) are untouched.

### Verification

- Full node unit suite passes (151 files, 1891 tests).
- `tsc` (lint:types) clean for all changed files; eslint clean (including removing now-unused `deprecation/deprecation` disable directives in `exports.js`); Prettier clean.
- Repo-wide grep confirms no remaining references to any removed class in `src/`, `test/`, `tck/`, `examples/`, or `common_js_test/`.

Related: #3952 (exclude deprecated classes from coverage) is now moot for the removed classes.


Fantomasa (approved):
LGTM

Reccetech:
Question on the `SystemDeleteTransaction` / `SystemUndeleteTransaction` removal here.

These got pulled in v2.87.0 as part of this `remove long-deprecated stuff` sweep, lumped in with genuinely-dead APIs like `LiveHash*` and `EthereumFlow`. The removal table justifies them as *""privileged ops removed from HAPI""* — but that doesn't seem right. The protobufs are still present (`services_system_delete.proto` / `services_system_undelete.proto` are still in the repo), and consensus nodes still support these **for files** (deprecated only for contracts). The old classes just carried a bare `@deprecated` / ""Deprecated: Do not use."" tag with no rationale or replacement, so I think the ""removed from HAPI"" reasoning may have been an incorrect assumption.

Also worth flagging: JS is the only SDK that dropped these — Java, Go, Rust, Swift, and C++ all still have them. And per the PR notes, JS can now no longer even parse `systemDelete` / `systemUndelete` bodies via `Transaction.fromBytes` (the `TRANSACTION_REGISTRY` entries are gone), which matters for tools like the Transaction Tool that may receive such transactions.

So — was there a specific reason these two were removed, or did they get swept up with the genuinely-dead APIs? Given CN still honors the file variant and every other SDK retains them, should we add them back (at least the file path)?

PR #4296, opened 2026-07-28
feat(cryptography): bundle ESM-only deps into CJS build and migrate to @noble v2

Fantomasa:
## Description

Dependabot's `@noble/ciphers` 1.2.1 → 2.2.0 bump (#4272) failed CI because @noble packages are **ESM-only from v2 onward** — the same root cause that forced the `@noble/curves` 2.2.0 revert (#4259) and the emergency 2.86.2 release. The per-file babel CJS transpile turns imports into `require("@noble/...")` calls, which throw `ERR_REQUIRE_ESM` on Node <20.19.

Instead of ignoring @noble v2+ in Dependabot forever, this PR makes the CJS artifact immune to ESM-only dependencies and migrates all @noble/@scure packages to v2:

### Build change (`packages/cryptography`)
- Replace `babel src -d lib` with an **esbuild bundle**: `lib/index.cjs` now inlines `@noble/*` and `@scure/*`; every other dependency stays external and resolves from `node_modules`. The bundle's only external requires are `asn1js`, `bignumber.js`, and `forge-light` — all CJS-compatible.
- ESM (`src/`) and react-native entries are untouched; `tsc` still emits per-file `.d.ts` into `lib/`.
- Removed the now-unused babel build machinery (`.babelrc.json`, `build/replace-module-paths.cjs`, babel devDependencies).

### Dependency migration
`@noble/ciphers` 1.2.1 → 2.2.0, `@noble/curves` 1.8.1 → 2.2.0, `@noble/hashes` 1.8.0 → 2.2.0, `@scure/base` 1.2.4 → 2.2.0, `@scure/bip32` 1.6.2 → 2.2.0 (scure v2 pins the noble v2 versions, so the tree dedupes to a single copy of each).

v2 API changes applied:
- subpath imports need explicit `.js` suffixes
- `secp256k1.sign`/`verify` get `prehash: false` — inputs are already keccak256 digests, and **v2 defaults to `prehash: true`** (sha256 of the input), which would have silently changed every signature
- `utils.randomPrivateKey` → `utils.randomSecretKey`, `ProjectivePoint.fromHex`/`toRawBytes` → `Point.fromBytes`/`toBytes`, `Signature.fromCompact` → `Signature.fromBytes(bytes, 'compact')`
- `verify()` keeps returning `false` on malformed signature bytes (v2 throws where v1 returned `false`)

### Dependabot
Removed the `@noble/curves >=2.0.0` ignore rule added after #4259 — ESM-only majors can no longer break the CJS artifact. Note this bundle immunizes only the **module-format** axis: a future `@noble` major with API changes will still produce a failing Dependabot PR, but it will fail in the unit tests (golden-byte vectors) — the correct gate — rather than at runtime in the TCK.

### Behavior change
`PublicKey.verify()` for ECDSA now rejects signatures that are not exactly 64 bytes. Previously a 65-byte recoverable signature (`r || s || v`) was accepted by reading only its first 64 bytes; `@noble/curves` v2 rejects any non-64-byte compact signature, and `verify()` returns `false` (it does not throw — the boolean contract is preserved). Hedera ECDSA signatures are 64-byte compact and no internal caller passes recoverable signatures, so this affects only external callers that passed a 65-byte signature directly. Pinned by a new regression test in the `EcdsaPrivateKey` unit suite.

## Verification
- Byte-for-byte v1↔v2 parity checked for every primitive used: secp256k1 sign/verify/recovery, ed25519, AES-CBC/CTR, md5/sha2/keccak/hmac/pbkdf2, base64, BIP-32 derivation
- `packages/cryptography`: build (prettier, dpdm, tsc, eslint) + 170 unit tests pass on node **and** browser
- Root SDK: build (incl. UMD browser bundles) + 1899 unit tests pass (includes golden-byte signing vectors)
- `common_js_test`: strict CJS smoke test (`--no-experimental-require-module`, the tripwire that caught #4259) + mocha suite pass
- CJS bundle target set to `node18` to match the package's declared `engines` floor (also bumped from a stale `>=12` to `>=18`)

Closes #4272

## Related Issues
- #4259 (revert of the @noble/curves 2.2.0 bump)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


.github/dependabot.yml
```diff
@@ -16,7 +16,3 @@ updates:
           - dependency-name: "*"
             update-types:
                 ["version-update:semver-minor", "version-update:semver-patch"]
-          # @noble/curves v2+ is ESM-only and breaks the CommonJS (lib/*.cjs)
-          # build of @hiero-ledger/cryptography — see #4259
-          - dependency-name: "@noble/curves"
```
ddeskov-limechain, line 21: This removes the guard, but CI cannot catch the bug it was guarding against.

- `build.yml` and `common_js.yml` both run Node **22 and 24** only. On those versions `require(esm)` works natively, so `ERR_REQUIRE_ESM` never happens.
- `common_js_test/Taskfile.yml` runs mocha **without** `--no-experimental-require-module`. The description says that flag is the tripwire that caught #4259 — but it was used locally, not in CI.
- The cryptography unit tests run against `src/` (`test/vitest-node.config.ts`), so nothing tests `lib/index.cjs` either.

So after this merges, any new ESM-only dependency that is not `@noble`/`@scure` breaks consumers on Node 18–20.18 and CI stays green.

One line fixes it, in `common_js_test/Taskfile.yml`:

```yaml
    test:
        deps:
            - install
        env:
            NODE_OPTIONS: --no-experimental-require-module
        cmds:
            - ./node_modules/.bin/mocha --exit --inline-diffs -r @babel/register -r chai/register-expect.js "src/test.js" --timeout 120000
```

(I checked `NODE_OPTIONS=--no-experimental-require-module` is accepted on Node 24.)

ddeskov-limechain (commented):
Read the linked issue (#4272) and the full diff. The v2 migration itself checks out — I verified against the real `@noble/curves@2.2.0` typings that `prehash` defaults to `true` in v2 (so `prehash: false` is required and correct), that `sign` returns bytes, and that `Point.fromBytes`/`toBytes`, `Signature.fromBytes(sig, "compact")`, `randomSecretKey` and the `CHash`/`Cipher` type paths all exist. Also confirmed esbuild does not run `onResolve` on the entry point (so the `/^[^./]/` filter is safe), that only `asn1js`, `bignumber.js` and `forge-light` stay external, and that noble's MIT banners survive in the bundle.

Four comments inline. Only the first one I would want fixed before merge.

packages/cryptography/package.json
```diff
@@ -35,18 +35,18 @@
         "src/"
     ],
     "engines": {
-        "node": ">=12.0.0"
+        "node": ">=18.0.0"
```
ddeskov-limechain: All three new noble deps declare `engines: { node: ">= 20.19.0" }` (`@noble/ciphers`, `@noble/curves`, `@noble/hashes` — checked on npm).

The CJS path is fine: noble is inlined and the target is `node18`. But the `import` condition points at `src/index.js`, which loads `@noble/curves@2.2.0` directly. On Node 18 that gives an `EBADENGINE` warning on install, and it is not tested anywhere — CI is 22/24 only.

Either bump the floor, or the `>=18` claim (and the comment in `bundle-cjs.cjs`) should say it only holds for the CJS entry.

```suggestion
        "node": ">=20.19.0"
```

packages/cryptography/src/primitive/ecdsa.js
```diff
@@ -99,10 +100,14 @@ export function verify(keydata, message, signature) {
     const msg = hex.encode(message);
     const data = hex.decode(keccak256(`0x${msg}`));
 
-    const r = BigInt("0x" + hex.encode(signature.subarray(0, 32)));
-    const s = BigInt("0x" + hex.encode(signature.subarray(32, 64)));
-
-    return secp256k1.verify({ r, s }, data, keydata);
+    try {
+        return secp256k1.verify(signature, data, keydata, { prehash: false });
```
ddeskov-limechain: The 65-byte rejection is a real behaviour change on public API (`PublicKey.verify()`), and `CHANGELOG.md` is not updated. The repo keeps a manual Keep a Changelog with per-PR links, so this needs an entry under `### Changed` for the next release.

If you would rather not break it at all, keeping v1 behaviour is two lines:

```suggestion
    // Accept 65-byte recoverable signatures (r || s || v), as v1 did.
    const sig = signature.length === 65 ? signature.subarray(0, 64) : signature;

    try {
        return secp256k1.verify(sig, data, keydata, { prehash: false });
```

Either way works — just pick one and make sure the changelog says so.

packages/cryptography/build/bundle-cjs.cjs
```diff
@@ -0,0 +1,36 @@
...
+                        if (BUNDLED_DEPS.test(args.path)) {
+                            return null;
+                        }
+                        return { path: args.path, external: true };
+                    });
+                },
+            },
+        ],
+    })
+    .catch(() => process.exit(1));
```
ddeskov-limechain: Nit: the error object is dropped. `logLevel: "info"` covers normal build errors, but a plugin or config error would exit 1 with no message.

```suggestion
    .catch((err) => {
        console.error(err);
        process.exit(1);
    });
```

.github/dependabot.yml
Fantomasa: Good catch — the flag was only on the smoke-test line; the mocha suite ran without it. Fixed in 9aade7e6: the `common_js_test` `test` task now sets `NODE_OPTIONS=--no-experimental-require-module` at task level, so both the smoke test and the full mocha suite run strict in CI (`common_js.yml` runs `task test` on Node 22/24). Verified the env var reproduces the failure (`require('@noble/curves/secp256k1.js')` → `ERR_REQUIRE_ESM`) and that the mocha suite still passes under it.

packages/cryptography/package.json
Fantomasa: Bumped to `>=20.19.0` in 9aade7e6 to match the floor the three `@noble` v2 deps declare for the ESM entry (we don't test <20.19 anywhere, so `>=18` was an unverified claim). The CJS entry is still built for `node18` since the deps are inlined; I updated the `bundle-cjs.cjs` comment to say so. One thing worth a maintainer call: this makes cryptography's floor stricter than the root SDK's `>=18`, whose ESM path pulls the same noble runtime — happy to align the root SDK in a follow-up if you'd prefer.

packages/cryptography/src/primitive/ecdsa.js
Fantomasa: Went with preserving v1 behavior to avoid the public-API break: in 9aade7e6 `verify()` takes the 64-byte `r||s` prefix of a 65-byte recoverable signature, so callers passing `r||s||v` keep working. Updated the regression test to assert 65-byte → true and other non-64 lengths → false, and added a `# Unreleased` CHANGELOG entry for the re-applied migration (noting verify() behavior is unchanged from v1).

packages/cryptography/build/bundle-cjs.cjs
Fantomasa: Fixed in 9aade7e6 — the catch now `console.error(error)` before `process.exit(1)` so a plugin/config failure isn't silent. (Used `error` rather than `err` to match the surrounding code style.)

ddeskov-limechain (approved):
LGTM

andrewb1269 (approved):
Approve the dependabot update

Issue #4305, opened 2026-08-04
Replace the ping() / pingAll()


mustafauzunn:
Replace the ping() / pingAll() probe (before August 13, 2026) Replace the AccountBalanceQuery liveness probe inside Client.ping() and Client.pingAll() with a COST_ANSWER query. This is a purely internal change with no public API impact. It must ship before the release 77 testnet rollout to avoid breaking Solo readiness checks and connectivity tests.

All SDKs must replace the AccountBalanceQuery probe inside ping() and pingAll() with a CryptoService/getAccountInfo request for account 0.0.2 with ResponseType = COST_ANSWER. Setting COST_ANSWER instructs the node to return the query fee without executing it — no HBAR is charged. A healthy node returns a cost response, which the probe treats as success. A node that is degraded or unreachable fails at the gRPC layer, which the probe treats as failure.

More info: https://github.com/hiero-ledger/sdk-collaboration-hub/blob/main/proposals/account-balance-query-deprecation.md

Issue #4286, opened 2026-07-24
Implement MirrorNodeAccountBalanceQuery as mirror node replacement for AccountBalanceQuery


ddeskov-limechain:
# Implement `MirrorNodeAccountBalanceQuery` (mirror node replacement for `AccountBalanceQuery`)

## Background

`AccountBalanceQuery` is being deprecated (hard error on execute — see companion issue #4285) because the consensus node removes `CryptoService/cryptoGetBalance` in release 0.77. This issue adds the replacement query per the collaboration hub proposal **as merged**:

- [AccountBalanceQuery Mirror Node Migration](https://github.com/hiero-ledger/sdk-collaboration-hub/blob/main/proposals/account-balance-query-mirror-node-migration.md)

> **Note:** the proposal was reworked by [sdk-collaboration-hub#280](https://github.com/hiero-ledger/sdk-collaboration-hub/pull/280) (merged) after mirror node team review. Earlier revisions of this issue described the pre-#280 design (accounts endpoint, token balances, `setContractId`, 404 semantics) — that design is obsolete; see this issue's edit history if needed.

## Public API

New standalone class following the existing mirror node REST query pattern (`MirrorNodeContractCallQuery`, `FeeEstimateQuery`). Does **not** extend the base `Query` class (no consensus-node gRPC machinery, no payment logic — the query is free).

```javascript
import { MirrorNodeAccountBalanceQuery } from "@hiero-ledger/sdk";

const balance = await new MirrorNodeAccountBalanceQuery()
    .setAccountId("0.0.12345") // or EVM address / alias / contract ID
    .execute(client);

console.log(balance.hbars.toString());
```

- New read-only **`MirrorNodeAccountBalance`** return type carrying **`hbars` only**. Token balances and decimals are deliberately out of scope: the balances endpoint does not paginate them, and unbounded pagination against `/accounts/{id}/tokens` is a DDoS risk for the mirror node.
- **No `setContractId`** — the balances endpoint resolves contract IDs through `setAccountId`.
- `setAccountId` accepts anything the mirror node resolves: `shard.realm.num`, EVM address (`0x…`), public key alias, or contract ID.

## Implementation

1. `GET {client.mirrorRestApiBaseUrl}/balances?account.id={id}` — parse `balances[0].balance` (tinybars) → `MirrorNodeAccountBalance.hbars`.
[...]

PR #4306, opened 2026-08-04
chore: add AccountBalanceQuery deprecation message

### LINKED ISSUE #4285: Deprecate AccountBalanceQuery
# Deprecate `AccountBalanceQuery`

## Background

The consensus node is removing the `CryptoService/cryptoGetBalance` endpoint in release **0.77** (decision final). The query has **already been removed from the previewnet, testnet, and mainnet throttle configurations**, and [hiero-consensus-node#26457](https://github.com/hiero-ledger/hiero-consensus-node/issues/26457) removes it from the genesis throttle config, test-clients/yahcli, and marks it deprecated in the protobufs.

This issue implements the SDK side per the collaboration hub proposal:
- [AccountBalanceQuery Deprecation](https://github.com/hiero-ledger/sdk-collaboration-hub/blob/main/proposals/account-balance-query-deprecation.md)
- Companion: [AccountBalanceQuery Mirror Node Migration](https://github.com/hiero-ledger/sdk-collaboration-hub/blob/main/proposals/account-balance-query-mirror-node-migration.md) (replacement class — tracked in #4286)

> **Note — corrections to the proposals (do not follow the proposal verbatim on these points):**
> 1. The two proposals describe themselves as "mutually exclusive." In practice **both** are being implemented: `AccountBalanceQuery` gets the hard-error behavior from the deprecation proposal, **and** `MirrorNodeAccountBalanceQuery` is added as the replacement (#4286). There is no functional "migration window" — the network throttle configs already reject the query.
> 2. **Update (Aug 2026):** the proposal now ships in **two stages**. Replacing the `Client.ping()` / `pingAll()` probe is **Stage 1**, tracked separately in #4305, and must land **before the release 77 testnet rollout (est. August 13, 2026)** using `getAccountInfo(0.0.2)` with `ResponseType = COST_ANSWER` — the previously discussed gRPC health/connectivity check was evaluated and is **not implemented** on Hedera consensus nodes. This issue is **Stage 2** (deprecation only), targeted **at or before the mainnet rollout (est. September 9, 2026)**; Stage 1 must complete first. The Solo-readiness dependency now hangs on #4305, not this issue.
> 3. The proposal's error message points generically to "the mirror node REST API." Since `MirrorNodeAccountBalanceQuery` is being implemented, the message should point to it (text below).

### 1. Deprecate `AccountBalanceQuery`

- Add JSDoc `@deprecated` on the class pointing to `MirrorNodeAccountBalanceQuery` (`eslint-plugin-deprecation` is already configured, so call sites get lint diagnostics).
- On construction, emit `console.warn` (same pattern as `Executable.setMaxRetries` / `ManagedNetwork.setNetworkName`):
  ```
  Deprecated: AccountBalanceQuery is no longer supported. Use MirrorNodeAccountBalanceQuery or the mirror node REST API (GET /api/v1/accounts/{id}) to retrieve account balances.
  ```
- Override the internal execution path (precedent: `src/account/AccountAllowanceAdjustTransaction.js` overriding `_execute()`) so `execute()` **rejects immediately without making any network call**, with the same message. No public fields or methods are added or removed.

### 2. Audit remaining internal usages

Grep the repo for all other `AccountBalanceQuery` usages (examples, integration tests, docs, internal helpers) and migrate or remove them — the `Client.ping()` / `pingAll()` path itself is migrated in #4305 (Stage 1). The XTS dry run on the CN side surfaced hidden dependencies — assume there are more. (The SDK TCK suite is currently the only failing XTS suite on main — coordinate with the TCK fix work.)

## Acceptance criteria

1. Constructing an `AccountBalanceQuery` emits the `console.warn` deprecation warning.
2. `execute()` rejects with the deprecation error **without any network call to a consensus node**.
3. No remaining internal code path constructs or executes `AccountBalanceQuery` (examples/docs updated; ping path handled in #4305).
4. Unit tests for 1–2; the no-network-call check (2) is the critical one.

## References

- CN issue: https://github.com/hiero-ledger/hiero-consensus-node/issues/26457
- Stage 1 (ping replacement): #4305
- TCK coverage is tracked separately in `hiero-ledger/hiero-sdk-tck` per the proposal's test plan.

Jexsie:
## **Description**
Deprecates the `AccountBalanceQuery` class by ensuring it cannot be used to make network calls and providing clear messaging and tests around its deprecation. Users are now directed to use `MirrorNodeAccountBalanceQuery` or the mirror node REST API instead. The most important changes are:

* Added a clear deprecation message (`ACCOUNT_BALANCE_QUERY_DEPRECATION_MESSAGE`) explaining that `AccountBalanceQuery` is no longer supported and guiding users to alternatives. This message is logged as a warning whenever the class is constructed. 
* Updated the `execute` and `_execute` methods of `AccountBalanceQuery` to immediately reject with an error containing the deprecation message, preventing any network calls from being made.

 **Testing**

* Added comprehensive unit tests in `test/unit/AccountBalanceQuery.js` to verify that:
  - A warning is logged on construction and deserialization.
  - All execution paths (`execute`, `_execute`, and `getCost`) reject with the deprecation error without making any network calls.

**Related issue(s)**:

Fixes https://github.com/hiero-ledger/hiero-sdk-js/issues/4285

**Notes for reviewer**:

**Checklist**


Jexsie:
CI failures are related to [here](https://github.com/hiero-ledger/hiero-sdk-js/pull/4300#issuecomment-5175707698)

ddeskov-limechain:
Hi @Jexsie — thanks for the PR, and apologies for a scope change that landed mid-flight.

The collaboration hub proposal was updated last week ([`241c5eb`](https://github.com/hiero-ledger/sdk-collaboration-hub/commit/241c5ebde82f14d7176b1b62e76d1e4cd6d35ab8)) and splits this work into two stages, which changes this PR in two ways:

1. **The ping/pingAll replacement moved to its own issue, #4305** (Stage 1, assigned to @Fantomasa), so please drop the channel-level ping half of this PR (`Channel`/`NodeChannel`/`WebChannel`/`NativeChannel` ping methods, the `Client.ping` rewrite, `getNodesByKey`, and the related tests). #4285 has been rescoped to deprecation only.
2. **The mechanism changed too**: the updated proposal evaluated the gRPC connectivity/health-check approach and rejected it (not implemented on Hedera consensus nodes). The probe must instead use `CryptoService/getAccountInfo` for account `0.0.2` with `ResponseType = COST_ANSWER` — so the channel-ping code unfortunately can't be salvaged for #4305 either.

The deprecation half (warn on construction, reject on `execute()`, no network call) matches the updated proposal — please keep all of that. Sorry for the churn, and thanks!

ddeskov-limechain:
Thanks for the PR @Jexsie. The class-level change itself is in good shape — the `@deprecated` JSDoc, the `console.warn` on construction, the hard reject in `execute()`/`_execute()` with the exact message from #4285, and the unit tests all match the spec (and `getCost` is correctly covered via `CostQuery`'s delegation to `_execute`).

However, the PR is incomplete: it implements only **Scope 1** of #4285 and skips **Scope 2 / acceptance criterion 3 — "No remaining internal code path constructs or executes `AccountBalanceQuery`"**. Because this PR turns the class into a hard error, every remaining internal caller now breaks at runtime, which is why CI on this PR is red (integration tests, the unit test job, and the examples job are all failing — e.g. `AccountBalanceIntegrationTest.js` 5/5 failed, `ClientConstantsIntegrationTest.js` 53/53 failed). Per the collaboration hub proposal ([account-balance-query-deprecation.md](https://github.com/hiero-ledger/sdk-collaboration-hub/blob/main/proposals/account-balance-query-deprecation.md)): *"SDK teams should audit all internal uses of AccountBalanceQuery — not only the public ping() path."*

The central problem the audit has to solve is **token balances**: `AccountBalance` carries `hbars` + `tokens` + `tokenDecimals`, but `MirrorNodeAccountBalanceQuery` is deliberately hbars-only, and `AccountInfoQuery` does not carry token balances. So every internal caller that needs tokens must get them from the mirror node directly.

What still needs to be done in this PR:

1. **New `MirrorNodeTokenBalanceQuery`** wrapping `GET /api/v1/accounts/{id}/tokens?token.id={tokenId}` — `setAccountId` + **required** `setTokenId`, always exactly one request, returning that token's balance **and decimals**. Per steven-sheehy's review on the hub proposal, the SDK deliberately offers no "fetch all token balances" method (unbounded and needless load on the mirror node when callers want one token); callers who truly need a full portfolio should page the mirror REST API themselves. NFT entries report the serial count, matching old `AccountBalance.tokens` semantics. Reuse the retry/deadline/`mirrorRestApiBaseUrl` mechanics from `MirrorNodeAccountBalanceQuery` (`_fetchJson`), and keep it browser-compatible (needed by `LocalProviderWeb`).

2. **`src/Wallet.js`, `src/LocalProvider.js`, `src/LocalProviderWeb.js`** — `getAccountBalance()` must keep working with its existing `Promise<AccountBalance>` contract: hbar via `MirrorNodeAccountBalanceQuery`; the `tokens`/`tokenDecimals` maps remain **empty by design**, with the JSDoc and CHANGELOG stating this explicitly and pointing to `MirrorNodeTokenBalanceQuery` for per-token balances. Do **not** deprecate these methods, and do not leave them constructing `AccountBalanceQuery`.

3. **Integration tests** (~15 files):
   - Tests *of* `AccountBalanceQuery` (`AccountBalanceIntegrationTest.js`, the balance cases in `ClientConstantsIntegrationTest.js`, …): remove or rewrite to assert the new behavior (warn on construction, reject on execute); the unit tests already cover the core.
   - hbar-only balance assertions in other tests: migrate to `AccountInfoQuery` (consensus node, immediately consistent).
   - **Token-balance assertions** (`balance.tokens.get(...)` in `TokenAirdrop`, `TokenReject`, `TokenRejectFlow`, `TokenClaimAirdrop`, `TokenCancelAirdrop`, `TokenAssociate`, `TokenDissociate`, `TokenBurn`, `TopicCreate` integration tests): migrate to the token query from (1) with its `setTokenId` filter. The mirror node lags consensus by a few seconds, so assert via a small poll/retry helper (add one under `test/integration/utils/` — none exists today) rather than a single read, or these suites will be flaky.

4. **Examples** (~21 files) — migrate to `MirrorNodeAccountBalanceQuery` (already on `main` via #4286); examples are exactly where users should see the replacement. Where an example transfers hbar and immediately reads the balance, account for mirror-node lag (short poll/retry), or the examples CI job will stay flaky. **One amendment to this point:** the examples that display *token* balances (`token/airdrop.js`, `token/reject.js`, `token/hip-542-auto-account-create.js`, `token/exempt-custom-fees.js`, `consensus/create-with-revenue.js`, …) cannot use the hbars-only `MirrorNodeAccountBalanceQuery` — use the token query from (1) for those.

5. **`tck/methods/account.ts`** — the TCK server's `getAccountBalance` JSON-RPC method wraps `AccountBalanceQuery` (including token balances in its response). Rework it on the queries from (1)+(2), and please coordinate with the ongoing TCK/XTS work (per #4285, the SDK TCK suite is currently the only failing XTS suite).

6. Minor: in the new `getCost` unit test, also assert the rejection message contains the deprecation text — right now it would pass even if the rejection came from an unrelated error.

A repo-wide `grep -rn "AccountBalanceQuery" src/ test/ examples/ tck/` should come back clean (class definition, exports, and the deprecation unit tests only), and all CI jobs need to be green.

**On timing:** we need this in the upcoming release. Consensus node release 0.77 already rejects `CryptoService/cryptoGetBalance` on testnet (since Aug 13), and the mainnet rollout is estimated for **September 9, 2026**, so this must land ahead of that release cut. If the remaining scope can't be completed within the next few days, we'll unfortunately have to close this PR and implement the change ourselves to make the release train — please let us know either way.

Jexsie:
> Examples (~21 files) — migrate to MirrorNodeAccountBalanceQuery (already on main via https://github.com/hiero-ledger/hiero-sdk-js/issues/4286); examples are exactly where users should see the replacement. Where an example transfers hbar and immediately reads the balance, account for mirror-node lag (short poll/retry), or the examples CI job will stay flaky.

I was waiting for this. Thank you, will push updates

test/integration/ClientConstantsIntegrationTest.js
Jexsie, line 1: Removed `ClientConstantsIntegrationTest.js` and replaced it with a unit test.

**Why it couldn't just be migrated**

The old test wasn't actually testing anything. For each constant, it set the node account ID on a free AccountBalanceQuery and ran it against env.client — the local network, which only has node 0.0.3.

Because the query was free, no payment was needed, so Executable.transactionNodeIds stayed empty. That meant the "node is not in the client's network" guard never fired, and node selection quietly fell back to picking a random node. In practice, 0.0.3 answered all 53 assertions, no matter which node ID the test asked for.

Switching to a paid AccountInfoQuery does make the guard fire — and then only the 0.0.3 cases pass, which confirms the rest were never real.

**What replaces it**

test/unit/ClientConstants.js checks the things the old test could genuinely have verified, with no network involved:

- every address is host:port, with a numeric port
- every value is an AccountId with num > 0
- ALL_WEB_NETWORK_NODES covers all three web maps
- ALL_NETWORK_IPS keys keep their trailing colon — that's the lookup convention in NodeChannel, and a missing colon would silently never match

src/LocalProvider.js
```diff
@@ -86,13 +86,20 @@ export default class LocalProvider {
...
      */
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
     getAccountBalance(accountId) {
-        return new AccountBalanceQuery()
-            .setAccountId(accountId)
-            .execute(this._client);
+        return Promise.reject(
+            new Error(ACCOUNT_BALANCE_QUERY_DEPRECATION_MESSAGE),
+        );
     }
```
ddeskov-limechain, line 108: Point 2 of my review comment: keep this working — hbar from the mirror node, empty token maps, no `@deprecated`. Needs real imports for `MirrorNodeAccountBalanceQuery` and `AccountBalance` (a typedef today); the message import can go.

```suggestion
    /**
     * The account's hbar balance, read from the mirror node.
     *
     * The `tokens` and `tokenDecimals` maps are always empty: the consensus
     * node no longer serves balances, and the mirror node balance endpoint
     * is hbar-only. Use `MirrorNodeTokenBalanceQuery` for a token balance.
     *
     * @param {AccountId | string} accountId
     * @returns {Promise<AccountBalance>}
     */
    async getAccountBalance(accountId) {
        const { hbars } = await new MirrorNodeAccountBalanceQuery()
            .setAccountId(accountId)
            .execute(this._client);

        return new AccountBalance({
            hbars,
            tokens: null,
            tokenDecimals: null,
        });
    }
```

src/LocalProviderWeb.js
```diff
@@ -81,13 +81,20 @@ export default class LocalProviderWeb {
...
      */
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
     getAccountBalance(accountId) {
-        return new AccountBalanceQuery()
-            .setAccountId(accountId)
-            .execute(this._client);
+        return Promise.reject(
+            new Error(ACCOUNT_BALANCE_QUERY_DEPRECATION_MESSAGE),
+        );
     }
```
ddeskov-limechain, line 102: Same as `LocalProvider.js` — keep it working:

```suggestion
    /**
     * The account's hbar balance, read from the mirror node.
     *
     * The `tokens` and `tokenDecimals` maps are always empty: the consensus
     * node no longer serves balances, and the mirror node balance endpoint
     * is hbar-only. Use `MirrorNodeTokenBalanceQuery` for a token balance.
     *
     * @param {AccountId | string} accountId
     * @returns {Promise<AccountBalance>}
     */
    async getAccountBalance(accountId) {
        const { hbars } = await new MirrorNodeAccountBalanceQuery()
            .setAccountId(accountId)
            .execute(this._client);

        return new AccountBalance({
            hbars,
            tokens: null,
            tokenDecimals: null,
        });
    }
```

ddeskov-limechain (commented):
Thanks for the update. The deprecation core, the unit tests, and the ClientConstants cleanup all look good. Using `AccountInfoQuery` for the integration tests and the TCK instead of mirror-node reads is fine by me — it is immediately consistent, so no polling is needed there.

Two points from my earlier comment are still open, both blocking:

1. **`MirrorNodeTokenBalanceQuery` is still missing.** After this PR, SDK users have no public way to read a token balance: `MirrorNodeAccountBalanceQuery` is hbar-only, and `AccountInfoQuery.tokenRelationships` (what the examples now show) is itself deprecated since HIP-367 and capped at 1000 entries. We need the new query — required `setTokenId`, one request to `GET /api/v1/accounts/{id}/tokens?token.id=X`, returning balance + decimals — and the token examples should use it.
2. **`Wallet` / `LocalProvider` / `LocalProviderWeb` `getAccountBalance()` must keep working**, not reject. Hbar from the mirror node, empty token maps, documented. Rejecting breaks every signer/provider consumer at runtime. Suggestions inline; `test/unit/node/AccountBalanceDeprecation.js` then flips to assert the new behavior.

Smaller points inline (fixed 5s sleeps in the examples, the get-balance example). Please also add a CHANGELOG entry for the behavior change.

Timing is unchanged — this needs to make the release before the mainnet 0.77 rollout (~Sept 9).

src/Wallet.js
```diff
@@ -162,11 +162,17 @@ export default class Wallet {
...
+     *
      * @returns {Promise<AccountBalance>}
      */
     getAccountBalance() {
-        return this.call(
-            new AccountBalanceQuery().setAccountId(this.accountId),
+        return Promise.reject(
+            new Error(ACCOUNT_BALANCE_QUERY_DEPRECATION_MESSAGE),
         );
     }
```
ddeskov-limechain, line 176: Keep this one working too — it can delegate to the provider:

```suggestion
    /**
     * The wallet account's hbar balance, read from the mirror node.
     *
     * The `tokens` and `tokenDecimals` maps are always empty — use
     * `MirrorNodeTokenBalanceQuery` for a token balance.
     *
     * @returns {Promise<AccountBalance>}
     */
    getAccountBalance() {
        if (this.provider == null) {
            throw new Error("wallet does not contain a provider");
        }

        return this.provider.getAccountBalance(this.accountId);
    }
```

examples/account/multi-app-transfer.js
```diff
@@ -76,13 +76,18 @@ async function main() {
         ).getReceipt(client)
     ).accountId;
 
+    // Both accounts were only just created. The mirror node ingests consensus
+    // state asynchronously, so wait for it to see them — otherwise it reports a
+    // zero balance for an account that has already been funded.
+    await new Promise((resolve) => setTimeout(resolve, 5000));
```
ddeskov-limechain: Fixed sleeps are still flaky on slow runners and waste time on fast ones. Please poll instead, like `untilMirror` in `MirrorNodeAccountBalanceQueryIntegrationTest.js`. Same for the other examples with 5s sleeps.

examples/token/airdrop.js
```diff
@@ -196,29 +196,33 @@ async function main() {
...
 
-    let account3Balance = await new AccountBalanceQuery()
+    let account3Balance = await new AccountInfoQuery()
         .setAccountId(accountId3)
         .execute(client);
 
     console.log(
         "Account1 balance after airdrop: ",
-        account1Balance.tokens.get(tokenId).toInt(),
+        account1Balance.tokenRelationships
```
ddeskov-limechain: `tokenRelationships` is deprecated (HIP-367) and capped at 1000 entries, so the examples would teach users a field that is on its way out. Once `MirrorNodeTokenBalanceQuery` exists, the token examples should use it.

examples/account/get-balance.js
```diff
@@ -27,14 +28,14 @@ async function main() {
     );
 
     try {
-        const balance = await new AccountBalanceQuery()
+        const info = await new AccountInfoQuery()
```
ddeskov-limechain: Once `Wallet.getAccountBalance()` works again, this example can just call it — no paid query needed for a plain hbar read.

test/integration/ClientConstantsIntegrationTest.js
ddeskov-limechain: Good catch — agreed. The old test never really pinned the node, so it was not testing what its name said. The new unit test covers what is actually verifiable.

src/query/MirrorNodeTokenBalanceQuery.js
```diff
@@ -0,0 +1,283 @@
...
+     * throws immediately. `deadline` is the epoch-millisecond cutoff shared by
+     * every attempt of the whole operation.
+     *
+     * @private
+     * @param {string} url
+     * @param {Client} client
+     * @param {?number} deadline
+     * @returns {Promise<unknown>}
+     */
+    async _fetchJson(url, client, deadline) {
```
ddeskov-limechain, line 206: Non-blocking nit: `_fetchJson` and `sleep` are now duplicated between this class and `MirrorNodeAccountBalanceQuery` (~100 lines of identical retry logic). They could live next to `isRetryableNetworkError` in `network/mirrorRestRetry.js` and be shared. Fine as a follow-up.

ddeskov-limechain:
All points addressed — thanks! `MirrorNodeTokenBalanceQuery` matches what we agreed (required `setTokenId`, one request, balance + decimals), the provider methods keep working with empty token maps, the polling helpers replace the fixed sleeps, and the CHANGELOG covers the behavior change well. I resolved the earlier threads; one non-blocking nit left inline.

On the red CI — both failures look unrelated to this PR:

- `Test using Node 22` failed on a single chromium test (`TransactionIntegrationTest` HIP-1300), a file this PR does not touch. The caught error had no `status` (transport-level), and the same job passed on the previous commit, which differs only by formatting.
- The examples job failed on 20 examples with one identical signature: consensus-node `timeout exceeded` on `0.0.3` early in `main()`. It hit untouched examples (`signing/*`, `token/change-keys`, `schedule/identical-transaction`, ...) as much as migrated ones, and there are zero mirror/polling errors in the whole log — the local node was unhealthy on that run.

I'll rerun the failed jobs; with a green run this is good to go from my side.

Jexsie:
> All points addressed — thanks! `MirrorNodeTokenBalanceQuery` matches what we agreed (required `setTokenId`, one request, balance + decimals), the provider methods keep working with empty token maps, the polling helpers replace the fixed sleeps, and the CHANGELOG covers the behavior change well. I resolved the earlier threads; one non-blocking nit left inline.
> 
> On the red CI — both failures look unrelated to this PR:
> 
> * `Test using Node 22` failed on a single chromium test (`TransactionIntegrationTest` HIP-1300), a file this PR does not touch. The caught error had no `status` (transport-level), and the same job passed on the previous commit, which differs only by formatting.
> * The examples job failed on 20 examples with one identical signature: consensus-node `timeout exceeded` on `0.0.3` early in `main()`. It hit untouched examples (`signing/*`, `token/change-keys`, `schedule/identical-transaction`, ...) as much as migrated ones, and there are zero mirror/polling errors in the whole log — the local node was unhealthy on that run.
> 
> I'll rerun the failed jobs; with a green run this is good to go from my side.

I had everything green locally. With an exception of `consensus/pub-sub.js` example. Though I get no error of any kind.

ddeskov-limechain:
Closing the loop on the examples job: it is broken repo-wide right now, not by this PR. Today's run on `feat/tck-ping-methods` — a branch with no example changes — failed with the identical failure list and ~2-minute cadence (`ethereum-transaction-example`, `serialize-deserialize`, `time-drift`, `contract/*`, `file/create`, ...), all with the same consensus `timeout exceeded` on `0.0.3`: https://github.com/hiero-ledger/hiero-sdk-js/actions/runs/32134478170. Main's examples job has also been red since Aug 14. A green examples run is not achievable on any branch today, so I would not hold this PR on it — the CI local node needs looking at separately.

ddeskov-limechain (changes requested):
Thanks for the update. Three things before this can go in, then it is good from my side:

1. **Unknown account behavior changed on main.** #4335 (merged Sept 1) makes `MirrorNodeAccountBalanceQuery` throw `MirrorNodeStatusError` with `Status.InvalidAccountId` instead of returning zero. `test/unit/node/AccountBalanceDeprecation.js` still expects zero for account 0.0.10, which is the failing unit test. Please flip that test to assert the error, and mention in the CHANGELOG "Changed" entry that `Wallet` / `LocalProvider.getAccountBalance()` now throw `MirrorNodeStatusError` for an unknown account rather than `PrecheckStatusError`.
2. **CHANGELOG header.** Put the new sections under `# Unreleased` above `# v2.87.0`; the release PR renames it.
3. **`MirrorNodeTokenBalanceQuery` scope note.** The cross-SDK proposal for this query (hiero-ledger/sdk-collaboration-hub#281) is still under review and its current draft returns a page with a cursor and an optional token filter. We are shipping the single-token form now to make the release 77 cutoff. Please add a sentence to the class JSDoc and the CHANGELOG "Added" entry that the API may change once that proposal is finalized.

The examples job failures are the same infra signature as last time (timeouts on untouched examples), not this PR.

ddeskov-limechain (dismissed):
LGTM

ddeskov-limechain:
@Jexsie please fix the CI so we can merge your PR

Jexsie:
Updated. Thank you @ddeskov-limechain

Fantomasa (approved):
LGTM

PR #4313, opened 2026-08-06
fix: ship CommonJS type declarations for node16 module resolution

### LINKED ISSUE #2722: Type checking errors when requiring `@hashgraph/sdk`
### Description

When loading the SDK with `require('@hashgraph/sdk')`, I get a bunch of errors like

```
node_modules/@hashgraph/sdk/lib/Transfer.d.ts:75:20 - error TS1340: Module 'long' does not refer to a type, but is used as a type here. Did you mean 'typeof import('long')'?

75 export type Long = import("long");
                      ~~~~~~~~~~~~~~

node_modules/long/umd/index.d.ts:1:18 - error TS1479: The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. Consider writing a dynamic 'import("../index.js")' call instead.

1 import Long from "../index.js";
```

and the error

```
scripts/create-token.js:21:9 - error TS1479: The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. Consider writing a dynamic 'import("@hashgraph/sdk")' call instead.
  To convert this file to an ECMAScript module, change its file extension to '.mjs', or add the field `"type": "module"` to 'package.json'.

21 require('@hashgraph/sdk');
           ~~~~~~~~~~~~~~~~
```

### Steps to reproduce

In a new `npm` JavaScript project, add the SDK `2.55.1` and `typescript`. The `package.json`'s `type` is left untouched (it's using `commonjs` as default.

Use the following `tsconfig.json`

```json
{
    "compilerOptions": {
        "allowJs": true,
        "checkJs": true,
        "module": "Node16",
        "moduleResolution": "Node16",
        "resolveJsonModule": true,
        "noEmit": true,
        "strict": true,
        "exactOptionalPropertyTypes": true,
        "noFallthroughCasesInSwitch": true,
        "noImplicitAny": true,
        "noImplicitOverride": true,
        "noImplicitReturns": true,
        "noImplicitThis": true,
        "noPropertyAccessFromIndexSignature": true,
        "noUnusedLocals": true,
        "noUnusedParameters": true,
    },
}
```

Then create a minimal script that requires `@hashgraph/sdk`, for example

```javascript
require('@hashgraph/sdk');
```

The errors mentioned above appear when type-checking

```console
npx tsc 
```

### Additional context

If, however, I set `skipLibCheck: true` in `tsconfig.json`, I get only the following error mentioned above (errors `Module 'long' does not refer to a type` disappear)

```console
$ npx tsc
scripts.js:21:9 - error TS1479: The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. Consider writing a dynamic 'import("@hashgraph/sdk")' call instead.
  To convert this file to an ECMAScript module, change its file extension to '.mjs', or add the field `"type": "module"` to 'package.json'.

21 require('@hashgraph/sdk');
           ~~~~~~~~~~~~~~~~
```

The last version of the SDK is installed

```console
$ npm why @hashgraph/sdk
@hashgraph/sdk@2.55.1 dev
node_modules/@hashgraph/sdk
  dev @hashgraph/sdk@"^2.55.1" from the root project
```

### Hedera network

other

### Version

2.55.1

### Operating system

macOS

Fantomasa:
**Description**:

CommonJS projects type-checking with `moduleResolution: node16`/`nodenext` could not `require("@hiero-ledger/sdk")`. Three classes of errors (all reproduced with the issue's exact tsconfig):

- **TS1479** on the consumer's `require(...)` line — the packages ship only `.d.ts` declarations inside a `"type": "module"` package, so tsc treats the typings as ESM-only even though a real CJS build (`lib/*.cjs`) exists.
- **TS1340** (`Module 'long' does not refer to a type`, ~100 occurrences incl. `bignumber.js`) — the generated declarations referenced `import("long")` in a form that only resolves against the CJS (`export =`) typings, while node16 ESM consumers resolve the ESM typings.
- **TS2835** — extensionless relative imports in JSDoc type references.

**Changes**:

- Mirror every `lib/**/*.d.ts` to a CJS-flavored `.d.cts` at build time (new `scripts/generate-cts-declarations.js`): relative specifiers are rewritten `.js` → `.cjs` (matching what babel does to the runtime build) and `long`/`bignumber.js` type references are rewritten to their `export =` form.
- Declare per-condition `types` in the `exports` map of both `@hiero-ledger/sdk` and `@hiero-ledger/cryptography`: `import` resolves the ESM `.d.ts`, `require` resolves the new `.d.cts`.
- Normalize JSDoc to `import("long").default` / `import("bignumber.js").default` (the ESM-typings form) and add explicit `.js` extensions to relative JSDoc type imports.
- Switch the build tsconfig to `moduleResolution: "bundler"` so `npx tsc` validates the same ESM typings that node16/bundler consumers resolve. Files that silently relied on the ambient `Long` global (only provided by long's CJS typings) got explicit typedefs, and one unsound `instanceof Long` narrowing in `TopicMessage` was replaced with a `typeof === "number"` check.
- Stop redeclaring `Error.stack` in cryptography's `BadKeyError` (broke consumers with `exactOptionalPropertyTypes`).
- `task build` now cleans `lib/` first — stale declarations from deleted sources (e.g. `EthereumFlow.d.ts`) were being packed into releases.

**Related issue(s)**:

Fixes #2722

**Notes for reviewer**:

- Verified with the issue's exact repro (CJS project, `module: Node16`, `exactOptionalPropertyTypes`, full lib check — no `skipLibCheck`): `npx tsc` exits 0 on TS 5.9.3 and TS 7.0.2 with current `@types/node`, and `require("@hiero-ledger/sdk")` works at runtime. ESM consumers (`Node16`/`NodeNext`) keep type-checking and importing cleanly.
- `npx @arethetypeswrong/cli --pack` is green across all resolution modes (node10, node16-CJS, node16-ESM, bundler) for both packages.
- `@hiero-ledger/cryptography` needs to be published together with the next SDK release for the fix to be complete (its `exports`/`.d.cts` changes are required by the SDK's `.d.cts`).

## Checklist

- [x] Tested (issue repro matrix above; SDK unit tests 1942 passed, cryptography 170 passed, strict-CJS suite passes; `eslint` and `prettier` clean)


ddeskov-limechain (dismissed):
Three non-blocking comments worth leaving:
1. node10 consumers get ~73 new errors. moduleResolution: "node" ignores exports, so it lands on the root types: ./lib/index.d.ts — the ESM flavor with import("long").default — while resolving long to its export = Long typings. I reproduced it: TS2694: Namespace 'Long' has no exported member 'default'. Pre-existing for bignumber.js (36 sites already had .default on main), so this widens an existing break to ~109 rather than creating one, and it only bites with skipLibCheck: false. Note the PR body's attw --pack evidence doesn't cover this — attw checks resolution, not member access.
2. packages/cryptography build doesn't rm -rf lib. The PR added that to the root build precisely to stop stale declarations from being packed — cryptography has the identical problem, now compounded by stale .d.cts mirrors too.
3. Nothing guards the fix. common_js_test is runtime mocha on plain JS; no CI job type-checks a CJS consumer, so [#2722](https://github.com/hiero-ledger/hiero-sdk-js/issues/2722) can silently reopen. And switching the build tsconfig to moduleResolution: "bundler" picks the one mode that permits extensionless relative imports — the TS2835 class just fixed can creep straight back. nodenext would catch both, if it emits cleanly.

ddeskov-limechain (approved):
LGTM

Issue #3953, opened 2026-04-23
epic: improve unit test coverage across the SDK


venilinvasilev:
## Overview

The SDK currently has significant coverage gaps across its source files. This Epic tracks all individual issues created to improve test coverage file-by-file.

## Context

Coverage was analyzed across all four test suites (`node`, `browser`, `node-integration`, `browser-integration`). A line is considered uncovered only when missed in **every** suite.

**Baseline stats (after excluding fully deprecated classes via #3952):**

| Band | Files |
|------|-------|
| 0% coverage | ~76 |
| 1–49% | ~122 |
| 50–79% | ~69 |
| 80–99% | 15 |
| 100% | 28 |

### Prerequisites
- [ ] #3952 — Exclude fully deprecated classes from vitest coverage configs

## Scope

Issues will be created for files in the following priority order:
1. **Tier 1** — Files at 0% coverage (no tests at all)
2. **Tier 2** — Files at 1–49% coverage with ≤100 lines (small, focused additions)
3. **Tier 3** — Files at 1–49% coverage with >100 lines (larger effort)

Files with only deprecated *methods* (but an active class) are included. Fully deprecated classes are excluded via #3952.

PR #4316, opened 2026-08-09
test: add unit tests for TokenRelationship

### LINKED ISSUE #4007: [Beginner]: Add unit tests for TokenRelationship
### 🐥 Beginner Friendly

This issue is a great fit for contributors who are ready to explore the Hiero JavaScript
codebase a little more and take on slightly more independent work. We strongly recommend
completing a Good First Issue before tackling a Beginner Issue.

Beginner Issues often involve reading existing JavaScript code, understanding how
different parts of the SDK fit together, and making small, thoughtful updates
that follow established patterns.

The goal is to support skill growth while keeping the experience approachable,
well-scoped, and enjoyable.

### 👾 Description of the Issue

`src/account/TokenRelationship.js` has **0% test coverage**. It is a value class that holds a token's relationship to an account and contains non-trivial conditional logic in its `_fromProtobuf` / `_toProtobuf` round-trip that is currently untested.

The interesting logic lives in `_fromProtobuf` ([`src/account/TokenRelationship.js`](../blob/main/src/account/TokenRelationship.js)):

```js
const isKycGranted =
    relationship.kycStatus == null || relationship.kycStatus === 0
        ? null
        : relationship.kycStatus === 1;

const isFrozen =
    relationship.freezeStatus == null || relationship.freezeStatus === 0
        ? null
        : relationship.freezeStatus === 1;
```

Both `kycStatus` and `freezeStatus` follow the same three-way mapping:

| Proto value | JS field value |
|---|---|
| `null` or `0` (not applicable) | `null` |
| `1` (granted / frozen) | `true` |
| `2` (revoked / unfrozen) | `false` |

The `balance` field also has a three-way branch: `null → Long.ZERO`, an existing `Long` passes through, and a plain number is converted via `Long.fromValue`.

`_toProtobuf` is the inverse: `null → 0`, `true → 1`, `false → 2`.

Similar patterns to reference: [`src/account/HbarAllowance.js`](../blob/main/src/account/HbarAllowance.js) and its test at `test/unit/HbarAllowance.js`.

### 💡 Proposed Solution

1. A new test file `test/unit/TokenRelationship.js` exists
2. `_fromProtobuf` is tested for all three values of `kycStatus` and `freezeStatus` (null/0, 1, 2)
3. `_fromProtobuf` is tested for all three `balance` inputs (null, Long, number)
4. A round-trip test (`_fromProtobuf` → `_toProtobuf`) confirms the inverse mapping is correct
5. Unit tests validate all cases

### 👩‍💻 Implementation Steps

- Browse `test/unit/` for the structure and import conventions used by similar tests (`HbarAllowance.js`, `ProxyStaker.js`)
- Create `test/unit/TokenRelationship.js`
- Import `TokenRelationship` from `../../src/account/TokenRelationship.js`
- Write tests covering:
  - `_fromProtobuf` with `kycStatus` of `null`, `0`, `1`, `2`
  - `_fromProtobuf` with `freezeStatus` of `null`, `0`, `1`, `2`
  - `_fromProtobuf` with `balance` of `null`, a `Long` value, and a plain number
  - `_toProtobuf` inverse: verify `null → 0`, `true → 1`, `false → 2` for both status fields
- Run both test environments and confirm all tests pass:
  - `task test:unit:node -- test/unit/TokenRelationship.js`
  - `task test:unit:browser -- test/unit/TokenRelationship.js`
- Keep the change small and localized to the new test file only

### ✅ Acceptance Criteria

- [ ] Scope: Changes are limited to the new test file
- [ ] API Consistency: Matches established patterns elsewhere in the SDK
- [ ] Tests: Both of the following pass with no failures:
  - `task test:unit:node -- test/unit/TokenRelationship.js`
  - `task test:unit:browser -- test/unit/TokenRelationship.js`
- [ ] Linting: `task lint` passes
- [ ] Behavior: No unrelated behavior changes
- [ ] Review: All code review feedback addressed

### 📋 Step-by-Step Contribution Guide

To help keep contributions consistent and easy to review, remember these key steps:

- [ ] Comment `/assign` to request the issue and get assigned. We only allow one assignment at a time.
- [ ] Fork the repository and create a branch from an updated main
- [ ] Make the requested changes and sign each commit using `-s -S`
- [ ] Push your branch and open a pull request

Here are some best practices to follow:
- [ ] Follow the existing project conventions and examples
- [ ] Avoid breaking public APIs
- [ ] Include tests and/or example updates where appropriate
- [ ] Address all code review feedback

Read [Workflow Guide](manual/training/workflow.md) for step-by-step workflow guidance.
Read [README.md](README.md) for setup instructions.

❗ Pull requests **cannot be merged** without signed commits (use `-s -S`).
See the [Signing Guide](manual/training/signing.md).

### 🤔 Additional Information

If you have questions while working on this issue, feel free to ask!

You can reach the community and maintainers here:
[Hiero-SDK-JavaScript Discord](https://discord.com/channels/905194001349627914/1337424656138899537)

Whether you need help finding the right file, understanding existing code,
or confirming your approach — we're happy to help.

Part of Epic #3953 — improve unit test coverage across the SDK.

sarbojitrana:
Fixes #4007

## Description

`src/account/TokenRelationship.js` had 0% unit test coverage despite the non-trivial three-way conditional logic in its `_fromProtobuf` / `_toProtobuf` round-trip. This PR adds `test/unit/TokenRelationship.js`.

## What is covered

**`_fromProtobuf`**

- `kycStatus` → `isKycGranted`: missing, `null` and `0` map to `null`; `1` maps to `true`; `2` maps to `false`
- `freezeStatus` → `isFrozen`: same three-way mapping
- `balance`: `null` and missing default to `Long.ZERO`; an existing `Long` is passed through by reference (asserted with identity, so the `instanceof` branch is genuinely exercised); a plain number and a string are converted via `Long.fromValue`
- `automaticAssociation`: `true` / `false` pass through, missing and `null` map to `null`
- `tokenId` and `symbol` mapping

**`_toProtobuf`**

- Inverse status mapping: `null → 0`, `true → 1`, `false → 2` for both `kycStatus` and `freezeStatus`
- Full proto shape via `deep.equal`

**Round-trip**

- All nine `kycStatus` × `freezeStatus` combinations survive `_fromProtobuf` → `_toProtobuf` → `_fromProtobuf` unchanged
- A `null` status normalizes to `0` on the way back out

**Constructor**

- Property assignment and `Object.freeze`

The tests follow the conventions in `test/unit/HbarAllowance.js` as suggested in the issue.

## Result

`src/account/TokenRelationship.js` is now at **100% statements, branches, functions and lines**:

```
-------------------|---------|----------|---------|---------|-------------------
File               | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
-------------------|---------|----------|---------|---------|-------------------
All files          |     100 |      100 |     100 |     100 |
 ...elationship.js |     100 |      100 |     100 |     100 |
-------------------|---------|----------|---------|---------|-------------------
```

## Checklist

- [x] Scope: changes are limited to the new test file — no production code touched
- [x] `task test:unit:node -- test/unit/TokenRelationship.js` — 39 passed
- [x] `task test:unit:browser -- test/unit/TokenRelationship.js` — 39 passed
- [x] `npx prettier "test/unit/TokenRelationship.js" --check` passes
- [x] Commit is DCO signed off and GPG signed (`-s -S`)

Note: `task lint` currently reports pre-existing Prettier formatting failures in `src/channel/NativeMirrorChannel.js`, `src/exports.js` and `test/unit/NativeChannel.js` on a clean `main`. Those are unrelated to this change and left untouched.

Part of Epic #3953.

### DIFF
--- a/test/unit/TokenRelationship.js
+++ b/test/unit/TokenRelationship.js
@@ -1,0 +1,377 @@
+import Long from "long";
+import TokenRelationship from "../../src/account/TokenRelationship.js";
+import TokenId from "../../src/token/TokenId.js";
+
+describe("TokenRelationship", function () {
+    const tokenId = new TokenId(0, 0, 100);
+    const symbol = "TEST";
+
+    /**
+     * Builds a protobuf token relationship, overriding only the fields a test
+     * cares about so each case stays focused on a single branch.
+     *
+     * @param {object} [overrides]
+     * @returns {object}
+     */
+    function protoWith(overrides = {}) {
+        return {
+            tokenId: tokenId._toProtobuf(),
+            symbol,
+            balance: Long.fromNumber(10),
+            kycStatus: 0,
+            freezeStatus: 0,
+            automaticAssociation: false,
+            ...overrides,
+        };
+    }
+
+    describe("constructor", function () {
+        it("should store properties correctly", function () {
+            const balance = Long.fromNumber(50);
+
+            const relationship = new TokenRelationship({
+                tokenId,
+                symbol,
+                balance,
+                isKycGranted: true,
+                isFrozen: false,
+                automaticAssociation: true,
+            });
+
+            expect(relationship.tokenId.toString()).to.equal(
+                tokenId.toString(),
+            );
+            expect(relationship.symbol).to.equal(symbol);
+            expect(relationship.balance.toNumber()).to.equal(50);
+            expect(relationship.isKycGranted).to.be.true;
+            expect(relationship.isFrozen).to.be.false;
+            expect(relationship.automaticAssociation).to.be.true;
+        });
+
+        it("should freeze the object", function () {
+            const relationship = new TokenRelationship({
+                tokenId,
+                symbol,
+                balance: Long.ZERO,
+                isKycGranted: null,
+                isFrozen: null,
+                automaticAssociation: null,
+            });
+
+            expect(Object.isFrozen(relationship)).to.be.true;
+        });
+    });
+
+    describe("_fromProtobuf", function () {
+        it("should map the token id and symbol", function () {
+            const relationship = TokenRelationship._fromProtobuf(protoWith());
+
+            expect(relationship.tokenId.toString()).to.equal("0.0.100");
+            expect(relationship.symbol).to.equal(symbol);
+        });
+
+        describe("kycStatus", function () {
+            it("should map a missing kycStatus to null", function () {
+                const proto = protoWith();
+                delete proto.kycStatus;
+
+                const relationship = TokenRelationship._fromProtobuf(proto);
+
+                expect(relationship.isKycGranted).to.be.null;
+            });
+
+            it("should map a null kycStatus to null", function () {
+                const relationship = TokenRelationship._fromProtobuf(
+                    protoWith({ kycStatus: null }),
+                );
+
+                expect(relationship.isKycGranted).to.be.null;
+            });
+
+            it("should map a kycStatus of 0 (KycNotApplicable) to null", function () {
+                const relationship = TokenRelationship._fromProtobuf(
+                    protoWith({ kycStatus: 0 }),
+                );
+
+                expect(relationship.isKycGranted).to.be.null;
+            });
+
+            it("should map a kycStatus of 1 (Granted) to true", function () {
+                const relationship = TokenRelationship._fromProtobuf(
+                    protoWith({ kycStatus: 1 }),
+                );
+
+                expect(relationship.isKycGranted).to.be.true;
+            });
+
+            it("should map a kycStatus of 2 (Revoked) to false", function () {
+                const relationship = TokenRelationship._fromProtobuf(
+                    protoWith({ kycStatus: 2 }),
+                );
+
+                expect(relationship.isKycGranted).to.be.false;
+            });
+        });
+
+        describe("freezeStatus", function () {
+            it("should map a missing freezeStatus to null", function () {
+                const proto = protoWith();
+                delete proto.freezeStatus;
+
+                const relationship = TokenRelationship._fromProtobuf(proto);
+
+                expect(relationship.isFrozen).to.be.null;
+            });
+
+            it("should map a null freezeStatus to null", function () {
+                const relationship = TokenRelationship._fromProtobuf(
+                    protoWith({ freezeStatus: null }),
+                );
+
+                expect(relationship.isFrozen).to.be.null;
+            });
+
+            it("should map a freezeStatus of 0 (FreezeNotApplicable) to null", function () {
+                const relationship = TokenRelationship._fromProtobuf(
+                    protoWith({ freezeStatus: 0 }),
+                );
+
+                expect(relationship.isFrozen).to.be.null;
+            });
+
+            it("should map a freezeStatus of 1 (Frozen) to true", function () {
+                const relationship = TokenRelationship._fromProtobuf(
+                    protoWith({ freezeStatus: 1 }),
+                );
+
+                expect(relationship.isFrozen).to.be.true;
+            });
+
+            it("should map a freezeStatus of 2 (Unfrozen) to false", function () {
+                const relationship = TokenRelationship._fromProtobuf(
+                    protoWith({ freezeStatus: 2 }),
+                );
+
+                expect(relationship.isFrozen).to.be.false;
+            });
+        });
+
+        describe("balance", function () {
+            it("should default a null balance to zero", function () {
+                const relationship = TokenRelationship._fromProtobuf(
+                    protoWith({ balance: null }),
+                );
+
+                expect(relationship.balance.toNumber()).to.equal(0);
+            });
+
+            it("should default a missing balance to zero", function () {
+                const proto = protoWith();
+                delete proto.balance;
+
+                const relationship = TokenRelationship._fromProtobuf(proto);
+
+                expect(relationship.balance.toNumber()).to.equal(0);
+            });
+
+            it("should pass an existing Long balance through untouched", function () {
+                const balance = Long.fromNumber(1234);
+
+                const relationship = TokenRelationship._fromProtobuf(
+                    protoWith({ balance }),
+                );
+
+                expect(relationship.balance).to.equal(balance);
+                expect(relationship.balance.toNumber()).to.equal(1234);
+            });
+
+            it("should convert a plain number balance to a Long", function () {
+                const relationship = TokenRelationship._fromProtobuf(
+                    protoWith({ balance: 1234 }),
+                );
+
+                expect(relationship.balance instanceof Long).to.be.true;
+                expect(relationship.balance.toNumber()).to.equal(1234);
+            });
+
+            it("should convert a string balance to a Long", function () {
+                const relationship = TokenRelationship._fromProtobuf(
+                    protoWith({ balance: "1234" }),
+                );
+
+                expect(relationship.balance instanceof Long).to.be.true;
+                expect(relationship.balance.toNumber()).to.equal(1234);
+            });
+        });
+
+        describe("automaticAssociation", function () {
+            it("should keep a true automaticAssociation", function () {
+                const relationship = TokenRelationship._fromProtobuf(
+                    protoWith({ automaticAssociation: true }),
+                );
+
+                expect(relationship.automaticAssociation).to.be.true;
+            });
+
+            it("should keep a false automaticAssociation", function () {
+                const relationship = TokenRelationship._fromProtobuf(
+                    protoWith({ automaticAssociation: false }),
+                );
+
+                expect(relationship.automaticAssociation).to.be.false;
+            });
+
+            it("should map a null automaticAssociation to null", function () {
+                const relationship = TokenRelationship._fromProtobuf(
+                    protoWith({ automaticAssociation: null }),
+                );
+
+                expect(relationship.automaticAssociation).to.be.null;
+            });
+
+            it("should map a missing automaticAssociation to null", function () {
+                const proto = protoWith();
+                delete proto.automaticAssociation;
+
+                const relationship = TokenRelationship._fromProtobuf(proto);
+
+                expect(relationship.automaticAssociation).to.be.null;
+            });
+        });
+    });
+
+    describe("_toProtobuf", function () {
+        /**
+         * @param {object} [overrides]
+         * @returns {TokenRelationship}
+         */
+        function relationshipWith(overrides = {}) {
+            return new TokenRelationship({
+                tokenId,
+                symbol,
+                balance: Long.fromNumber(10),
+                isKycGranted: null,
+                isFrozen: null,
+                automaticAssociation: false,
+                ...overrides,
+            });
+        }
+
+        it("should produce the correct proto structure", function () {
+            const balance = Long.fromNumber(10);
+
+            const proto = relationshipWith({
+                balance,
+                isKycGranted: true,
+                isFrozen: false,
+                automaticAssociation: true,
+            })._toProtobuf();
+
+            expect(proto).to.deep.equal({
+                tokenId: tokenId._toProtobuf(),
+                symbol,
+                balance,
+                kycStatus: 1,
+                freezeStatus: 2,
+                automaticAssociation: true,
+            });
+        });
+
+        it("should map a null isKycGranted to a kycStatus of 0", function () {
+            expect(
+                relationshipWith({ isKycGranted: null })._toProtobuf()
+                    .kycStatus,
+            ).to.equal(0);
+        });
+
+        it("should map a true isKycGranted to a kycStatus of 1", function () {
+            expect(
+                relationshipWith({ isKycGranted: true })._toProtobuf()
+                    .kycStatus,
+            ).to.equal(1);
+        });
+
+        it("should map a false isKycGranted to a kycStatus of 2", function () {
+            expect(
+                relationshipWith({ isKycGranted: false })._toProtobuf()
+                    .kycStatus,
+            ).to.equal(2);
+        });
+
+        it("should map a null isFrozen to a freezeStatus of 0", function () {
+            expect(
+                relationshipWith({ isFrozen: null })._toProtobuf().freezeStatus,
+            ).to.equal(0);
+        });
+
+        it("should map a true isFrozen to a freezeStatus of 1", function () {
+            expect(
+                relationshipWith({ isFrozen: true })._toProtobuf().freezeStatus,
+            ).to.equal(1);
+        });
+
+        it("should map a false isFrozen to a freezeStatus of 2", function () {
+            expect(
+                relationshipWith({ isFrozen: false })._toProtobuf()
+                    .freezeStatus,
+            ).to.equal(2);
+        });
+    });
+
+    describe("round-trip", function () {
+        const statuses = [
+            { proto: 0, value: null },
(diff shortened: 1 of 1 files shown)


test/unit/TokenRelationship.js
```diff
@@ -0,0 +1,377 @@
...
+            const relationship = new TokenRelationship({
+                tokenId,
+                symbol,
+                balance,
+                isKycGranted: true,
+                isFrozen: false,
+                automaticAssociation: true,
+            });
+
+            expect(relationship.tokenId.toString()).to.equal(
```
Fantomasa: `toString()` equality is weaker than the constructor's actual passthrough contract — `TokenId.toString()` drops the checksum, so a constructor that cloned/re-parsed the id (losing a checksum set via `TokenId.fromString("0.0.100-abcde")`) would still pass. `expect(relationship.tokenId).to.equal(tokenId)` pins the real behavior.

test/unit/TokenRelationship.js
```diff
@@ -0,0 +1,377 @@
...
+                isFrozen: null,
+                automaticAssociation: null,
+            });
+
+            expect(Object.isFrozen(relationship)).to.be.true;
+        });
+    });
+
+    describe("_fromProtobuf", function () {
+        it("should map the token id and symbol", function () {
```
Fantomasa: `tokenId` is the only field in this suite without missing/null cases, and it's the only one that actually crashes: it's optional in `ITokenRelationship`, and `TokenRelationship._fromProtobuf({ symbol: "T" })` throws `TypeError: Cannot read properties of undefined (reading 'shardNum')` because the source calls `TokenId._fromProtobuf(undefined)` unguarded. Worth a test pinning that path (or a guard in the source).

test/unit/TokenRelationship.js
```diff
@@ -0,0 +1,377 @@
...
+            const relationship = TokenRelationship._fromProtobuf(protoWith());
+
+            expect(relationship.tokenId.toString()).to.equal("0.0.100");
+            expect(relationship.symbol).to.equal(symbol);
```
Fantomasa: Missing `symbol` is untested too: it's optional (`string|null`) in the proto interface and the source casts it without a guard, so absence yields `relationship.symbol === undefined` on a property documented as `string`. A missing/null twin (like the other fields get) would pin that.

test/unit/TokenRelationship.js
```diff
@@ -0,0 +1,377 @@
...
+            it("should default a missing balance to zero", function () {
+                const proto = protoWith();
+                delete proto.balance;
+
+                const relationship = TokenRelationship._fromProtobuf(proto);
+
+                expect(relationship.balance.toNumber()).to.equal(0);
+            });
+
+            it("should pass an existing Long balance through untouched", function () {
```
Fantomasa: All balance fixtures in the file are signed SDK `Long`s, but a real protobufjs decode yields its own Long class with `unsigned: true`, for which `instanceof Long` (the SDK's) is false — so production always takes the `Long.fromValue` branch and the passthrough branch this test pins is dead on the real decode path. Consider a case built from an actual decoded `proto.TokenRelationship` (or a foreign `{ low, high, unsigned: true }` Long) with a value above 2^63, compared via `toString()` since `toNumber()` is lossy past 2^53.

test/unit/TokenRelationship.js
```diff
@@ -0,0 +1,377 @@
...
+                const balance = Long.fromNumber(1234);
+
+                const relationship = TokenRelationship._fromProtobuf(
+                    protoWith({ balance }),
+                );
+
+                expect(relationship.balance).to.equal(balance);
```
Fantomasa: `.to.equal(balance)` pins reference identity rather than the contract. `Long.fromValue(aLong)` returns a new instance, so an observably identical refactor (unconditional `Long.fromValue`) fails here with an unhelpful "expected Long to equal Long" diff. Asserting instanceof + value, like the neighboring tests, pins the same branch without punishing equivalent refactors.

test/unit/TokenRelationship.js
```diff
@@ -0,0 +1,377 @@
...
+                expect(relationship.balance.toNumber()).to.equal(1234);
+            });
+
+            it("should convert a plain number balance to a Long", function () {
```
Fantomasa: `ITokenRelationship` declares `balance?: (Long|null)` — plain `number`/`string` aren't legal inputs, so this test and the next pin coercion behavior outside the declared contract. A refactor tightening decode to the declared type goes red here with no contract-visible change, while the green tests advertise a tolerance the interface never promised.

test/unit/TokenRelationship.js
```diff
@@ -0,0 +1,377 @@
...
+                const relationship = TokenRelationship._fromProtobuf(
+                    protoWith({ balance: 1234 }),
+                );
+
+                expect(relationship.balance instanceof Long).to.be.true;
```
Fantomasa: `expect(x instanceof Long).to.be.true` only reports "expected false to be true" on failure; `expect(relationship.balance).to.be.instanceOf(Long)` — the form used elsewhere in the repo — shows the offending value. Same on line 202.

test/unit/TokenRelationship.js
```diff
@@ -0,0 +1,377 @@
...
+         * @returns {TokenRelationship}
+         */
+        function relationshipWith(overrides = {}) {
+            return new TokenRelationship({
+                tokenId,
+                symbol,
+                balance: Long.fromNumber(10),
+                isKycGranted: null,
+                isFrozen: null,
+                automaticAssociation: false,
```
Fantomasa, line 294: `_toProtobuf` emits `automaticAssociation: null` when unset, but no `_toProtobuf`/round-trip test ever overrides this `false` default to `null`, so that outbound passthrough is unasserted — a `?? false` change in the source would keep the whole suite green. One case with `automaticAssociation: null` would cover it.

test/unit/TokenRelationship.js
```diff
@@ -0,0 +1,377 @@
...
+
+            const proto = relationshipWith({
+                balance,
+                isKycGranted: true,
+                isFrozen: false,
+                automaticAssociation: true,
+            })._toProtobuf();
+
+            expect(proto).to.deep.equal({
+                tokenId: tokenId._toProtobuf(),
```
Fantomasa: The expected `tokenId` is built by calling `_toProtobuf()` on the same `TokenId` instance the relationship holds, so this leg checks delegation in lockstep rather than the absolute wire shape — and with `0,0` shard/realm a field swap inside `TokenId._toProtobuf` would go unnoticed. A literal `{ shardNum, realmNum, tokenNum }` expectation (with a distinct-values id like `1.2.3`) would pin the actual shape.

test/unit/TokenRelationship.js
```diff
@@ -0,0 +1,377 @@
...
+                symbol,
+                balance,
+                kycStatus: 1,
+                freezeStatus: 2,
+                automaticAssociation: true,
+            });
+        });
+
+        it("should map a null isKycGranted to a kycStatus of 0", function () {
```
Fantomasa: These six tests are the same 5-line block across a 2-field × 3-value matrix — the `statuses` table used by the round-trip describe below already shows the table-driven form. Folding these into one table keeps the same failure localization with one row per value (see `test/unit/TokenInfo.js` for the existing pattern on exactly this mapping).

test/unit/TokenRelationship.js
```diff
@@ -0,0 +1,377 @@
...
+    });
+
+    describe("round-trip", function () {
+        const statuses = [
+            { proto: 0, value: null },
+            { proto: 1, value: true },
+            { proto: 2, value: false },
+        ];
+
+        for (const kyc of statuses) {
```
Fantomasa: `kycStatus` and `freezeStatus` are two independent ternaries in the source, so the 3×3 grid adds no branch coverage over 3 diagonal cases — and these mappings are already asserted by the dedicated describes above and the per-field `_toProtobuf` tests. An intentional mapping change would need edits in three places and turn ~15 tests red for one fact; three diagonal cases would keep the same coverage.

test/unit/TokenRelationship.js
```diff
@@ -0,0 +1,377 @@
...
+                    const proto = protoWith({
+                        kycStatus: kyc.proto,
+                        freezeStatus: freeze.proto,
+                    });
+
+                    const relationship = TokenRelationship._fromProtobuf(proto);
+
+                    expect(relationship.isKycGranted).to.equal(kyc.value);
+                    expect(relationship.isFrozen).to.equal(freeze.value);
+                    expect(relationship._toProtobuf()).to.deep.equal(proto);
```
Fantomasa: This round-trip is lossless only because `protoWith` omits `decimals` (uint32 field 6 of the real `TokenRelationship` message, which the network may populate) — the class silently drops it, and adding `decimals: 8` to the fixture fails all 9 grid cases. A fixture comment or a small test documenting that `decimals` is currently dropped would make the omission intent rather than accident.

test/unit/TokenRelationship.js
```diff
@@ -0,0 +1,377 @@
...
+
+                    const roundTripped = TokenRelationship._fromProtobuf(
+                        relationship._toProtobuf(),
+                    );
+
+                    expect(roundTripped.tokenId.toString()).to.equal(
+                        relationship.tokenId.toString(),
+                    );
+                    expect(roundTripped.symbol).to.equal(relationship.symbol);
+                    expect(roundTripped.balance.toNumber()).to.equal(
```
Fantomasa: These per-field re-assertions can't fail independently of the `deep.equal` on line 340 — and this balance one compares a `Long` with itself (`roundTripped.balance` and `relationship.balance` are the same instance after two passthroughs), so it's a tautology in all 9 grid cases. The `deep.equal` alone carries the round-trip guarantee.

test/unit/TokenRelationship.js
```diff
@@ -0,0 +1,377 @@
...
+                        relationship.isFrozen,
+                    );
+                    expect(roundTripped.automaticAssociation).to.equal(
+                        relationship.automaticAssociation,
+                    );
+                });
+            }
+        }
+
+        it("should normalize a null kycStatus and freezeStatus to 0 on the way back out", function () {
```
Fantomasa: This is a mechanical composition of four facts already asserted individually (null status → null property at lines 83/126; null property → status 0 at lines 280/301), so it can't fail unless one of those tests also fails — it can be dropped without losing detection power.

Fantomasa (changes requested):

test/unit/TokenRelationship.js
sarbojitrana: Fixed — now `expect(relationship.tokenId).to.equal(tokenId)`. Did the same for `balance` in that test, since the constructor is plain assignment for both.

test/unit/TokenRelationship.js
sarbojitrana: Good catch, and confirmed: `TokenRelationship._fromProtobuf({ symbol: "T" })` throws `TypeError: Cannot read properties of undefined (reading 'shardNum')`, and the null case throws the same reading `null`. Added both as tests pinning the current behavior. I left the source unguarded since the issue scopes this PR to the test file — happy to open a follow-up adding the guard if you want it.

test/unit/TokenRelationship.js
sarbojitrana: Added. Confirmed a missing `symbol` yields `undefined` on a property documented as `string`; added a `null` twin alongside it.

test/unit/TokenRelationship.js
sarbojitrana: You are right, and I checked it rather than take it on faith. Encoding and decoding a real `proto.TokenRelationship`:

```
decoded.balance ctor name : Long
instanceof SDK Long       : false
same class object?        : false
unsigned flag             : true
```

protobufjs resolves its own copy of the `long` package, so `instanceof Long` against the SDK's copy is false and production always takes the `Long.fromValue` branch. Added a case built from an actual decode with `18446744073709551000`, asserted via `toString()`.

test/unit/TokenRelationship.js
sarbojitrana: Switched to instanceof + value. Agreed the identity assertion punished an equivalent refactor for no gain — especially given your point above that the passthrough branch is unreachable on a real decode anyway.

test/unit/TokenRelationship.js
sarbojitrana: Dropped the string case. I kept the plain-number one: acceptance criterion 3 on #4007 asks for `balance` to be tested with `null`, a `Long` and a number specifically, so removing it would fail the issue as written. Added a comment marking it as outside the declared `Long|null` type. Say the word if you would rather have it out and I will drop it and note the deviation on the issue.

test/unit/TokenRelationship.js
sarbojitrana: Done, both occurrences — and `to.be.instanceOf` is now used throughout the file.

test/unit/TokenRelationship.js
sarbojitrana: Added a dedicated `automaticAssociation: null` case on the `_toProtobuf` side.

test/unit/TokenRelationship.js
sarbojitrana: Fixed. The shared fixture id is now `1.2.3` so shard/realm/num are distinct, and the expectation is a literal `{ shardNum, realmNum, tokenNum }` rather than a call back into `TokenId._toProtobuf`.

test/unit/TokenRelationship.js
sarbojitrana: Folded into one table following the `test/unit/TokenInfo.js` shape, including the per-assertion message so a failure still names the offending value. Did the same on the `_fromProtobuf` side, which had the same duplication across five values x two fields.

test/unit/TokenRelationship.js
sarbojitrana: Reduced to the three diagonal cases, with a comment noting the two ternaries are independent so the diagonal covers the same branches. Coverage is unchanged at 100%.

test/unit/TokenRelationship.js
sarbojitrana: Confirmed — `decimals` is uint32 field 6 in `services_basic_types.proto`, a real decode populates it, and the class drops it silently. Added a `should drop decimals` test asserting it is absent from both the instance and `_toProtobuf()` output, plus a comment on the fixture explaining why it is omitted there.

test/unit/TokenRelationship.js
sarbojitrana: Removed — the `deep.equal` carries it, and you are right that the balance one was comparing an instance with itself.

test/unit/TokenRelationship.js
sarbojitrana: Dropped.

test/unit/TokenRelationship.js
ddeskov-limechain: Keeping the plain-number case is right — the issue's acceptance criteria ask for it explicitly, and the comment marking it as outside the declared type is exactly the context a future reader needs. Resolving.

test/unit/TokenRelationship.js
ddeskov-limechain: Confirmed the crash reproduces. Yes, please open a follow-up issue for the missing guard — keeping this PR test-only is the correct scope, and the two throwing tests here pin the current behavior until then. Resolving.

Issue #4330, opened 2026-08-18
`Client.ping()` silently probes a random node when the given node account ID is not in the network map


ddeskov-limechain:
### Description

`Client.ping()` reports success for a node account ID that is not in the client's network map — the probe is silently sent to a **random node** from the map instead of failing.

```js
const client = Client.forNetwork({ "127.0.0.1:50211": new AccountId(3) });
client.setOperator(operatorId, operatorKey);

await client.ping("1000000.0.0"); // resolves successfully, probes 0.0.3
```

### Root cause

`ManagedNetwork.getNode(key)` falls back to a random network entry when the key is not present:

https://github.com/hiero-ledger/hiero-sdk-js/blob/main/src/client/ManagedNetwork.js#L407-L432

```js
const lockedNodes = this._network.get(key.toString());
if (lockedNodes) {
    // ... return the requested node
} else {
    const nodes = Array.from(this._network.keys());
    const randomNodeAccountId =
        nodes[Math.floor(Math.random() * nodes.length)];
    // ... return a random node
}
```

Because a keyed lookup can never return `null`/`undefined`, the guard in `Executable._getExecutionNode` that should produce `NodeAccountId not recognized: ...` is unreachable:

https://github.com/hiero-ledger/hiero-sdk-js/blob/main/src/Executable.js#L603-L621

### Impact

- `Client.ping(x)` / `Client.pingAll()` cannot be trusted to have probed the node they name: for an unknown id the probe lands on an arbitrary node and reports success. This fails the TCK `ClientPing` specification (test 3, "Ping a node that isn't in the network map" — expected to fail with an internal error; see hiero-ledger/hiero-sdk-tck#677).
- The same fallback applies to every pinned execution: a query or transaction pinned via `setNodeAccountIds` to an id missing from the map (a typo, or a node removed by an address-book update) silently executes against a random node instead of erroring.

### Suggested fix

A keyed `getNode` miss should throw (making the existing `_getExecutionNode` guard reachable), matching the Java and Go SDKs' behavior of rejecting unknown node ids. If any caller intentionally relies on the random fallback (e.g. re-routing transactions whose pinned node disappeared mid-flight from a network update), that call site should opt into the fallback explicitly rather than getting it by default.

### Steps to reproduce

Run the snippet above against any network, or run the TCK `ClientPing` suite (hiero-ledger/hiero-sdk-tck#677) against the JS SDK server: test 3 fails with the ping resolving instead of erroring.

PR #4312, opened 2026-08-06
fix: replace AccountBalanceQuery ping probe with a COST_ANSWER query

Fantomasa:
## Description

Stage 1 of the [AccountBalanceQuery deprecation proposal](https://github.com/hiero-ledger/sdk-collaboration-hub/blob/main/proposals/account-balance-query-deprecation.md): the consensus node `CryptoService/cryptoGetBalance` endpoint is being throttled to zero in release 77 (testnet ≈ August 13, 2026), which would break the `AccountBalanceQuery`-based liveness probe inside `Client.ping()` / `Client.pingAll()` — and with it Solo readiness checks and connectivity tests.

### Changes

- `Client.ping()` now probes the node with a `CryptoService/getAccountInfo` query for account `<shard>.<realm>.2` with `ResponseType = COST_ANSWER` (via `AccountInfoQuery.getCost()`). The node returns the query fee without executing the query, so no HBAR is charged and the target account is irrelevant. A cost response is success; a gRPC-layer failure propagates, exactly as before. `pingAll()` keeps delegating to `ping()` per node, so node backoff behaviour is unchanged.
- `CostQuery` no longer requires an operator. The query header always carries a payment transaction — the node validates its presence for `COST_ANSWER` but never processes it — and it is simply left unsigned when no operator is configured. The signed path is byte-for-byte unchanged. This preserves `ping()`/`pingAll()` on operator-less clients (e.g. plain `Client.forNetwork(...)` connectivity checks) and makes `Query.getCost()` usable without an operator.
- New unit tests (`test/unit/node/ClientPingMocking.js`) covering the proposal's Stage 1 test plan: the probe sends `cryptoGetInfo` with `COST_ANSWER` for account `0.0.2` (and no `cryptogetAccountBalance`), a cost response resolves as success, `pingAll()` probes every node, an operator-less ping attaches an unsigned payment and succeeds, and a gRPC failure rejects.

No public API changes.

### Verification

- Full node unit suite passes (1931 tests), `tsc`, `eslint`, `prettier` clean.
- Local Solo (consensus node v0.73.0): `ClientIntegrationTest` 16/16 and `AccountInfoIntegrationTest` 5/5 pass; a direct operator-less `ping()`/`pingAll()` against the Solo node succeeds.
- Public testnet: operator-less `ping()` accepted by multiple nodes.

## Related issue(s)

Fixes #4305

## Checklist

- [x] Documented (JSDoc on `Client.ping()`)
- [x] Tested (unit + integration against Solo and testnet)

PR #4331, opened 2026-08-18
feat: add ping and pingAll JSON-RPC methods to the TCK server

ddeskov-limechain:
### Description

Adds `ping` and `pingAll` JSON-RPC methods to the TCK server, mapping directly onto `Client.ping(nodeAccountId)` / `Client.pingAll()`, per the [ClientPing test specification](https://github.com/hiero-ledger/hiero-sdk-tck/blob/main/docs/test-specifications/crypto-service/ClientPing.md) (hiero-ledger/hiero-sdk-tck#677). A successful probe returns a `SUCCESS` result; a failed probe rejects and the server's existing error middleware reports it as the JSON-RPC error object the specification requires (`-32001` with `data.status` for precheck rejections, `-32603` otherwise).

Also bumps the server's pinned SDK from 2.84.0 to 2.88.0 — the latest release, which carries the COST_ANSWER `getAccountInfo` ping probe (#4312) that the specification verifies on the wire.

### Verification

Ran the TCK `ClientPing` suite (hiero-ledger/hiero-sdk-tck#689) against this server on a local solo deployment: 5 of 6 tests pass, including the wire-level assertion that `Client.ping()` sends a COST_ANSWER `CryptoService/getAccountInfo` query for account `0.0.2` and no `CryptoService/cryptoGetBalance`.

The remaining failure (test 3, "Ping a node that isn't in the network map") is an SDK conformance gap, not a server issue: `ManagedNetwork.getNode` silently substitutes a random node for an unknown node account ID, so the ping resolves instead of erroring — tracked in #4330. Keeping this server a thin passthrough means the suite stays red for JS until the SDK is fixed, which is the point of the conformance test.

### Related issues

- hiero-ledger/hiero-sdk-tck#677 (JavaScript task)
- #4330

### Checklist

- [x] Documented (JSDoc on both methods)
- [x] Tested (TCK `ClientPing` suite, live run)

### DIFF
--- a/tck/package.json
+++ b/tck/package.json
@@ -13,7 +13,7 @@
     "repository": "https://github.com/hiero-ledger/hiero-sdk-js",
     "dependencies": {
         "@hiero-ledger/proto": "^2.30.0",
-        "@hiero-ledger/sdk": "2.84.0",
+        "@hiero-ledger/sdk": "2.87.0",
         "ansi-regex": "6.2.2",
         "ansi-styles": "6.2.3",
         "body-parser": "2.3.0",
--- a/tck/methods/sdk.ts
+++ b/tck/methods/sdk.ts
@@ -21,6 +21,14 @@ interface SetOperatorParams {
     readonly sessionId?: string;
 }
 
+/**
+ * Parameters for probing a single node's health
+ */
+interface PingParams {
+    readonly nodeAccountId: string;
+    readonly sessionId?: string;
+}
+
 export default {
     /**
      * Reports the SDK name and version under test, read from the installed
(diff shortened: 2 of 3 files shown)


tck/package.json
```diff
@@ -13,7 +13,7 @@
     "repository": "https://github.com/hiero-ledger/hiero-sdk-js",
     "dependencies": {
         "@hiero-ledger/proto": "^2.30.0",
-        "@hiero-ledger/sdk": "2.84.0",
+        "@hiero-ledger/sdk": "2.87.0",
```
Fantomasa: Non-blocking: 2.88.0 was released on Sep 9, so this could move to the latest release if you touch the lockfile again. Also worth knowing: the Dockerfile runs `pnpm add @hiero-ledger/sdk@^2.70.0` before installing, so the Docker image floats to the latest 2.x and only local runs honour this pin. Fine for now, just noting it so nobody expects the image to be pinned.

ddeskov-limechain: Bumped the pin to 2.88.0 in d466d49. Agreed on the Dockerfile floating to latest 2.x; leaving that as-is for now.

Fantomasa (approved):

PR #4333, opened 2026-08-18
fix: reject unknown node account IDs

### LINKED ISSUE #4330: `Client.ping()` silently probes a random node when the given node account ID is not in the network map

saqibmanzoor-dev:
**Description**:

Fix `ManagedNetwork.getNode()` to reject unknown node account IDs instead of silently routing requests to a random node.

* Throw an error when a requested node account ID is not present in the network map
* Prevent pinned executions from being routed to an unintended random node

**Related issue(s)**:

Fixes #4330

**Notes for reviewer**:

Previously, `getNode()` selected a random node when the requested node account ID was not found in the network map. This change makes the lookup fail with `NodeAccountId not recognized: ...` instead.

**Checklist**


Jexsie (changes requested):
Hey @saqibmanzoor-dev. Thank you for your contribution. Can you have your commits signed. Follow the articles [here](https://docs.github.com/en/authentication/managing-commit-signature-verification) to setup your machine for signing commits. Thank you

ddeskov-limechain (changes requested):
Thanks for picking this up. The change is in the right place and it does fix the ping case, but a few things are needed before it fully closes #4330:

1. **One error path, not two.** `_getExecutionNode` still has the `currentNode == null` guard with the same message, which is now unreachable. Simplest is to drop the fallback branch and let `getNode` return `undefined` on a miss, so the existing guard fires; or keep the throw and delete the guard. Not both.
2. **Multi-node pinned executions.** A transaction pinned to several nodes where one has since dropped out of the network map (address-book refresh, or the `updateNetwork` after INVALID_NODE_ACCOUNT) used to recover by advancing to the next node. With this change it throws on the first unknown ID. Please skip an unknown pinned ID while other pinned IDs remain, and throw only when none of them resolve.
3. **Tests.** Please add a `ManagedNetwork` unit test for the unknown-key case, and an `Executable` test showing a pinned unknown ID rejects with `NodeAccountId not recognized` before any channel call. The stubbing pattern is in `test/unit/Executable.js`.
4. **CHANGELOG.** Add a `### Fixed` entry under `# Unreleased` with a link to #4330, and note the behavior change: a receipt or record query pinned to a node that has since left the map now errors instead of silently going to another node.
5. **Formatting and DCO.** Run `pnpm format` (the `}  else {` block is off), and sign off the commit with `git commit --amend -s` so the DCO check passes.

I'll approve the CI run once the sign-off is in.

One process note for next time: please comment on the issue and ask to be assigned before starting on it. That way we can confirm nobody else is already working on it and avoid two people fixing the same thing. I've assigned #4330 to you now.

ddeskov-limechain:
@saqibmanzoor-dev are you still working on this PR? We need to deliver it and merge it. We will be forced to do it from our side if you do not respond soon.

saqibmanzoor-dev:
Hi @ddeskov-limechain , yes, I’ve fixed the DCO issue on the original commit and pushed the corrected history with --force-with-lease. The existing PR has been updated. Thanks for the reminder!

ddeskov-limechain:
@saqibmanzoor-dev Just a reminder that the DCO is still failing as well as some tests on the CI.

ddeskov-limechain:
Closing this PR: it has not been updated for too long after the review on 2026-09-08 (the DCO check was still failing and the review items were not addressed). #4330 is now implemented in #4375, which builds on the same `getNode` change and adds the multi-node skip, tests and CHANGELOG entry. Thank you for the initial work.

Issue #4334, opened 2026-08-20
`MirrorNodeAccountBalanceQuery` returns 0 instead of failing for an unknown account


Dosik13:
### Description

`MirrorNodeAccountBalanceQuery` (added in #4320, released in v2.87.0) reports `hbars = 0` for an account the mirror node does not know. The balances endpoint answers an unknown account with HTTP 200 and an empty `balances` array rather than a 404, and the parse collapses that to zero:

```js
const balance = response.balances?.[0]?.balance ?? 0;
```

Two problems:

1. An error condition becomes a plausible-looking balance. A wallet renders "0 ℏ" for a typo'd or wrong-network account instead of reporting the account does not exist.
2. It breaks migration from `AccountBalanceQuery`, which failed with `INVALID_ACCOUNT_ID`. Callers that caught that status silently get a zero.

An account that genuinely exists holding nothing returns a populated entry with `"balance": 0`, so the two cases are distinguishable — confirmed against the public testnet endpoint.

### Expected

An account the mirror node does not know should fail with `INVALID_ACCOUNT_ID`, as `AccountBalanceQuery` did. This is what the other SDKs already do:

- hiero-sdk-java throws `PrecheckStatusException(Status.INVALID_ACCOUNT_ID, null)` — "Report it the way the consensus node did"
- hiero-sdk-go (#1814) returns `ErrHederaPreCheckStatus{Status: StatusInvalidAccountID}`

ddeskov-limechain:
Fixed by #4335 and released in [v2.88.0](https://github.com/hiero-ledger/hiero-sdk-js/releases/tag/v2.88.0): `MirrorNodeAccountBalanceQuery.execute()` now throws `MirrorNodeStatusError` with `Status.InvalidAccountId` for an account the mirror node does not know, matching the Java and Go SDKs. An account that exists but holds nothing still returns `0 ℏ`.

PR #4320, opened 2026-08-11
feat(sdk): add MirrorNodeAccountBalanceQuery as mirror node replacement for AccountBalanceQuery

ddeskov-limechain:
## Summary

Implements the mirror-node replacement for the deprecated `AccountBalanceQuery` per #4286, following the [migration proposal](https://github.com/hiero-ledger/sdk-collaboration-hub/blob/main/proposals/account-balance-query-mirror-node-migration.md) as merged (updated by [sdk-collaboration-hub#280](https://github.com/hiero-ledger/sdk-collaboration-hub/pull/280) after mirror node team review). New standalone class following the existing mirror REST query pattern (`MirrorNodeContractCallQuery`, `FeeEstimateQuery`) — plain HTTP, no `Query` base class, no payment.

```js
const balance = await new MirrorNodeAccountBalanceQuery()
    .setAccountId("0.0.12345") // or EVM address / alias / contract ID
    .execute(client);
console.log(balance.hbars.toString());
```

- Single `GET /api/v1/balances?account.id={id}` request; returns the new read-only `MirrorNodeAccountBalance` type carrying `hbars` only. Token balances are deliberately out of scope per the proposal (the balances endpoint does not paginate them, and unbounded `/tokens` pagination is a mirror node DDoS risk).
- `setAccountId` accepts `shard.realm.num`, EVM address, public-key alias, and contract IDs (the balances endpoint resolves all forms — no separate `setContractId`). Alias-key IDs are sent as the base32 alias the mirror node requires; the DER-hex form from `AccountId.toString()` gets a 400.
- A non-existent account yields an empty `balances` array (HTTP 200), returned as zero hbars per the proposal.
- 400 → immediate error carrying the mirror error detail, no retry; 5xx/network failures → exponential backoff from `client.minBackoff` up to `client.maxAttempts`. `requestTimeout` (parameter or client default) is a total operation budget across all attempts, and `AbortSignal.timeout` is guarded for React Native.
- Eventual consistency (mirror ingest lag) and the 2^53−1 JSON number precision limit are documented in the class JSDoc.

## Testing

- 13 unit tests: isomorphic suite (setters, ID-form resolution to the `account.id` parameter, empty-array → zero, immutability) + MSW wire suite (parameter and parsing, empty-array → zero, 400 no-retry, 503-then-200 retry).
- Integration test (CI solo network): operator balance via polling with a deadline (no fixed sleeps), non-existent account → zero hbars.
[...]

PR #4335, opened 2026-08-20
fix: mirror balance unknown account

Dosik13:
## Description

`MirrorNodeAccountBalanceQuery` (added in #4320, released in v2.87.0) returns `hbars = 0` for an account the mirror node does not know. The balances endpoint reports an unknown account as HTTP 200 with an empty `balances` array rather than a 404, and the parse collapsed that to zero:

```js
const balance = response.balances?.[0]?.balance ?? 0;
```

That turns an error condition into a plausible-looking balance — a wallet renders "0 ℏ" for a typo'd or wrong-network account — and it breaks callers that previously caught `INVALID_ACCOUNT_ID` from the `AccountBalanceQuery` this query replaces.

### Changes

- `MirrorNodeAccountBalanceQuery.execute` now distinguishes three response shapes instead of two, matching hiero-sdk-go#1814: a `balances` field that is not an array (missing, null, or the wrong shape) is a malformed response and throws a plain `Error`; an **empty** array means the mirror node knows no such account and throws `PrecheckStatusError` carrying `Status.InvalidAccountId`; otherwise the balance is parsed as before. An account that exists holding nothing returns a populated entry with `"balance": 0`, so a genuine zero is never mistaken for a missing account.
- `StatusError.transactionId` and `PrecheckStatusError.nodeId` are now nullable, so a status-bearing error can describe a failure that reached no consensus node — which is what a mirror-node REST query is. `PrecheckStatusError`'s message names only the parts that exist: an error with a transaction and a node keeps its byte-identical wording, and a transaction-less one reads `query failed with status INVALID_ACCOUNT_ID` rather than claiming a precheck occurred. Both `toJSON()` implementations serialize the absent fields as `null` instead of throwing.
- The class JSDoc documents two consequences of moving off the consensus node: a freshly created account transiently fails with `INVALID_ACCOUNT_ID` until the mirror node ingests it, so create-then-read flows must retry; and a **deleted** account is reported as an ordinary zero balance, because the balances endpoint does not expose its deleted flag (`AccountBalanceQuery` failed with `ACCOUNT_DELETED`). Neither is fixable within a single request to this endpoint.

## Related issue(s)
#4334 

## Checklist

- [x] Documented (JSDoc on the query class and `execute`)
- [x] Tested (unit; integration test updated but needs CI)


src/query/MirrorNodeAccountBalanceQuery.js
```diff
@@ -104,22 +116,35 @@ export default class MirrorNodeAccountBalanceQuery {
...
+                contractFunctionResult: null,
+            });
+        }
 
         return new MirrorNodeAccountBalance({
-            hbars: Hbar.fromTinybars(Long.fromValue(balance)),
+            hbars: Hbar.fromTinybars(
+                Long.fromValue(response.balances[0].balance),
+            ),
         });
```
Fantomasa: The guard covers the array but not the entry: a missing `balance` field crashes inside `long` with an opaque `TypeError: Cannot read properties of undefined (reading 'low')`, and a non-number silently misparses (`Long.fromValue("abc")` is `0`, the silent-zero class this PR fixes).

```suggestion
        const balance = response.balances[0].balance;
        if (typeof balance !== "number") {
            throw new Error(
                `Failed to query ${url}: balance is not a number`,
            );
        }

        return new MirrorNodeAccountBalance({
            hbars: Hbar.fromTinybars(Long.fromValue(balance)),
        });
```

src/query/MirrorNodeAccountBalanceQuery.js
```diff
@@ -104,22 +116,35 @@ export default class MirrorNodeAccountBalanceQuery {
...
-                client,
-                deadline,
-            )
+            await this._fetchJson(url, client, deadline)
         );
 
-        // The balances endpoint returns an empty array (not a 404) for an
-        // account that does not exist; the balance is zero in that case.
-        const balance = response.balances?.[0]?.balance ?? 0;
+        if (!Array.isArray(response.balances)) {
```
Fantomasa: A 200 whose JSON body is literally `null` throws a raw TypeError here before the clear error below can fire.

```suggestion
        if (!Array.isArray(response?.balances)) {
```

examples/account/get-balance-mirror-node.js
```diff
@@ -43,7 +45,11 @@ async function main() {
             `${operatorId.toString()} balance = ${balance.hbars.toString()}`,
         );
     } catch (error) {
-        console.error(error);
+        if (error.status === Status.InvalidAccountId) {
+            console.error(`${operatorId.toString()} does not exist`);
+        } else {
+            console.error(error);
+        }
```
Fantomasa: The identity compare breaks in mixed require/import graphs (two `Status` singletons), and the repo's error-path idiom is a type guard first (`error-handling-example.js:72`). Also "does not exist" contradicts the header's own caveat that a fresh account transiently fails until ingested. Suggest `error instanceof PrecheckStatusError && error.status === Status.InvalidAccountId` (plus the import) and a message like "is not known to the mirror node yet (nonexistent, or not yet ingested)".

src/query/MirrorNodeAccountBalanceQuery.js
Fantomasa: A missing `balance` field crashes inside `long` with an opaque TypeError, and a non-number silently misparses (`Long.fromValue("abc")` is `0`).

```suggestion
        const balance = response.balances[0].balance;
        if (typeof balance !== "number") {
            throw new Error(
                `Failed to query ${url}: balance is not a number`,
            );
        }

        return new MirrorNodeAccountBalance({
            hbars: Hbar.fromTinybars(Long.fromValue(balance)),
        });
```

src/StatusError.js
```diff
@@ -9,15 +9,16 @@
...
  * @property {string} message
  */
 
 export default class StatusError extends Error {
     /**
      * @param {object} props
      * @param {Status} props.status
-     * @param {TransactionId} props.transactionId
+     * @param {?TransactionId} props.transactionId - null for failures that
+     * are not tied to a transaction, such as a mirror-node REST query
```
Fantomasa: This widens the emitted d.ts for `ReceiptStatusError` and `RecordStatusError` too, although they always set the field, so `e.transactionId.toString()` stops compiling under `strictNullChecks`. Consider redeclaring the property non-null in those subclasses.

src/PrecheckStatusError.js
```diff
@@ -34,14 +34,19 @@ export default class PrecheckStatusError extends StatusError {
     /**
      * @param {object} props
      * @param {Status} props.status
-     * @param {TransactionId} props.transactionId
-     * @param {AccountId} props.nodeId
+     * @param {?TransactionId} props.transactionId - null when the failure is
+     * not tied to a transaction
+     * @param {?AccountId} props.nodeId - null when no consensus node was
+     * involved
```
Fantomasa: The class JSDoc above still describes only the transaction pre-check case and promises the transaction ID; worth updating for the new transaction-less role.

src/query/MirrorNodeAccountBalanceQuery.js
Fantomasa: A 200 whose JSON body is literally `null` throws a raw TypeError here instead of the clear error below.

```suggestion
        if (!Array.isArray(response?.balances)) {
```

src/PrecheckStatusError.js
```diff
@@ -13,7 +13,7 @@ import StatusError from "./StatusError.js";
  * @typedef {object} PrecheckStatusErrorJSON
  * @property {string} name
  * @property {string} status
- * @property {string} transactionId
+ * @property {?string} transactionId
  * @property {?string | null} nodeId
```
Fantomasa:
```suggestion
 * @property {?string} nodeId
```

Same type spelled twice; matches the `?string` used for `transactionId` above.

src/query/MirrorNodeAccountBalanceQuery.js
Fantomasa: A 200 whose JSON body is literally `null` throws a raw TypeError here instead of the clear error below.

```suggestion
        if (!Array.isArray(response?.balances)) {
```

src/query/MirrorNodeAccountBalanceQuery.js
```diff
@@ -104,22 +116,35 @@ export default class MirrorNodeAccountBalanceQuery {
...
+        // An existing account with no hbar still has a `"balance": 0` entry, so
+        // an empty list means the account is unknown.
+        if (response.balances.length === 0) {
+            throw new PrecheckStatusError({
+                status: Status.InvalidAccountId,
+                transactionId: null,
+                nodeId: null,
+                contractFunctionResult: null,
+            });
+        }
```
Fantomasa, line 147: Deliberate per #4334, but it reverses released v2.87.0 behavior (unknown account returned 0) and the nullable `transactionId`/`nodeId` types reach TS consumers; worth an explicit release-notes callout since the PR title carries no breaking marker.

src/query/MirrorNodeAccountBalanceQuery.js
Fantomasa, line 151: A missing `balance` field previously returned 0; now it crashes inside `long` with an opaque TypeError, and a non-number still silently misparses (`Long.fromValue("abc")` is `0`).

```suggestion
        const balance = response.balances[0].balance;
        if (typeof balance !== "number") {
            throw new Error(
                `Failed to query ${url}: balance is not a number`,
            );
        }

        return new MirrorNodeAccountBalance({
            hbars: Hbar.fromTinybars(Long.fromValue(balance)),
        });
```

src/StatusError.js
Fantomasa: This widens the emitted d.ts for `ReceiptStatusError` and `RecordStatusError` too, although they always set the field, so `e.transactionId.toString()` stops compiling under `strictNullChecks`. Consider redeclaring the property non-null in those subclasses.

src/PrecheckStatusError.js
Fantomasa, line 17:
```suggestion
 * @property {?string} nodeId
```

Same type spelled twice; matches the `?string` used for `transactionId` above.

src/PrecheckStatusError.js
Fantomasa: The class JSDoc above still describes only the transaction pre-check case and promises the transaction ID; worth updating for the new transaction-less role.

Fantomasa (changes requested):

src/query/MirrorNodeAccountBalanceQuery.js
Dosik13: done

src/StatusError.js
Dosik13: done

src/query/MirrorNodeAccountBalanceQuery.js
Dosik13: done

src/PrecheckStatusError.js
Dosik13: This file is no longer touched by the PR, so the ?string on the line above that this was matching is gone.

src/PrecheckStatusError.js
Dosik13: I removed the changes from this file so i think we are good

Fantomasa (approved):

PR #4338, opened 2026-08-22
fix(transaction): reject bodies that set more than one data field

### LINKED ISSUE #4337: Transaction.fromBytes should reject transaction bodies that set more than one data field
## Summary

`Transaction.fromBytes` should reject a transaction body that sets more than one `data` field.

## Details

A transaction body is meant to set exactly one `data` field. That field is the transaction type, for example `cryptoTransfer` or `cryptoDelete`.

It is possible to craft bytes that set more than one `data` field. protobufjs, the decoder we use, keeps only the last such field and drops the others while decoding. The dropped fields do not appear on the decoded object. So we cannot notice the problem by looking at the decoded object. The body is ambiguous.

## Impact

Low severity. This needs already-trusted input, such as an enrolled member with a signing key, or a transaction file shared inside a trusted group. There is no outsider path.

## Suggested fix

Before decoding, read the raw body bytes and count how many fields belong to the `data` group. If more than one is set, reject the transaction at parse time. This is simpler and safer than re-encoding and comparing, which is unreliable because different SDKs order fields differently.

ddeskov-limechain:
## What

`Transaction.fromBytes` now rejects a transaction body that sets more than one `data` field.

## Why

A transaction body should set exactly one `data` field, which is the transaction type. It is possible to craft bytes that set more than one. When that happens, protobufjs keeps only the last field and drops the rest while decoding. The dropped fields are not visible on the decoded object, so they cannot be checked after decoding. The body is ambiguous.

This is a low severity, defense-in-depth fix. It only affects already-trusted input, such as an enrolled member or a shared transaction file. There is no known outsider path.

## How

Before decoding, we read the raw body bytes and count how many fields belong to the transaction `data` group. If more than one is set, we throw and refuse to build the transaction. The list of `data` field numbers is taken from the transaction registry, so it stays correct as new transaction types are added.

We do not re-encode and compare against the original bytes. That is unreliable because different SDKs order fields differently.

## Tests

- Added a unit test that builds a body with two `data` fields and checks that `fromBytes` throws.
- Ran the Transaction unit tests and they pass.

Closes #4337

### DIFF
--- a/src/transaction/Transaction.js
+++ b/src/transaction/Transaction.js
@@ -83,26 +83,28 @@ export const TRANSACTION_REGISTRY = new Map();
  * @type {Set<number>}
  */
 let dataFieldNumbers = new Set();
-let dataFieldRegistrySize = -1;
 
 /**
  * @returns {Set<number>}
  */
 function getDataFieldNumbers() {
-    if (
-        dataFieldNumbers.size === 0 ||
-        TRANSACTION_REGISTRY.size !== dataFieldRegistrySize
-    ) {
-        /** @type {Set<number>} */
-        const numbers = new Set();
-        for (const name of TRANSACTION_REGISTRY.keys()) {
+    if (dataFieldNumbers.size === 0) {
+        const prototype = HieroProto.proto.TransactionBody.prototype;
+        for (const name of Object.keys(prototype)) {
+            // The generated `data` oneof getter returns the name only if it
+            // belongs to the oneof, which lets us test membership.
+            const carrier = Object.create(prototype);
+            carrier[name] = 1;
+            if (carrier.data !== name) {
+                continue;
+            }
             const probe = HieroProto.proto.TransactionBody.encode({
                 [name]: {},
             }).finish();
-            numbers.add(HieroProto.Reader.create(probe).uint32() >>> 3);
+            dataFieldNumbers.add(
+                HieroProto.Reader.create(probe).uint32() >>> 3,
+            );
         }
-        dataFieldNumbers = numbers;
-        dataFieldRegistrySize = TRANSACTION_REGISTRY.size;
     }
     return dataFieldNumbers;
 }
--- a/src/transaction/Transaction.js
+++ b/src/transaction/Transaction.js
@@ -75,6 +75,70 @@ function isMultiTransactionIdType(transactionDataCase) {
 export const TRANSACTION_REGISTRY = new Map();
 
 /**
+ * Field numbers of the `data` oneof on `TransactionBody`, i.e. the transaction
+ * type fields. A valid body sets exactly one of these. Derived from the
+ * transaction registry so it stays in sync as transaction types are added.
+ * Rebuilt when the registry grows, because types register lazily on import.
+ *
+ * @type {Set<number>}
+ */
+let dataFieldNumbers = new Set();
+let dataFieldRegistrySize = -1;
+
+/**
+ * @returns {Set<number>}
+ */
+function getDataFieldNumbers() {
+    if (
+        dataFieldNumbers.size === 0 ||
+        TRANSACTION_REGISTRY.size !== dataFieldRegistrySize
+    ) {
+        /** @type {Set<number>} */
+        const numbers = new Set();
+        for (const name of TRANSACTION_REGISTRY.keys()) {
+            const probe = HieroProto.proto.TransactionBody.encode({
+                [name]: {},
+            }).finish();
+            numbers.add(HieroProto.Reader.create(probe).uint32() >>> 3);
+        }
+        dataFieldNumbers = numbers;
+        dataFieldRegistrySize = TRANSACTION_REGISTRY.size;
+    }
+    return dataFieldNumbers;
+}
+
+/**
+ * Reject a transaction body that sets more than one `data` field.
+ *
+ * A crafted `TransactionBody` can carry several `data` fields on the wire.
+ * protobufjs keeps only the last one when decoding, so the extra fields are
+ * invisible on the decoded object and cannot be counted there. We inspect the
+ * raw bytes instead, before decoding, so an ambiguous body is refused.
+ *
+ * @param {Uint8Array} bodyBytes
+ */
+function assertSingleDataField(bodyBytes) {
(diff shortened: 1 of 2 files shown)


src/transaction/Transaction.js
```diff
@@ -74,6 +74,70 @@ function isMultiTransactionIdType(transactionDataCase) {
...
+            const probe = HieroProto.proto.TransactionBody.encode({
+                [name]: {},
+            }).finish();
+            numbers.add(HieroProto.Reader.create(probe).uint32() >>> 3);
+        }
+        dataFieldNumbers = numbers;
+        dataFieldRegistrySize = TRANSACTION_REGISTRY.size;
+    }
+    return dataFieldNumbers;
+}
```
Fantomasa: Nice work on this one, the wire-level scan is the right approach and the premise checks out: protobufjs really does keep only the last `data` field, with no trace of the earlier one on the decoded object.

There is one gap though: `TRANSACTION_REGISTRY` covers only 52 of the 69 fields in the `data` oneof, so the types the SDK does not register (`systemDelete`, `systemUndelete`, `uncheckedSubmit`, `nodeStakeUpdate`, `stateSignatureTransaction`, and 12 more) are never counted. A body carrying `systemDelete` + `cryptoTransfer` passes this check and decodes as a plain `TransferTransaction`, silently dropping the `systemDelete`. That is exactly the ambiguity this PR wants to reject.

The generated prototype can enumerate the full oneof instead: every field name is an enumerable prototype property, and the generated `data` getter doubles as a membership test. That covers all 69 fields, needs no registry-growth rebuild, and stays in sync with the protobuf package on its own. Verified locally: the smuggled `systemDelete` + `cryptoTransfer` body now throws, and all 99 Transaction unit tests still pass.

```suggestion
/**
 * Field numbers of the `data` oneof on `TransactionBody`, i.e. the transaction
 * type fields. A valid body sets exactly one of these. Derived from the
 * generated protobuf prototype so it covers every field in the oneof, even
 * transaction types the SDK does not register (e.g. `systemDelete`), and it
 * stays in sync as the protobuf definitions evolve.
 *
 * @type {Set<number> | null}
 */
let dataFieldNumbers = null;

/**
 * @returns {Set<number>}
 */
function getDataFieldNumbers() {
    if (dataFieldNumbers == null) {
        /** @type {Set<number>} */
        const numbers = new Set();
        const prototype = HieroProto.proto.TransactionBody.prototype;
        for (const name of Object.keys(prototype)) {
            // The generated `data` oneof getter returns the name only if it
            // belongs to the oneof, which lets us test membership.
            const carrier = Object.create(prototype);
            carrier[name] = 1;
            if (carrier.data !== name) {
                continue;
            }
            const probe = HieroProto.proto.TransactionBody.encode({
                [name]: {},
            }).finish();
            numbers.add(HieroProto.Reader.create(probe).uint32() >>> 3);
        }
        dataFieldNumbers = numbers;
    }
    return dataFieldNumbers;
}
```

Fantomasa: Nice catch, and the wire-level scan is the right approach. One gap: `TRANSACTION_REGISTRY` covers only 52 of the 69 fields in the `data` oneof, so a body pairing an unregistered type with a registered one (e.g. `systemDelete` + `cryptoTransfer`) passes the check and decodes as a plain `TransferTransaction`, silently dropping the `systemDelete`.

Deriving the set from the generated prototype covers the full oneof (the `data` getter doubles as a membership test). Verified locally: the smuggled body now throws and all 99 Transaction unit tests still pass.

```suggestion
/**
 * Field numbers of the `data` oneof on `TransactionBody`, i.e. the transaction
 * type fields. A valid body sets exactly one of these. Derived from the
 * generated protobuf prototype so it covers every field in the oneof, even
 * transaction types the SDK does not register (e.g. `systemDelete`).
 *
 * @type {Set<number> | null}
 */
let dataFieldNumbers = null;

/**
 * @returns {Set<number>}
 */
function getDataFieldNumbers() {
    if (dataFieldNumbers == null) {
        /** @type {Set<number>} */
        const numbers = new Set();
        const prototype = HieroProto.proto.TransactionBody.prototype;
        for (const name of Object.keys(prototype)) {
            // The generated `data` oneof getter returns the name only if it
            // belongs to the oneof, which lets us test membership.
            const carrier = Object.create(prototype);
            carrier[name] = 1;
            if (carrier.data !== name) {
                continue;
            }
            const probe = HieroProto.proto.TransactionBody.encode({
                [name]: {},
            }).finish();
            numbers.add(HieroProto.Reader.create(probe).uint32() >>> 3);
        }
        dataFieldNumbers = numbers;
    }
    return dataFieldNumbers;
}
```

Fantomasa, line 112: Nice catch, and the wire-level scan is the right approach. One gap: `TRANSACTION_REGISTRY` covers only 52 of the 69 fields in the `data` oneof, so a body pairing an unregistered type with a registered one (e.g. `systemDelete` + `cryptoTransfer`) passes the check and decodes as a plain `TransferTransaction`, silently dropping the `systemDelete`.

Deriving the set from the generated prototype covers the full oneof (the `data` getter doubles as a membership test). Verified locally: the smuggled body now throws and all 99 Transaction unit tests still pass. The doc comment above would need a matching touch-up.

```suggestion
let dataFieldNumbers = new Set();

/**
 * @returns {Set<number>}
 */
function getDataFieldNumbers() {
    if (dataFieldNumbers.size === 0) {
        const prototype = HieroProto.proto.TransactionBody.prototype;
        for (const name of Object.keys(prototype)) {
            // The generated `data` oneof getter returns the name only if it
            // belongs to the oneof, which lets us test membership.
            const carrier = Object.create(prototype);
            carrier[name] = 1;
            if (carrier.data !== name) {
                continue;
            }
            const probe = HieroProto.proto.TransactionBody.encode({
                [name]: {},
            }).finish();
            dataFieldNumbers.add(
                HieroProto.Reader.create(probe).uint32() >>> 3,
            );
        }
    }
    return dataFieldNumbers;
}
```

Fantomasa (changes requested):

Fantomasa (dismissed):
LGTM

Fantomasa (approved):

PR #4343, opened 2026-08-27
fix: restore `SystemDeleteTransaction`/`SystemUndeleteTransaction`

### LINKED ISSUE #4342: `SystemDeleteTransaction` / `SystemUndeleteTransaction` removal in v2.87.0 breaks `Transaction.fromBytes()`
### Description

#4269 (tracked by #4264) removed `SystemDeleteTransaction` and `SystemUndeleteTransaction` as long-deprecated, and that shipped as a breaking change in `v2.87.0`. Both are still in use.

The removal reached further than the two missing exports. Each class registered itself in the transaction registry at import time:

```js
TRANSACTION_REGISTRY.set("systemDelete", SystemDeleteTransaction._fromProtobuf);
```

Deleting the files unregistered the `systemDelete` and `systemUndelete` keys, so `Transaction.fromBytes()` can no longer deserialize an **already-signed** system delete or undelete transaction — it now hits the registry's own "we forgot to update the registry" guard at `src/transaction/Transaction.js:431`:

```
Error: (BUG) Transaction.fromBytes() not implemented for type systemDelete
```

That affects any consumer round-tripping existing transaction bytes, whether or not it ever calls the constructors — including tools that only inspect or co-sign transactions built elsewhere.

### Steps to reproduce

On `@hiero-ledger/sdk@2.87.0`:

**1. The exports are gone**

```js
import { SystemDeleteTransaction } from "@hiero-ledger/sdk";
// undefined on 2.87.0; present in 2.86.2
```

**2. Existing signed bytes no longer parse** (the sharper regression)

```js
import { Transaction } from "@hiero-ledger/sdk";

// bytes for a systemDelete transaction, produced by SDK <= 2.86.2,
// another Hiero SDK, or any tool that builds one
const tx = Transaction.fromBytes(bytes);

// 2.86.2 -> SystemDeleteTransaction
// 2.87.0 -> Error: (BUG) Transaction.fromBytes() not implemented for type systemDelete
```

The same applies to `systemUndelete`. Note the consensus node still exposes both RPCs on `FileService` and `SmartContractService`, and the protobufs still define both bodies — only the JS SDK dropped them.

### Expected

`Transaction.fromBytes()` should deserialize `systemDelete` and `systemUndelete` bodies, and both classes should be constructible, as in v2.86.2.

### Hedera network

Any (not network-dependent; `fromBytes` fails locally, before any request).

### Version

v2.87.0 (regression introduced by #4269; last working version 2.86.2).

Dosik13:
## Description

`SystemDeleteTransaction` and `SystemUndeleteTransaction` were removed in #4269 ("remove long-deprecated classes and methods"), which shipped as a breaking change in v2.87.0. The transaction tool still builds and parses these transaction bodies through the JS SDK, so this restores both classes.

The removal was wider than the two missing constructors. Both files registered themselves in the transaction registry at import time:

```js
TRANSACTION_REGISTRY.set("systemDelete", SystemDeleteTransaction._fromProtobuf);
```

Deleting the files unregistered those keys, so `Transaction.fromBytes()` could no longer deserialize an **already-signed** system delete/undelete transaction at all — a break for any tool round-tripping existing transaction bytes, independent of whether it ever calls the constructors.

Both classes are restored byte-for-byte as they were before #4269 and remain marked `@deprecated`. Every other removal from #4269 (`LiveHash*`, `EthereumFlow`, `AccountAllowanceAdjustTransaction`, and `toSolidityAddress()` / `fromSolidityAddress()`) is deliberately left removed.

### Changes

- Restored `src/system/SystemDeleteTransaction.js` and `src/system/SystemUndeleteTransaction.js`, byte-identical to their v2.86.2 content, `@deprecated` tags retained.
- Re-added both exports to `src/exports.js` with their `// eslint-disable-next-line deprecation/deprecation` comments, in their original alphabetical position. #4269 left a stray blank line at that spot; this consumes it, so spacing matches v2.86.2 exactly.
- Restored `test/integration/SystemIntegrationTest.js`, byte-identical.
- Added both paths to the deprecated-class exclusion lists in the six `test/vitest-*.config.ts` coverage configs, following the convention #4325 established. This is cosmetic — see notes for reviewer.
- Dropped four dead `ignore` entries from `.github/codecov.yaml` — `AccountAllowanceAdjustTransaction`, `LiveHashAddTransaction`, `LiveHashDeleteTransaction` and `LiveHashQuery`, whose files #4269 deleted while leaving their ignores behind. All four are absent from the tree, so removing a glob that matches nothing is a no-op for Codecov. The `src/system/System*Transaction.js` entries are kept, since those files exist again and are still `@deprecated`.

#### Public API changes

- **Restores two exports** removed in v2.87.0: `SystemDeleteTransaction` and `SystemUndeleteTransaction`. Both stay `@deprecated`, so consumers using them get a lint warning rather than a hard break.
- **Restores registry behaviour:** `Transaction.fromBytes()` deserializes `systemDelete` / `systemUndelete` bodies again, rather than failing on an unknown data case.
- **Nothing else changes.** No other class, method, or emitted type is touched.

**Related issue(s)**:

Fixes #4342

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/.github/codecov.yaml
+++ b/.github/codecov.yaml
@@ -6,8 +6,6 @@ coverage:
                 threshold: 1%
 ignore:
     - "src/token/TokenNftsUpdateTransaction.js"
-    - "src/system/SystemUndeleteTransaction.js"
-    - "src/system/SystemDeleteTransaction.js"
     - "src/contract/StorageChange.js"
     - "src/contract/ContractStateChange.js"
     - "src/transaction/SignatureMapLegacy.js"
--- a/.github/codecov.yaml
+++ b/.github/codecov.yaml
@@ -6,10 +6,6 @@ coverage:
                 threshold: 1%
 ignore:
     - "src/token/TokenNftsUpdateTransaction.js"
-    - "src/account/AccountAllowanceAdjustTransaction.js"
-    - "src/account/LiveHashAddTransaction.js"
-    - "src/account/LiveHashDeleteTransaction.js"
-    - "src/account/LiveHashQuery.js"
     - "src/system/SystemUndeleteTransaction.js"
     - "src/system/SystemDeleteTransaction.js"
     - "src/contract/StorageChange.js"
--- a/src/system/SystemDeleteTransaction.js
+++ b/src/system/SystemDeleteTransaction.js
@@ -1,0 +1,239 @@
+// SPDX-License-Identifier: Apache-2.0
+
+import Transaction, {
+    TRANSACTION_REGISTRY,
+} from "../transaction/Transaction.js";
+import FileId from "../file/FileId.js";
+import ContractId from "../contract/ContractId.js";
+import Timestamp from "../Timestamp.js";
+
+/**
+ * @namespace proto
+ * @typedef {import("@hiero-ledger/proto").proto.ITransaction} HieroProto.proto.ITransaction
+ * @typedef {import("@hiero-ledger/proto").proto.ISignedTransaction} HieroProto.proto.ISignedTransaction
+ * @typedef {import("@hiero-ledger/proto").proto.TransactionBody} HieroProto.proto.TransactionBody
+ * @typedef {import("@hiero-ledger/proto").proto.ITransactionBody} HieroProto.proto.ITransactionBody
+ * @typedef {import("@hiero-ledger/proto").proto.ITransactionResponse} HieroProto.proto.ITransactionResponse
+ * @typedef {import("@hiero-ledger/proto").proto.ISystemDeleteTransactionBody} HieroProto.proto.ISystemDeleteTransactionBody
+ * @typedef {import("@hiero-ledger/proto").proto.IContractID} HieroProto.proto.IContractID
+ * @typedef {import("@hiero-ledger/proto").proto.IFileID} HieroProto.proto.IFileID
+ */
+
+/**
+ * @typedef {import("../channel/Channel.js").default} Channel
+ * @typedef {import("../account/AccountId.js").default} AccountId
+ * @typedef {import("../transaction/TransactionId.js").default} TransactionId
+ */
+
+/**
+ * Deprecated: Do not use.
+ * @deprecated
+ */
+export default class SystemDeleteTransaction extends Transaction {
+    /**
+     * @param {object} [props]
+     * @param {FileId | string} [props.fileId]
+     * @param {ContractId | string} [props.contractId]
+     * @param {Timestamp} [props.expirationTime]
+     */
+    constructor(props = {}) {
+        super();
+
+        /**
+         * @private
+         * @type {?FileId}
+         */
+        this._fileId = null;
+
+        /**
+         * @private
+         * @type {?ContractId}
+         */
+        this._contractId = null;
+
+        /**
+         * @private
+         * @type {?Timestamp}
+         */
+        this._expirationTime = null;
+
+        if (props.fileId != null) {
+            this.setFileId(props.fileId);
+        }
+
+        if (props.contractId != null) {
+            this.setContractId(props.contractId);
+        }
+
+        if (props.expirationTime != null) {
+            this.setExpirationTime(props.expirationTime);
+        }
+    }
+
+    /**
+     * @internal
+     * @param {HieroProto.proto.ITransaction[]} transactions
+     * @param {HieroProto.proto.ISignedTransaction[]} signedTransactions
+     * @param {TransactionId[]} transactionIds
+     * @param {AccountId[]} nodeIds
+     * @param {HieroProto.proto.ITransactionBody[]} bodies
+     * @returns {SystemDeleteTransaction}
+     */
+    static _fromProtobuf(
+        transactions,
+        signedTransactions,
+        transactionIds,
+        nodeIds,
+        bodies,
+    ) {
+        const body = bodies[0];
+        const systemDelete =
+            /** @type {HieroProto.proto.ISystemDeleteTransactionBody} */ (
+                body.systemDelete
+            );
+
+        return Transaction._fromProtobufTransactions(
+            // eslint-disable-next-line deprecation/deprecation
+            new SystemDeleteTransaction({
+                fileId:
+                    systemDelete.fileID != null
+                        ? FileId._fromProtobuf(
+                              /** @type {HieroProto.proto.IFileID} */ (
+                                  systemDelete.fileID
+                              ),
+                          )
+                        : undefined,
+                contractId:
+                    systemDelete.contractID != null
+                        ? ContractId._fromProtobuf(
+                              /** @type {HieroProto.proto.IContractID} */ (
+                                  systemDelete.contractID
+                              ),
+                          )
+                        : undefined,
+                expirationTime:
+                    systemDelete.expirationTime != null
+                        ? Timestamp._fromProtobuf(systemDelete.expirationTime)
+                        : undefined,
+            }),
+            transactions,
+            signedTransactions,
+            transactionIds,
+            nodeIds,
+            bodies,
+        );
+    }
+
+    /**
+     * @returns {?FileId}
+     */
+    get fileId() {
+        return this._fileId;
+    }
+
+    /**
+     * @param {FileId | string} fileId
+     * @returns {this}
+     */
+    setFileId(fileId) {
+        this._requireNotFrozen();
+        this._fileId =
+            fileId instanceof FileId ? fileId : FileId.fromString(fileId);
+
+        return this;
+    }
+
+    /**
+     * @returns {?ContractId}
+     */
+    get contractId() {
+        return this._contractId;
+    }
+
+    /**
+     * @param {ContractId | string} contractId
+     * @returns {this}
+     */
+    setContractId(contractId) {
+        this._requireNotFrozen();
+        this._contractId =
+            contractId instanceof ContractId
+                ? contractId
+                : ContractId.fromString(contractId);
+
+        return this;
+    }
+
+    /**
+     * @returns {?Timestamp}
+     */
+    get expirationTime() {
+        return this._expirationTime;
+    }
+
+    /**
+     * @param {Timestamp} expirationTime
+     * @returns {SystemDeleteTransaction}
+     */
+    setExpirationTime(expirationTime) {
+        this._requireNotFrozen();
+        this._expirationTime = expirationTime;
+        return this;
+    }
+
+    /**
+     * @override
+     * @internal
+     * @param {Channel} channel
+     * @param {HieroProto.proto.ITransaction} request
+     * @returns {Promise<HieroProto.proto.ITransactionResponse>}
+     */
+    _execute(channel, request) {
+        if (this._fileId != null) {
+            return channel.file.systemDelete(request);
+        } else {
+            return channel.smartContract.systemDelete(request);
+        }
+    }
+
+    /**
+     * @override
+     * @protected
+     * @returns {NonNullable<HieroProto.proto.TransactionBody["data"]>}
+     */
+    _getTransactionDataCase() {
+        return "systemDelete";
+    }
+
+    /**
+     * @override
+     * @protected
+     * @returns {HieroProto.proto.ISystemDeleteTransactionBody}
+     */
+    _makeTransactionData() {
+        return {
+            fileID: this._fileId != null ? this._fileId._toProtobuf() : null,
+            contractID:
+                this._contractId != null
+                    ? this._contractId._toProtobuf()
+                    : null,
+            expirationTime:
+                this._expirationTime != null
+                    ? this._expirationTime._toProtobuf()
+                    : null,
+        };
+    }
+
+    /**
+     * @returns {string}
+     */
+    _getLogId() {
+        const timestamp = /** @type {import("../Timestamp.js").default} */ (
+            this._transactionIds.current.validStart
+        );
+        return `SystemDeleteTransaction:${timestamp.toString()}`;
+    }
+}
+
+// eslint-disable-next-line @typescript-eslint/unbound-method, deprecation/deprecation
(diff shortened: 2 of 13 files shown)


ddeskov-limechain (dismissed):

src/system/SystemDeleteTransaction.js
```diff
@@ -0,0 +1,243 @@
...
+     */
+    setContractId(contractId) {
+        this._requireNotFrozen();
+        this._contractId =
+            contractId instanceof ContractId
+                ? contractId
+                : ContractId.fromString(contractId);
+
+        return this;
+    }
```
Fantomasa: The sibling setters defensively copy: `typeof x === "string" ? X.fromString(x) : x.clone()` (see `FileDeleteTransaction.setFileId`, `ContractDeleteTransaction.setContractId`). Storing the caller's instance by reference aliases its mutable checksum state across transactions, and an id object from a duplicated/CJS copy of the SDK fails `instanceof` and lands in `fromString(object)`. Applies to all four id setters in both classes. With the sibling clearing from the previous comment folded in:

```suggestion
    setContractId(contractId) {
        this._requireNotFrozen();
        this._fileId = null;
        this._contractId =
            typeof contractId === "string"
                ? ContractId.fromString(contractId)
                : contractId.clone();

        return this;
    }
```

src/system/SystemDeleteTransaction.js
```diff
@@ -0,0 +1,243 @@
...
+    /**
+     * @returns {?ContractId}
+     */
+    get contractId() {
+        return this._contractId;
+    }
+
+    /**
+     * Targets a smart contract, submitting to `SmartContractService`, whose
+     * `systemDelete` RPC is deprecated in the protobufs. Prefer `setFileId`.
```
Fantomasa: The protobufs say more than deprecated: `services_system_delete.proto` documents the `contractID` option as unsupported, never implemented, and specified to fail with `INVALID_FILE_ID` or `MISSING_ENTITY_ID`, so "prefer `setFileId`" reads too soft. Related: with neither id set, `_execute` silently defaults to this dead `smartContract.systemDelete` route; a client-side error for the missing id would fail faster. Same doc text in `SystemUndeleteTransaction.js` lines 139-140.

```suggestion
     * Targets a smart contract, submitting to `SmartContractService`, whose
     * `systemDelete` RPC was never implemented on the network; the protobufs
     * document it as always failing with `INVALID_FILE_ID` or
     * `MISSING_ENTITY_ID`. Only the file form (`setFileId`) is functional.
```

src/system/SystemDeleteTransaction.js
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    /**
+     * @override
+     * @internal
+     * @param {Channel} channel
+     * @param {HieroProto.proto.ITransaction} request
+     * @returns {Promise<HieroProto.proto.ITransactionResponse>}
+     */
+    _execute(channel, request) {
```
Fantomasa: Neither restored class overrides `_validateChecksums(client)`, so with `client.setAutoValidateChecksums(true)` a wrong-network checksum in `setFileId("0.0.150-vfmkw")` is silently accepted: the base implementation is a no-op. `FileDeleteTransaction.js:119` and `ContractDeleteTransaction.js` are the precedent (add the `Client` typedef import alongside `Channel`):

```js
    /**
     * @param {Client} client
     */
    _validateChecksums(client) {
        if (this._fileId != null) {
            this._fileId.validateChecksum(client);
        }
        if (this._contractId != null) {
            this._contractId.validateChecksum(client);
        }
    }
```

src/system/SystemDeleteTransaction.js
```diff
@@ -0,0 +1,243 @@
...
+        return {
+            fileID: this._fileId != null ? this._fileId._toProtobuf() : null,
+            contractID:
+                this._contractId != null
+                    ? this._contractId._toProtobuf()
+                    : null,
+            expirationTime:
+                this._expirationTime != null
+                    ? this._expirationTime._toProtobuf()
+                    : null,
```
Fantomasa: `expirationTime` in `SystemDeleteTransactionBody` is a `TimestampSeconds`, so the `nanos` half of `Timestamp._toProtobuf()` is silently dropped on encode: `fromBytes(toBytes())` comes back with `nanos = 0` and round-trip equality fails. Worth documenting second-precision on `setExpirationTime` (or truncating there) so the loss is visible at the API boundary.

src/system/SystemUndeleteTransaction.js
```diff
@@ -0,0 +1,209 @@
...
+ * @typedef {import("../transaction/TransactionId.js").default} TransactionId
+ */
+
+/**
+ * Restore a file or smart contract removed by `SystemDeleteTransaction`.
+ * Privileged: by default only accounts 2-60 may submit this, so an ordinary
+ * payer is rejected. Content removed by `FileDeleteTransaction` is not
+ * retained and cannot be restored this way.
+ */
+export default class SystemUndeleteTransaction extends Transaction {
```
Fantomasa: About 110 of these 209 lines are a character-for-character copy of `SystemDeleteTransaction` (fields, all four id accessors, `_execute` routing, `_getLogId`), and the pair had already drifted once before removal (the old Undelete documented a `props.expirationTime` that never existed). Every fix from this review lands twice. Consider a small shared base like `src/token/AbstractTokenTransferTransaction.js`; fine as a follow-up if you prefer to keep the restore verbatim.

test/integration/SystemIntegrationTest.js
```diff
@@ -0,0 +1,77 @@
...
+import IntegrationTestEnv from "./client/NodeIntegrationTestEnv.js";
+
+describe("SystemIntegration", function () {
+    let env;
+
+    beforeAll(async function () {
+        env = await IntegrationTestEnv.new();
+    });
+
+    it("should be executable when file id is not set", async function () {
```
Fantomasa: All four tests are named "should be executable..." but assert that an error was thrown, and the bare `catch (_)` passes for any error at all, a `TypeError` from a broken export included. The suite idiom asserts the specific status, e.g. `FileDeleteIntegrationTest.js:62` checks `error.toString().includes(Status.InvalidFileId.toString())`; the expected privileged-payer rejections here are just as concrete and assertable. The names also point the wrong way ("when file id is not set" actually sets a contract id), and line 46 has the typo "expiratiion".

.github/codecov.yaml
```diff
@@ -6,12 +6,6 @@ coverage:
                 threshold: 1%
 ignore:
     - "src/token/TokenNftsUpdateTransaction.js"
-    - "src/account/AccountAllowanceAdjustTransaction.js"
-    - "src/account/LiveHashAddTransaction.js"
-    - "src/account/LiveHashDeleteTransaction.js"
-    - "src/account/LiveHashQuery.js"
-    - "src/system/SystemUndeleteTransaction.js"
-    - "src/system/SystemDeleteTransaction.js"
```
Fantomasa: Dropping these ignores without adding unit tests is what makes `codecov/patch` fail on this PR (38.98%): `SystemUndeleteTransaction` sits at 3.84% because nothing anywhere instantiates it, and `SystemDeleteTransaction._fromProtobuf` is never exercised. A `toBytes`/`fromBytes` round-trip unit test per class (see `test/unit/NodeDeleteTransaction.js` for the shape) covers `_fromProtobuf`, `_makeTransactionData`, and the setters; alternatively keep the ignores until tests exist.

src/system/SystemDeleteTransaction.js
```diff
@@ -0,0 +1,243 @@
...
+     * @param {FileId | string} fileId
+     * @returns {this}
+     */
+    setFileId(fileId) {
+        this._requireNotFrozen();
+        this._fileId =
+            fileId instanceof FileId ? fileId : FileId.fromString(fileId);
+
+        return this;
+    }
```
Fantomasa, line 150: `fileID`/`contractID` are a protobuf `oneof`, but the encoder writes both fields unconditionally: with both set, the node decodes a contract delete while `_execute` routes to `FileService`. Clear the sibling like `ContractDeleteTransaction.setTransferContractId` does; same for `setContractId` below and both setters in `SystemUndeleteTransaction.js`.

```suggestion
    setFileId(fileId) {
        this._requireNotFrozen();
        this._contractId = null;
        this._fileId =
            fileId instanceof FileId ? fileId : FileId.fromString(fileId);

        return this;
    }
```

src/system/SystemDeleteTransaction.js
Fantomasa, line 176: Sibling setters defensively copy (`typeof x === "string" ? X.fromString(x) : x.clone()`, see `FileDeleteTransaction.setFileId`); storing the caller's instance by reference aliases its mutable checksum state. Applies to all four id setters in both classes; sibling clearing from the previous comment folded in:

```suggestion
    setContractId(contractId) {
        this._requireNotFrozen();
        this._fileId = null;
        this._contractId =
            typeof contractId === "string"
                ? ContractId.fromString(contractId)
                : contractId.clone();

        return this;
    }
```

src/system/SystemDeleteTransaction.js
Fantomasa: The protobufs document the `contractID` option as never implemented and always failing with `INVALID_FILE_ID`/`MISSING_ENTITY_ID`, so "prefer `setFileId`" reads too soft. Same doc text in `SystemUndeleteTransaction.js:139-140`.

```suggestion
     * Targets a smart contract, submitting to `SmartContractService`, whose
     * `systemDelete` RPC was never implemented on the network; the protobufs
     * document it as always failing with `INVALID_FILE_ID` or
     * `MISSING_ENTITY_ID`. Only the file form (`setFileId`) is functional.
```

src/system/SystemDeleteTransaction.js
```diff
@@ -0,0 +1,243 @@
...
+
+    /**
+     * @param {Timestamp} expirationTime
+     * @returns {SystemDeleteTransaction}
+     */
+    setExpirationTime(expirationTime) {
+        this._requireNotFrozen();
+        this._expirationTime = expirationTime;
+        return this;
+    }
```
Fantomasa, line 200: Sibling expiration setters accept `Timestamp | Date` (`FileCreateTransaction.setExpirationTime`); a `Date` passed here only blows up later during freeze/execute. Also the other setters use `@returns {this}`.

```suggestion
    /**
     * @param {Timestamp | Date} expirationTime
     * @returns {this}
     */
    setExpirationTime(expirationTime) {
        this._requireNotFrozen();
        this._expirationTime =
            expirationTime instanceof Date
                ? Timestamp.fromDate(expirationTime)
                : expirationTime;
        return this;
    }
```

src/system/SystemDeleteTransaction.js
Fantomasa, line 222: Neither restored class overrides `_validateChecksums`, so `setAutoValidateChecksums(true)` silently skips these ids; see `FileDeleteTransaction.js:119` (add the `Client` typedef import):

```js
    /**
     * @param {Client} client
     */
    _validateChecksums(client) {
        if (this._fileId != null) {
            this._fileId.validateChecksum(client);
        }
        if (this._contractId != null) {
            this._contractId.validateChecksum(client);
        }
    }
```

src/system/SystemDeleteTransaction.js
Fantomasa, line 254: The proto field is `TimestampSeconds`, so the `nanos` from `Timestamp._toProtobuf()` are dropped on encode and `fromBytes(toBytes())` returns `nanos = 0`. Worth documenting second-precision on `setExpirationTime` or truncating there.

src/system/SystemUndeleteTransaction.js
Fantomasa, line 35: ~110 of these 209 lines duplicate `SystemDeleteTransaction`, and every fix in this review lands twice. Consider a shared base like `src/token/AbstractTokenTransferTransaction.js`, fine as a follow-up.

test/integration/SystemIntegrationTest.js
Fantomasa: The names say "should be executable" but the tests assert an error, and `catch (_)` passes for any error at all. Assert the specific status like `FileDeleteIntegrationTest.js:62` does; also "when file id is not set" actually sets a contract id, and line 46 has the typo "expiratiion".

test/integration/SystemIntegrationTest.js
```diff
@@ -0,0 +1,77 @@
...
+        let errorThrown = false;
+
+        try {
+            await new SystemDeleteTransaction()
+                .setContractId(new ContractId(10))
+                .setExpirationTime(Timestamp.generate())
```
Fantomasa: `Timestamp.generate()` is always 3-8s in the past, so tests 1/2 duplicate 3/4 and all four fail the same precheck. With the CI genesis operator `0.0.2` (`GENESIS_OPERATOR_ID` in build.yml) a real delete/undelete round trip is possible, which would be the first test to instantiate `SystemUndeleteTransaction`.

.github/codecov.yaml
Fantomasa, line 14: Removing these ignores with no unit tests added is why `codecov/patch` fails (38.98%): nothing instantiates `SystemUndeleteTransaction`. Add `toBytes`/`fromBytes` round-trip tests per class (see `test/unit/NodeDeleteTransaction.js`) or keep the ignores until tests exist.

Fantomasa (changes requested):

src/system/SystemUndeleteTransaction.js
Dosik13: I don't think we need this because there won't be much changes in this transactions anyways and it is overcomplication for nothing

Fantomasa (approved):
LGTM, Thank you !

andrewb1269 (approved):

PR #4349, opened 2026-09-03
fix: copy maxBackoff when cloning ManagedNode

### LINKED ISSUE #4347: ManagedNode sets the wrong field
js/src/ManagedNode.js:92: 

`this._maxBackoff = props.cloneNode.node._minBackoff;`

Steve0x2a:
## Summary
- Fix ManagedNode clone path copying `_maxBackoff` from `_minBackoff` instead of `_maxBackoff`, which collapsed the backoff ceiling to the floor for cloned nodes.
- Add a unit test that proves clone preserves distinct min/max backoff defaults (8000 / 3600000).

Credit: @mustafauzunn for reporting.

Fixes #4347

## Test plan
- [x] `npx vitest --config=test/vitest-node.config.ts run test/unit/ManagedNode.js` (RED on unmodified main, GREEN after fix)

### DIFF
--- a/src/ManagedNode.js
+++ b/src/ManagedNode.js
@@ -89,7 +89,7 @@ export default class ManagedNode {
             this._minBackoff = props.cloneNode.node._minBackoff;
 
             /** @type {number} */
-            this._maxBackoff = props.cloneNode.node._minBackoff;
+            this._maxBackoff = props.cloneNode.node._maxBackoff;
         } else {
             throw new Error(
                 `failed to create ManagedNode: ${JSON.stringify(props)}`,
--- a/test/unit/ManagedNode.js
+++ b/test/unit/ManagedNode.js
@@ -1,0 +1,30 @@
+import MirrorNode from "../../src/MirrorNode.js";
+
+describe("ManagedNode", function () {
+    it("preserves distinct min and max backoff when cloning", function () {
+        const minBackoff = 8000;
+        const maxBackoff = 1000 * 60 * 60;
+
+        const node = new MirrorNode({
+            newNode: {
+                address: "testnet.mirrornode.hedera.com:443",
+                channelInitFunction: () => ({}),
+            },
+        });
+
+        expect(node.minBackoff).to.equal(minBackoff);
+        expect(node.maxBackoff).to.equal(maxBackoff);
+
+        // Clone via ManagedNode cloneNode constructor path
+        const cloned = new MirrorNode({
+            cloneNode: {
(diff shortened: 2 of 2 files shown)


src/ManagedNode.js
```diff
@@ -89,7 +89,7 @@ export default class ManagedNode {
             this._minBackoff = props.cloneNode.node._minBackoff;
 
             /** @type {number} */
-            this._maxBackoff = props.cloneNode.node._minBackoff;
+            this._maxBackoff = props.cloneNode.node._maxBackoff;
```
Fantomasa, line 92: Good catch, and thanks for the fix! I verified it locally: the new unit test fails on current `main` and passes with this change, so the regression coverage is real.

test/unit/ManagedNode.js
```diff
@@ -0,0 +1,30 @@
...
+                address: "testnet.mirrornode.hedera.com:443",
+                channelInitFunction: () => ({}),
+            },
+        });
+
+        expect(node.minBackoff).to.equal(minBackoff);
+        expect(node.maxBackoff).to.equal(maxBackoff);
+
+        // Clone via ManagedNode cloneNode constructor path
+        const cloned = new MirrorNode({
```
Fantomasa, line 19: Nice test, this covers the shared constructor path directly. Optional idea: the way this clone path is actually reached in production is `Node.toSecure()` / `toInsecure()` (via `Client.setTransportSecurity`), so a small extra case through that public API would also guard the subclass copy logic:

```js
import Node from "../../src/Node.js";
import AccountId from "../../src/account/AccountId.js";

it("preserves backoff values when cloning via toSecure()", function () {
    const node = new Node({
        newNode: {
            accountId: new AccountId(3),
            address: "0.testnet.hedera.com:50211",
            channelInitFunction: () => ({}),
        },
    });

    const secure = node.toSecure();

    expect(secure.minBackoff).to.equal(node.minBackoff);
    expect(secure.maxBackoff).to.equal(node.maxBackoff);
});
```

I ran this snippet locally and it passes with your fix. Feel free to skip it if you would rather keep the PR minimal.

Fantomasa (dismissed):

ddeskov-limechain:
@Steve0x2a are you still working on this issue? We may be forced to close it and implement the changes from our side if you do not respond soon.

Fantomasa (approved):
LGTM

ddeskov-limechain (approved):
LGTM

PR #4358, opened 2026-09-09
chore(release): v2.88.0

Fantomasa:
## Stable release: v2.88.0

**Versions**

- `@hiero-ledger/sdk`: 2.87.0 -> 2.88.0
- `@hiero-ledger/cryptography`: unchanged (1.21.0, only a test file changed since v2.87.0)
- `@hiero-ledger/proto`: unchanged (2.31.0)

Also in this commit: refreshed mainnet/testnet address books (`task update:addressbooks`) and a type fix in `examples/account/get-balance-mirror-node.js` that `task examples:build` required (`error` is `unknown` in a `catch`, so it now narrows with `instanceof MirrorNodeStatusError`).

Not included: PR #4306 (`AccountBalanceQuery` deprecation) is still open and will go into a later release.

### Changed
- **Breaking:** `MirrorNodeAccountBalanceQuery` now throws the new `MirrorNodeStatusError` carrying `Status.InvalidAccountId` for an account the mirror node does not know, instead of returning `0 ℏ` as in v2.87.0. This matches the `INVALID_ACCOUNT_ID` failure of the deprecated `AccountBalanceQuery` it replaces, and the Java and Go SDKs. Match on `error.status`, not on the error class: mirror REST queries reach no consensus node, so this is deliberately not a `PrecheckStatusError`. A just-created account fails transiently until the mirror node ingests it, so create-then-read flows should retry; an account that exists holding nothing still returns zero, and a deleted account is reported as an ordinary zero balance because the balances endpoint does not expose the deleted flag. A malformed response (missing `balances` array or non-numeric balance) now throws a plain `Error` instead of returning zero. [#4335](https://github.com/hiero-ledger/hiero-sdk-js/pull/4335)

### Fixed
- Restored `SystemDeleteTransaction` and `SystemUndeleteTransaction`, which v2.87.0 removed as long-deprecated. Deleting them also unregistered `systemDelete` / `systemUndelete` from the transaction registry, so `Transaction.fromBytes()` could no longer deserialize an already-signed system delete/undelete transaction. Both classes are back exactly as in v2.86.2 and remain `@deprecated`; every other removal from v2.87.0 stays removed. [#4343](https://github.com/hiero-ledger/hiero-sdk-js/pull/4343)
- `Transaction.fromBytes()` now rejects a transaction body that sets more than one `data` field. protobufjs silently kept only the last such field while decoding, so an ambiguous body could not be detected after the fact. This is a defense-in-depth fix for already-trusted input; there is no known untrusted-input path. [#4338](https://github.com/hiero-ledger/hiero-sdk-js/pull/4338) [#4337](https://github.com/hiero-ledger/hiero-sdk-js/issues/4337)


ddeskov-limechain (approved):
LGTM

Issue #4357, opened 2026-09-08
Examples CI job is unreliable: bulk consensus timeouts mid-run and pub-sub-chunked flake


Fantomasa:
## Problem

The `Run examples using Node 22` job (`build.yml`) fails on most runs across unrelated branches, including `main`. A green examples run is currently not achievable on any branch, so the job's red status has become noise that reviewers wave through, which is exactly how a real example regression will slip in.

Two distinct failure modes are showing up. Neither correlates with the changes on the branches where they fire.

### 1. Consensus node stops answering mid-run, examples fail in bulk with `timeout exceeded`

[Run 33852023118](https://github.com/hiero-ledger/hiero-sdk-js/actions/runs/33852023118) (a dependabot branch touching only a `tck/` dev dependency): 36 of 93 examples failed, with 53 errors of one identical shape:

```
t [Error]: timeout exceeded
    at l.execute (.../lib/Executable.js:1:6115)
    at async main (...) {
  nodeAccountId: '0.0.3'
}
```

That is `MaxAttemptsOrTimeoutError`: the SDK exhausted its `requestTimeout` budget against local node `0.0.3`.

Observations:

- The failures cluster in a window of the run (`file/create.js`, `signing/*`, `schedule/*`, `token/*`, `contract/*`, ...) while examples before and after complete fine, so the CI local network goes unhealthy for a stretch of the job rather than individual examples being broken.
- Severity varies run to run with the same signature: [run 32134478170](https://github.com/hiero-ledger/hiero-sdk-js/actions/runs/32134478170) (Aug 18) failed ~20 examples; [run 33704744186](https://github.com/hiero-ledger/hiero-sdk-js/actions/runs/33704744186) failed 6 of 93 with 8 timeouts.

### 2. `consensus/pub-sub-chunked.js` mirror-subscription flake

[Run 33723885831](https://github.com/hiero-ledger/hiero-sdk-js/actions/runs/33723885831) on `main` (Sep 3, `5b785ed5`): 92 of 93 examples passed and only `consensus/pub-sub-chunked.js` failed:

```
Error: Large topic message was not received! (Fail)
    at examples/consensus/pub-sub-chunked.js:135
```

[...]

PR #4368, opened 2026-09-15
fix(tck): return deletedAt/executedAt in ScheduleInfoQuery response

### LINKED ISSUE #4367: TCK server: ScheduleInfoQuery response uses `deleted`/`executed` instead of spec's `deletedAt`/`executedAt`
## Description

The TCK server's `getScheduleInfo` response maps the deletion and execution timestamps as `deleted` and `executed` (`tck/utils/helpers/schedule.ts`, `tck/response/schedule.ts`). The TCK spec's output table defines them as `deletedAt` and `executedAt`:

https://hiero-ledger.github.io/hiero-sdk-tck/test-specifications/schedule-service/ScheduleInfoQuery.html

The Go TCK server already returns `deletedAt` / `executedAt`. The Java server has the same bug, tracked in hiero-ledger/hiero-sdk-java#2950.

This was masked so far because the TCK tests that read these fields only asserted `undefined` or were guarded by an `if`. With hiero-ledger/hiero-sdk-tck#711 (fixed in hiero-ledger/hiero-sdk-tck#712) test #4 asserts a non-empty `deletedAt` and will fail against the JS server.

## Steps to reproduce

Run the TCK `ScheduleInfoQuery` suite against the JS TCK server after hiero-ledger/hiero-sdk-tck#712; test #4 "Query for the info of a deleted schedule" fails on the missing `deletedAt` property.

## Solution

Rename the two response keys to `executedAt` / `deletedAt` in the response interface and the mapper.

ddeskov-limechain:
## Description

The TCK server's `getScheduleInfo` response returned `deleted` / `executed`, while the TCK spec's output table defines `deletedAt` / `executedAt` (the Go TCK server already uses the spec names; the Java server is fixed in hiero-ledger/hiero-sdk-java#2951). Renames the two keys in the response interface and the mapper.

Needed by hiero-ledger/hiero-sdk-tck#712, which makes TCK test #4 assert a non-empty `deletedAt`.

## Related issue(s)

Fixes #4367

## Notes for reviewer

Two-line rename in `tck/response/schedule.ts` and `tck/utils/helpers/schedule.ts`; no SDK code touched.

## Checklist

- [x] Documented (n/a, spec already defines the names)
- [x] Tested (prettier passes; verified against the TCK spec output table)


ddeskov-limechain:
Both red jobs are unrelated to this change (it touches only `tck/`):

- **Run examples**: `Large topic message was not received`, the known examples-job instability in #4357 (failing on every recent run of this workflow).
- **Test using Node 22**: `ClientPingMocking.js` failed with `EADDRINUSE` on the mock gRPC port. The same tests failed the same way on the `deprecate-balance-query` branch 20 minutes later; filed as https://github.com/hiero-ledger/hiero-sdk-js/issues/4370.

Re-ran the failed jobs.

Fantomasa (approved):
LGTM

PR #4369, opened 2026-09-15
fix: make balance examples mirror-ingestion safe

### LINKED ISSUE #4365: Fix SDK examples for AccountBalanceQuery deprecation stages 1 and 2
## Description

Examples that execute `AccountBalanceQuery` will fail once Stage 2 makes execution reject locally. Migrating them to mirror-node reads also changes timing and response shape: the mirror may not have ingested a newly created account or the latest transaction, and the replacement account-balance query returns HBAR only.

This is the example migration and validation subtask of #4285. Coordinate with #4306, which already contains the main migrations; complete that work there or in a linked follow-up rather than implementing it twice.

### Stage 1: connectivity compatibility

Stage 1 (#4305, completed) replaces the internal `Client.ping()` / `pingAll()` balance probe with `getAccountInfo(0.0.2)` using `COST_ANSWER`, preserving the public API.

- Audit example setup, connectivity guidance, and any indirect provider/client paths for assumptions that a balance query is a readiness probe. Use the public ping APIs where a connectivity probe is needed.
- Verify applicable examples remain usable with the Stage 1 SDK. A successful ping proves connectivity, not mirror ingestion or balance freshness.
- This issue does not reimplement ping or address the separate unknown-node bug #4330.

### Stage 2: balance migration and correctness

- Replace remaining example execution of `AccountBalanceQuery`, including provider/signer wrappers. Use `MirrorNodeAccountBalanceQuery` or a compatible migrated provider for HBAR; use `MirrorNodeTokenBalanceQuery` or the appropriate mirror REST endpoint for token balances.
- Preserve each example's purpose. Account ID resolution, token reads, and HBAR reads may need different replacements. Mirror HTTP queries must not be passed through a path that assumes serialized gRPC queries or `executeWithSigner()` support.
- For reads immediately after account creation or a transaction, wait for the expected mirror state with a bounded deadline. Handle temporary missing-account responses for accounts the example just created. Surface malformed-input and other genuine errors; do not retry every error indiscriminately.
- Keep displayed/asserted token and HBAR values correct. Do not treat an HBAR-only response or empty token maps as proof of zero token holdings.
- Update comments and accompanying instructions. If removing an example capability is intentional, obtain agreement and document it instead of silently replacing it with an unconditional error.

## Affected examples

The September 14 audit of main at `a1cbd3e5a09f8a282c8d48a236e2bc590da9f1fe` found **80 direct constructor calls in 20 example files**. All paths below are relative to `examples/`.

| Area | Files |
|---|---|
| Account | `account/alias.js`, `account/allowance.js`, `account/create-with-thresholdkey.js`, `account/get-balance.js`, `account/multi-app-transfer.js`, `account/transfer-crypto.js` |
| Scheduling | `schedule/long-term-transaction.js`, `schedule/multi-sig-threshold.js`, `schedule/transfer.js` |
| Tokens | `token/airdrop.js`, `token/reject.js`, `token/nft-1-create-mint-burn-associate-transfer.js`, `token/nft-2-kyc-schedule.js`, `token/nft-3-pause-freeze-wipe-delete.js` |
| Other scripts | `batch-tx.js`, `consensus/create-with-revenue.js`, `network/async-testnet-client.js` |
| React Native | `react-native-example/services/hiero-service.ts`, `react-native-example-legacy/App.tsx` |
| REST signer | `simple_rest_signature_provider/src/client.js` |

At reviewed #4306 head `0691dd5af049d3ff94062a6771d767f719a47922`, all 80 direct calls are removed/replaced. That establishes migration coverage, not successful execution. Also validate `token/hip-542-auto-account-create.js`, which #4306 changes to a mirror token-balance read although it is not in the direct-call count.

## Remaining gaps observed in #4306

1. `account/allowance.js` reads balances immediately after account creation without waiting for mirror ingestion. A temporary `INVALID_ACCOUNT_ID` aborts the flow.
2. The added `untilMirror` helpers, for example in `account/transfer-crypto.js`, only repeat when the callback returns `null`; a temporary missing-account exception escapes on the first attempt. An offline reproduction confirmed one read and immediate failure even when the next simulated read would succeed.
3. Some post-transaction callers, including those in `token/airdrop.js`, omit the helper's previous/expected value, allowing a stale balance to be returned immediately.
4. `simple_rest_signature_provider/src/client.js` removes the balance demonstration and changes both balance methods to reject. Decide whether to migrate that capability or explicitly document an agreed removal.
5. The standard example runner excludes `batch-tx.js`, both React Native directories, and `simple_rest_signature_provider`; its result cannot validate those examples.

## Steps to reproduce

1. Use an isolated SDK build with Stage 2 enabled and a working local Solo consensus/mirror deployment.
2. Run an unmigrated example that executes `AccountBalanceQuery`, such as `account/transfer-crypto.js` from the audited main revision. The query rejects instead of reading a balance.
3. For migrated examples in #4306, exercise a newly created account before mirror ingestion, or simulate a temporary missing-account response followed by success. Verify the example waits and completes rather than aborting on the first response.
4. Exercise immediate post-transfer/token-operation reads and confirm the reported values reflect the completed operation rather than stale mirror state.

Observed validation so far: static call-site audit and offline reproductions of the ingestion-lag handling gaps. Full live execution of this example matrix remains required.

## Acceptance criteria

- [ ] Stage 1 compatibility audit is recorded, including any applicable readiness/setup paths; no example relies on the deprecated balance query as a connectivity probe.
- [ ] No runnable example or indirect example helper executes `AccountBalanceQuery`; explanatory deprecation references may remain.
- [ ] HBAR, token, alias/identity, and signer/provider examples retain their intended behavior, with any agreed capability removal documented.
- [ ] Post-creation and post-transaction reads handle expected ingestion lag within bounded time and return the expected values. Focused regression checks cover temporary missing-account followed by success, timeout, and genuine-error propagation.
- [ ] The affected standard examples pass against the Stage 2 SDK build on a healthy local network. Record SDK commit/version and network versions, and distinguish environment failures from example failures.
- [ ] Batch, both React Native variants, and the REST-signer example have explicit validation evidence from their supported environments, since they are excluded from the standard runner. Record any explicitly agreed deferral rather than implying CI covered them.
- [ ] Link the implementing PR(s), including #4306, and complete the example work before shipping the corresponding Stage 2 release.

## References

- Stage 1: #4305
- Stage 2 parent issue: #4285
- Existing migration implementation: #4306
- [AccountBalanceQuery deprecation proposal](https://github.com/hiero-ledger/sdk-collaboration-hub/blob/main/proposals/account-balance-query-deprecation.md)
- [Mirror-node replacement proposal](https://github.com/hiero-ledger/sdk-collaboration-hub/blob/main/proposals/account-balance-query-mirror-node-migration.md)

AgrimTawani:
## Summary

- replace removed `AccountBalanceQuery` usage with bounded mirror-node polling that shares one hard timeout budget
- use exact expected balances where the account is stable and lower bounds for node account `0.0.3`
- validate malformed mirror token responses and forward the optional provider `requestTimeout`
- run examples serially by default because they share one operator account and concurrent fees make balance assertions nondeterministic

## Maintainer review updates

- rebased onto current `main`; this PR now contains only the three issue commits
- fixed the revenue example to expect the actual 1 HBAR fee and reordered its transactions so later exact reads prove mirror ingestion
- restored the public `Long` value type and froze the timeout-test clock
- documented malformed token responses and provider timeouts in the changelog
- opened #4373 for the generated `Long` declaration incompatibility

The modern React Native typing adjustments and REST signer ESLint plugin change are included because those examples otherwise fail their required static checks.

Closes #4365

## Verification

Local verification on SDK commit `28a64cfe`:

- Node unit suite: 164 files, 2,082 passed, 1 skipped
- Browser unit suite: 140 files, 1,835 passed
- root, examples, REST signer, modern React Native, and legacy React Native TypeScript checks passed
- full SDK build, relevant ESLint, formatting, dependency-cycle check, workflow YAML parse, and `git diff --check` passed
- REST signer ESLint has 0 errors and 5 pre-existing warnings

The examples CI job now records execution of `batch-tx.js` and the REST signer client/server flow, plus both React Native TypeScript checks, against Node 22, Solo 0.69.1, consensus node 0.73.0, and mirror node 0.153.0. React Native device runtime was not available locally; both applications were statically verified and the CI checks are included for reproducible evidence.


AgrimTawani:
@ddeskov-limechain, the implementation for #4365 is ready for CI review. Could you please assign this PR and approve the fork workflow runs? GitHub blocks me from self-assigning, so the Assignee Check is currently failing and the Build & Test/Common JS workflows are waiting for maintainer approval. Local Node/browser unit suites, TypeScript, formatting, source lint, and dependency-cycle checks pass.

AgrimTawani:
Personal review found and fixed:

- Polling/provider retries could exceed their deadline; reads and backoff now share a hard timeout budget.
- Change-only checks could accept unrelated concurrent updates; critical balances now use exact expected state, the runner defaults to serial execution, and React Native waits for mirror ingestion.
- Several examples swallowed failures, leaked client timers, or logged temporary private keys; errors now fail the run, clients are closed, and key logging is removed.
- Added deadline, provider-timeout, malformed-response, and platform type/lint coverage.

Local Node/browser suites, examples, REST signer, and both React Native type-checks pass.

Fantomasa:
@AgrimTawani Please fix your merge conflicts

examples/consensus/create-with-revenue.js
```diff
@@ -123,20 +131,29 @@ async function main() {
...
+            aliceBalanceBefore,
+        );
 
-        let feeCollectorBalanceAfter = await new AccountBalanceQuery()
-            .setAccountId(operatorId)
-            .execute(client);
+        const expectedFeeCollectorBalance = Hbar.fromTinybars(
+            feeCollectorBalanceBefore
+                .toTinybars()
+                .add(Hbar.from(2, HbarUnit.Hbar).toTinybars()),
```
ddeskov-limechain: The topic fee is 1 Hbar (line 76). 2 Hbar is only the limit Alice allows. The collector gets +1, so this poll never matches and the example times out.

```suggestion
                .add(new Hbar(1).toTinybars()),
```

examples/consensus/create-with-revenue.js
```diff
@@ -86,13 +91,16 @@ async function main() {
...
+            client,
+            aliceAccountId,
+            undefined,
+            true,
+        );
 
-        let feeCollectorBalanceBefore = await new AccountBalanceQuery()
-            .setAccountId(operatorId)
-            .execute(client);
+        let feeCollectorBalanceBefore = await hbarBalance(client, operatorId);
```
ddeskov-limechain, line 105: This read can be stale. The operator paid the topic create just before, and the mirror may not show that fee yet. Then `before + 1 Hbar` is never seen and the poll times out.

Fix: move Step 2 (topic create, lines 68-87) above Step 1 (Alice create, lines 50-66). Alice's read above then proves the mirror also has the topic create, so this read is current.

```js
// Step 1: create the topic with an hbar custom fee (operator pays)
const customFee = new CustomFixedFee()
    .setAmount(new Hbar(1).toTinybars())
    .setFeeCollectorAccountId(operatorId);
const { topicId } = await (
    await new TopicCreateTransaction()
        .setAdminKey(operatorKey)
        .setFeeScheduleKey(operatorKey)
        .setCustomFees([customFee])
        .execute(client)
).getReceipt(client);

// Step 2: create alice (operator pays)
const { accountId: aliceAccountId } = await (
    await new AccountCreateTransaction()
        .setKeyWithoutAlias(aliceKey)
        .setInitialBalance(new Hbar(5))
        .setMaxAutomaticTokenAssociations(100)
        .execute(client)
).getReceipt(client);

// Alice shows up on the mirror only after her create is ingested, so the
// topic create is ingested too and the operator read below is current.
let aliceBalanceBefore = await hbarBalance(client, aliceAccountId, undefined, true);
let feeCollectorBalanceBefore = await hbarBalance(client, operatorId);
```

examples/consensus/create-with-revenue.js
```diff
@@ -188,13 +205,31 @@ async function main() {
...
+            aliceAccountId,
+            tokenId,
+            1,
+            true,
+        );
 
-        feeCollectorBalanceBefore = await new AccountBalanceQuery()
-            .setAccountId(operatorId)
-            .execute(client);
+        const feeCollectorHbarBefore = await hbarBalance(client, operatorId);
```
ddeskov-limechain, line 232: Same issue. The topic update (Step 6) is paid by the operator right before this read, so this read can miss that fee. Then the `equals` check in Step 8 throws.

Fix: move the token transfer to Alice (lines 175-183) below the topic update (Step 6). `aliceTokenBefore` (exact 1) then proves the update is ingested and this read is current.

```js
// Step 6: update the topic to charge the token (operator pays)
await (
    await new TopicUpdateTransaction()
        .setTopicId(topicId)
        .setCustomFees([customFeeToken])
        .execute(client)
).getReceipt(client);

// Then transfer the token to alice (operator pays). The exact
// aliceTokenBefore read below proves both transactions are ingested.
await (
    await new TransferTransaction()
        .addTokenTransfer(tokenId, client.operatorAccountId, -1)
        .addTokenTransfer(tokenId, aliceAccountId, 1)
        .execute(client)
).getReceipt(client);
```

examples/account/transfer-crypto.js
```diff
@@ -95,3 +104,39 @@ void main()
...
+                .setAccountId(accountId)
+                .execute(client, remainingMs);
+
+            if (expected == null) {
+                return hbars;
+            }
+
+            return hbars.toTinybars().equals(expected.toTinybars())
+                ? hbars
+                : null;
```
ddeskov-limechain: 0.0.3 is a node account. It also gets the node fee of this transfer, and on testnet the fees of every other transaction on that node. So `before + 1 Hbar` is never seen and the poll times out. Check a lower bound instead, and adjust the comment at lines 67-68.

```suggestion
            return hbars.toTinybars().greaterThanOrEqual(expected.toTinybars())
                ? hbars
                : null;
```

examples/react-native-example-legacy/App.tsx
```diff
@@ -115,3 +132,28 @@ const App = () => {
...
+    const deadline = Date.now() + 60_000;
+    for (;;) {
+        const remaining = deadline - Date.now();
+        if (remaining <= 0) {
+            throw new Error("mirror node did not ingest in time");
+        }
+        const balance = await new MirrorNodeAccountBalanceQuery()
+            .setAccountId(accountId)
+            .execute(client, remaining);
+        if (balance.hbars.toTinybars().equals(expectedTinybars)) {
```
ddeskov-limechain: Same node fee issue. 0.0.3 on testnet receives fees from every transaction on that node, so the exact value is never seen. The app waits 60 s and then alerts. Use a lower bound.

```suggestion
        if (balance.hbars.toTinybars().greaterThanOrEqual(expectedTinybars)) {
```

src/long.js
```diff
@@ -4,7 +4,7 @@ import BigNumber from "bignumber.js";
 
 /**
  * @typedef {{low: number, high: number, unsigned: boolean}} LongObject
- * @typedef {import("long").default} Long
+ * @typedef {import("long")} Long
```
ddeskov-limechain: This emits `export type Long = typeof import("long")` in `lib/long.d.ts`, which is the module namespace, not a Long value. It is also not needed for this issue. Please revert.

```suggestion
 * @typedef {import("long").default} Long
```

test/unit/node/AccountBalanceDeprecation.js
```diff
@@ -0,0 +1,159 @@
...
+            // Read over HTTP from the mirror node, not from a consensus node.
+            expect(getChannel).toHaveBeenCalledTimes(0);
+            expect(fetchMock).toHaveBeenCalledTimes(1);
+            expect(fetchMock.mock.calls[0][0]).to.include("/balances");
+
+            client.close();
+        });
+
+        it("forwards an explicit mirror request timeout", async function () {
+            const timeout = vi.spyOn(AbortSignal, "timeout");
```
ddeskov-limechain, line 77: `Date.now()` runs twice between the deadline and the fetch. When the millisecond ticks in between, the value is 4320 and the test flakes. Freeze the clock.

```suggestion
            const timeout = vi.spyOn(AbortSignal, "timeout");
            vi.spyOn(Date, "now").mockReturnValue(0);
```

ddeskov-limechain (changes requested):
Thanks, the shared polling helper and the exact checks are the right direction. A few things must change before this can go in.

**Blocking**

- `consensus/create-with-revenue.js`: the expected collector credit is 2 Hbar but the topic fee is 1 Hbar, so that poll always times out. Two operator reads there can also be stale (see inline).
- `account/transfer-crypto.js` and `react-native-example-legacy/App.tsx`: 0.0.3 is a node account and also receives node fees, so an exact balance is never reached. Use a lower bound (see inline).
- `src/long.js`: the typedef change emits a wrong public type. Please revert (see inline).
- Rebase on `main`. #4306 was squash merged, so this branch conflicts. Cherry-picking your two commits onto `main` applies cleanly and gives the same diff.

**Non-blocking**

- `skipLibCheck: true` in `examples/tsconfig.json` and the REST signer hides TS2694 errors from `lib/*.d.ts` (`import("long").default` does not resolve under `moduleResolution: node`). OK as an unblock, but please open a follow-up issue for the SDK declarations and link it here. That is also the real cause of the `TokenBalanceValue` casts in four examples.
- `run-all-examples.js` now defaults to concurrency 1, which makes the CI examples job slower. Fine given the shared operator, just say so in the PR description.
- The `react-native-example` type fixes and the REST signer eslint plugin swap are unrelated to the issue. Keep them if the lint needs them, but list them in the description.
- Please add a CHANGELOG line: `MirrorNodeTokenBalanceQuery` now throws on a malformed token response, and `Provider.getAccountBalance` accepts an optional `requestTimeout`.
- The issue asks for recorded run evidence for `batch-tx.js`, both React Native apps and the REST signer, since the runner skips them. Please add it with the SDK commit and the Solo and mirror versions used.

**Checked locally** on your two commits cherry-picked onto `main`: `tsc`, the full node unit suite (164 files, 2082 tests), examples `tsc` and prettier, REST signer `tsc` and eslint. No file under `examples/`, `common_js_test/` or `tck/` constructs `AccountBalanceQuery` anymore; the remaining mentions are comments.

AgrimTawani:
Addressed the review:

- rebased onto `main`; fixed the 1 HBAR expectation and both mirror-ingestion orderings
- changed node-account checks to lower bounds, restored the `Long` value type, and froze the test clock
- added changelog/docs, CI evidence steps, and opened #4373 for the declaration follow-up
- local Node/browser suites plus build, type, lint, formatting, and cycle checks pass

Build & Test and Common JS now need maintainer approval to run on the rebased fork commit.

AgrimTawani:
Pushed the examples CI fixes in `64d82029`:

- configured the HIP-542 custom client with the local mirror network
- signed airdrop cancellation with the treasury key
- corrected impossible 100/200 HBAR transfers from 20 HBAR accounts
- fixed nested example exclusions so streaming examples do not run in the bounded runner

Targeted discovery, formatting, ESLint, and examples TypeScript checks pass. Build & Test and Common JS need approval again for the new fork commit.

AgrimTawani:
@ddeskov-limechain All seven inline findings from your review are addressed, and the branch is rebased onto current `main`.

**Blocking items**

- `consensus/create-with-revenue.js` — the fee collector expectation is now `+1 HBAR`, matching the actual topic fee. The 2 HBAR value remains only as Alice's `CustomFeeLimit`, as you noted.
- `consensus/create-with-revenue.js` — both stale reads are fixed by reordering, exactly as you suggested. The topic create is now Step 1 and Alice's create is Step 2, so Alice appearing on the mirror proves the topic create was ingested before the operator read. Likewise the token transfer to Alice now happens after the topic update, so the exact `aliceTokenBefore` read proves both are present.
- `account/transfer-crypto.js` and `react-native-example-legacy/App.tsx` — node account `0.0.3` is now checked with `greaterThanOrEqual` lower bounds instead of exact equality, and the comment at the top of `transfer-crypto.js` was adjusted.
- `src/long.js` — the typedef is reverted to `import("long").default`. The `@ts-ignore` above it is retained with a pointer to #4373.
- `test/unit/node/AccountBalanceDeprecation.js` — `Date.now()` is now frozen with `vi.spyOn(Date, "now").mockReturnValue(0)`.
- Rebased onto `main`; the PR contains only the issue commits and GitHub reports it mergeable.

**Non-blocking items** — the changelog now documents both the `MirrorNodeTokenBalanceQuery` malformed-response behavior and the optional `Provider.getAccountBalance` `requestTimeout`; #4373 is open for the SDK declaration issue; and the PR description explains the concurrency-1 default and lists the React Native and REST signer changes.

One housekeeping note: I force-pushed to re-sign the commits so they show as **Verified** (they were signed before, but my committer email did not match the UID on my GPG key, so GitHub reported them unverified). The head moved from `64d82029` to `8e99a998` — **the tree is byte-identical**, only the commit metadata and signatures changed. DCO is green.

Build & Test and Common JS still need maintainer approval to run on the fork commit. Could you approve those runs and re-review when you have a moment?

AgrimTawani:
Implemented the requested balance-example and mirror-ingestion fixes, including the review updates and test hardening. Please review when you have a chance.

PR #4375, opened 2026-09-21
fix: reject unknown node account IDs instead of routing to a random node

### LINKED ISSUE #4330: `Client.ping()` silently probes a random node when the given node account ID is not in the network map

ddeskov-limechain:
**Description**:

`ManagedNetwork.getNode(key)` fell back to a random network entry when the key was not in the network map, so `Client.ping()` and any request pinned with `setNodeAccountIds` to an unknown node silently executed against an arbitrary node and reported success. The `NodeAccountId not recognized` guard in `Executable._getExecutionNode` was unreachable.

* `getNode(key)` now throws `NodeAccountId not recognized: <id>` on a miss; the random fallback and the unreachable guard are gone (one error path).
* `_getExecutionNode` skips a pinned ID that has left the map while another pinned ID still resolves, and rethrows only when none of them do, so a multi-node request still survives an address book refresh or the `updateNetwork()` after `INVALID_NODE_ACCOUNT_ID`.
* Tests: `ManagedNetwork` unknown-key case, `Executable` skip / reject-before-any-request cases, and a `ClientPingMocking` case for `ping("0.0.111")` against a single-node mock network. All five fail on `main`.
* CHANGELOG `### Fixed` entry noting the behavior change for receipt/record queries pinned to a node that has since left the map.

Supersedes #4333, which was not updated after review.

**Related issue(s)**:

Fixes #4330

**Notes for reviewer**:

The TCK server needs no change: `tck/methods/sdk.ts` already lets a rejected `ping` surface as a JSON-RPC error, which is what `ClientPing` test 3 (hiero-ledger/hiero-sdk-tck#677) expects.

**Checklist**


Fantomasa (approved):
LGTM

